PR #171, opened 2014-09-10
Hanging fix

sharadmv:
Puts a timeout on the refresh token query. I'm not sure if this is actually the hanging problem, but it's a good thing to have.

Don't merge yet please.


client/auth.py
```diff
@@ -46,7 +46,7 @@ def make_refresh_post(refresh_token):
         resource_endpoint='https://www.googleapis.com/oauth2/v1',
         client_id=CLIENT_ID, client_secret=CLIENT_SECRET)
     params = {"grant_type": "refresh_token"}
-    client.request_token(refresh_token=refresh_token, **params)
```
papajohn: Testing

PR #319, opened 2014-11-26
Remove -s shorthand, use --server instead

### LINKED ISSUE #315: Client should halt on unknown flag
You can currently mistype flags, and ok runs normally. For instance, one student entered:

python3 ok -summit

The correct behavior in this case is to halt and display the CLI help.

albert12132:
Fixes #315 . Command line typos like `-submit` were being interpreted by argparse as `-s ubmit` (the `server` flag). Now, the server can be specified only with `--server`.


albert12132:
This has been implemented in the new repo.

PR #325, opened 2015-01-10
Remove client

albert12132:
I've moved the client script to a [separate repository](https://github.com/Cal-CS-61A-Staff/ok-client) to maintain modularity. This pull request removes the client script from this repo.

Server tests appear to be failing. Not sure why though.


Some3141:
We're doing a server rewrite, so I think we can solve the failing tests while that's going on.

LGTM. Merging now.

PR #329, opened 2015-01-14
Fix broken server tests

papajohn:
Some server tests were failing because code was commented out. This PR re-instates that code, which may be problematic, but brings the code & tests back into agreement (a better starting point for model revisions).


Some3141:
Server tests are still broken. I think the fixes is uncommenting a line in utils.py and models.py. The commit that broke everything is here: https://github.com/Cal-CS-61A-Staff/ok/commit/79b65bcc56039478d66052a750af51f79ee5bc63

papajohn:
Hmm, tests pass locally for me, and I did previously make those changes to utils.py that you chatted me.

Travis died with an unrelated error.

    The command "if [[ $TRAVIS_PYTHON_VERSION == "2.7" ]]; then wget https://storage.googleapis.com/appengine-sdks/featured/google_appengine_1.9.15.zip -o gae_sdk.zip; fi" failed and exited with 8 during .

Some3141:
It's the google app engine version... I think we'll just occasionally have to bump it when the wget command fails. There's no "get latest" command to make this easier, but I did pull it into a file.

Anyway, LGTM. Merging now.

PR #330, opened 2015-01-15
Ok backend rewrite

papajohn:
Revise models to match the description in https://github.com/Cal-CS-61A-Staff/ok/wiki/Models


moowiz:
Thanks so much for doing all of this! I think I was getting tired of staring at the same code over and over again. 

Do we have tests for this code? :D

Some3141:
Update: Stephen and I are working on tests for this code.

papajohn:
FYI, I'll be available around 8pm tonight to continue helping with this, but not before. Thanks for working on the tests! I think the `app/seed` directory needs some revision too in order to `start_server` successfully.

server/app/api.py
```diff
@@ -285,25 +290,31 @@ def statistics(self):
...
+        'add_email': {
+            'methods': set(['PUT']),
             'web_args': {
-                'first_name': Arg(str),
-                'last_name': Arg(str),
-                'login': Arg(str),
+                'email': Arg(str)
+            }
+        },
+        'delete_email': {
```
papajohn: Once all emails are deleted, a user will not be accessible. I think the standard solution is to prohibit deleting the last email. I'm adding a `_pre_put_hook` to the `User` model, but I'm not sure that's the best way to enforce this issue.

server/app/api.py
```diff
@@ -423,15 +472,13 @@ def post(self, user, data):
         return super(AssignmentAPI, self).post(user, data)
 
     def assign(self, obj, user, data):
```
papajohn: Needs comment: What does it mean to assign an assignment?

server/app/api.py
```diff
@@ -967,59 +1244,43 @@ class GroupAPI(APIResource):
                 'members': KeyArg('User')
             }
         },
+        'invite': {
```
Sumukh: What's the purpose of this endpoint? I can't seem to find the associated method.

Some3141: Yeah, for some reason my invite method got deleted... -_-

papajohn:
Looks good to me. Merge whenever you want!

PR #332, opened 2015-01-25
Send course id back to client

Sumukh:
Send the courseId back to the client.  

I've tested it and it seems to work and doesn't seem to crash clients that don't accept the courseId. This isn't deployed yet. 

@Some3141

### DIFF
--- a/server/app/api.py
+++ b/server/app/api.py
@@ -1044,7 +1044,8 @@ class SubmissionAPI(APIResource):
                                                messages, submit, submitter)
         return (201, 'success', {
             'key': submission.key.id(),
-            'course': valid_assignment.course.id()
+            'course': valid_assignment.course.id(),
+            'email': user.email[0]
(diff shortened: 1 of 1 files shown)


server/app/api.py
```diff
@@ -1043,7 +1043,8 @@ def submit(self, user, assignment, messages, submit, submitter=None):
         submission = self.db.create_submission(user, valid_assignment,
                                                messages, submit, submitter)
         return (201, 'success', {
-            'key': submission.key.id()
+            'key': submission.key.id(),
+            'course': valid_assignment.course.id()
```
Sumukh: @Some3141 @albert12132  - Is there a reason to not send the full URL and have the client display that url?

Some3141:
LGTM. Merge whenever the server is ready. I think @Sumukh is coordinating this. Thanks!

PR #335, opened 2015-01-26
Add Sudo Mode for Admins

Sumukh:
An admin can su into another users account. This should be useful for looking at ok issues or viewing code (though it's intended for site admins - not staff because there are no course level restrictions). 

The way to access it is to login as a user and then go to `server-url.appspot.com/sudo/su/student@email.com`
It logs any attempts to sudo and displays a warning to admins. 

ok sudo 2015-01-25 23-11-13

I think this is a useful feature for admins and it should work with no further changes.

### DIFF
--- a/server/tests/integration/test_sudo.py
+++ b/server/tests/integration/test_sudo.py
@@ -21,6 +21,7 @@ from app import auth
 from app.authenticator import Authenticator, AuthenticationException
 
 class SudoUnitTest(APIBaseTestCase):
+    url_prefix = ''
     def get_accounts(self):
         """
         Returns the accounts you want to exist in your system.
@@ -45,15 +46,12 @@ class SudoUnitTest(APIBaseTestCase):
             )
         }
 
-    def get(self, url, *args, **kwds):
-        """
-        Makes a get request.
-        """
-        url = url
-        self.response = self.client.get(url, *args)
-
     def assertStatusCode(self, code): #pylint: disable=invalid-name
-        self.assertEqual(self.response.status_code, code, self.response.headers)
+        self.assertEqual(
+            self.response.status_code, code,
+            'response code ({}) != correct code ({}).\n{}'.format(
+                self.response.status_code, code,
+                self.response.get_data()[:100]))
 
     def setUp(self):
         super(SudoUnitTest, self).setUp()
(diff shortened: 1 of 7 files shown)


Some3141:
Looks good to me- can you add some tests to make sure that this can't be abused in any way? Your changes look ok to me, but I just want to be sure that future modifications don't break this since it's a very easy target for security leaks.

Sumukh:
@Some3141 - Pushed some tests - but I haven't figured out how to get the login working so the tests fail. Can you take a look? 

Right now it fails because it never logins - and the GET doesn't see a current_user.

Some3141:
Is there a reason that you're not using the assertStatusCode method here: https://github.com/Cal-CS-61A-Staff/ok/blob/c665a9b2c08683b7dec29e23fc20fdb6c22c1aa3/server/tests/test_base.py#L200?

Sumukh:
No great reason. I tried moving it into this class to see if it change the result - mainly because the responses being testing in this class should never be json data.

server/tests/integration/test_sudo.py
```diff
@@ -0,0 +1,93 @@
...
+                email=["dummy3@admin.com"],
+                is_admin=True
+            )
+        }
+
+    def get(self, url, *args, **kwds):
+        """
+        Makes a get request.
+        """
+        url = url
```
moowiz: Why is this here?

server/tests/integration/test_sudo.py
```diff
@@ -0,0 +1,93 @@
...
+    def testLoginToSudo(self):
+        self.login('admin')
+        self.get('/sudo/su/dummy@student.com')
+        self.assertStatusCode(200)
+        self.logout()
+
+    def testStudentAttemptToSudo(self):
+        self.login('student1')
+        self.get('/sudo/su/dummy@student.com')
+        self.assertStatusCode(404)
```
moowiz: We want a student trying to sudo to 404? Why, exactly?

server/tests/integration/test_sudo.py
Sumukh: That line shouldn't be there. I brought in the get function because this test shouldn't be making json requests.

server/tests/integration/test_sudo.py
Sumukh: It should be a 403. I think I used 404 because I hadn't implemented the 403 listener.

moowiz:
Don't change the assertStatusCode. It needs to be different for `test_sudo.py`.

Sumukh:
LGTM. Will merge and deploy with the rest of the admin features.

Issue #331, opened 2015-01-25
Final Submissions do not merge when group invite accepted


Sumukh:
Case: I have a final submit. If Soumya invites me and I accept it we both now have no Final Submission. Backups and normal submissions work. 

@papajohn is this a design choice? From what I can read in the wiki - it doesn't appear to be. Relevant section quoted below: 

If it's not a design choice: @Some3141 

>Whenever a new active member is added to a group, the *group final submission is updated to the most recent submission* of any active member of the group.
Whenever the final submission of a person changes, that person should be notified with a link to the new final submission. This can occur because (a) they joined a group with a more recent final submission than theirs, or (b) someone else joined their group with a more recent final submission than the group's current final submission.

papajohn:
That's a bug.

papajohn:
I'm investigating now. Any news, @moowiz ? Thanks.

papajohn:
Sorry, didn't make much progress, but at least I know what needs to be done. I can finish up after lecture. Started a PR: https://github.com/Cal-CS-61A-Staff/ok/pull/337

moowiz:
Lemme make a test and add it to the PR for you.

moowiz:
How were you checking that they had a final submission?

Sumukh:
The frontend relies on `get_course_info` to render the dashboard (https://github.com/Cal-CS-61A-Staff/ok/blob/master/server/app/models.py#L212)

moowiz:
Added a test for the bug.

PR #337, opened 2015-01-26
Update final submissions

papajohn:


papajohn:
Ok, making progress. We need some more tests to make sure final submissions are correct under many circumstances (leaving groups, etc.), but at least #331 should be fixed.

Sumukh:
@moowiz - Is this okay to deploy?

Sumukh:
@moowiz - Is this ready to deploy? I think this will fix another issue with the API.

moowiz:
LGTM. Looks like he fixed the bug!

Sumukh:
There might still be issues - I'm trying to fix them - here's what I have so far. I'll keep working on this branch. 

Changelog: 
1. Fixed key vs obj in queries 
2. `latest.group = group.key` - account for group being none 
3. final.key.delete()
4. edit final submission pre_put_hook (see fix_final branch) 
5. We also need to call update_final_submission on the first user who sent the invite because it currently shows No Submission after the invite - but it appears after the other user accepts the invite

PR #340, opened 2015-01-27
Fix and Test FinalSubmission

Sumukh:
I stopped the final branch PR (#337) from crashing in most cases - but there is still a bug (I think it's with the with the pre_put_hook)

Two issues: 
1. On invite - a user loses their final submission  until the other person accepts or rejects the invite
2. There's something off on the version deployed - which assigns a Final Submission to all of the users assignments (This might be affecting the actual datastore - which would be bad. The frontend is also impacted)

@papajohn - Can you take a look at the pre_put_hook of the Group method?

### DIFF
--- a/server/app/api.py
+++ b/server/app/api.py
@@ -1026,14 +1026,13 @@ class SubmissionAPI(APIResource):
                     datetime.datetime.now() >= valid_assignment.lock_date
 
         if submit and late_flag:
-            # Late submission. Do Not allow them to submit
+            # Late submission. Do not allow them to submit
             logging.info('Rejecting Late Submission', submitter)
             return (403, 'late', {
                 'late': True,
                 })
 
         models.Participant.add_role(user, valid_assignment.course, STUDENT_ROLE)
-
         submission = self.db.create_submission(user, valid_assignment,
                                                messages, submit, submitter)
         return (201, 'success', {
--- a/server/app/models.py
+++ b/server/app/models.py
@@ -945,7 +945,7 @@ class FinalSubmission(Base):
     group = ndb.KeyProperty(Group)
     submission = ndb.KeyProperty(Submission)
     queue = ndb.KeyProperty(Queue)
-    submitter = ndb.ComputedProperty(lambda x: x.submission.get().submitter)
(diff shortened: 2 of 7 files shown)


papajohn:
I'll add more tests to ferret out the rest of the bugs.

papajohn:
I added tests for the following scenarios: 
- create a group of 2 (invitation and acceptance)
- try to create a group of 2 but someone declines
- create a group of 2 and then someone leaves

Groups of 3 remain untested, but I think that's ok.

These same scenarios don't have integration tests. I'll try to add one now, but only have a few minutes.

Logic changes:
- get_final_submission now ignores your group if you're only invited, not a member
- pre-put hook for final submission is gone; I think it's too expensive and complicated. Instead, we should just write a lot of tests.
- Invite now contains a call to update_final_submission
- Various small fixes

papajohn:
Actually, I'm out of time for the next several hours — handing off to whoever wants to review this change.

moowiz:
I'll look at this.

Sumukh:
I can still reproduce the invite removes final submission bug. I don't think calling update_final_submission is fixed it - Here's how I test it in case you want to run it

Start the server and login as partner0@teamwork.com, remove partner1 from the Hog Group. Now invite partner1@teamwork.com back - the page will now tell you that you have no submission.

papajohn:
Investigating now. Thanks for the description.

Sumukh:
Perhaps the error has to do with get_group? When `update_final_submission` is called - get_group for the user that sent the invite returns None. I don't think it should return none - it should update the users submission to the group. 

`('Updating user', [u'partner0@teamwork.com'], 'Group: ', None)`

papajohn:
I'm having a hard time constructing a test that exercises the API in the
appropriate way, but so far I've noticed (and corrected locally) the
following error:

- get_final_submission doesn't query on assignment

Sumukh:
Interesting - can you push that and I'll also try to take a look?

papajohn:
Pushed a partial test and some small fixes. Since I can't seem to write a
working regression test, I'll try debugging through the UI now.

PR #350, opened 2015-01-29
Comment out seeding, uncoment in app.__init__ for local testing

Sumukh:
Completely remove the seed from running on init. 

Workflow change: To seed locally, uncomment the three lines in `app.__init__`


Sumukh:
Closing. It's not an issue with the flags - @Some3141 found out that this occurs when app.models is imported during usage of remote_api.

PR #354, opened 2015-01-29
Dash prettify

alvinwan:
- differentiating between assignment "status" and assignment "progress"
- HTML updated

New Version:
screen shot 2015-01-29 at 11 56 01 am

Old Version:
screen shot 2015-01-29 at 11 56 10 am


Some3141:
Can it be colored green when the assignment is done?

Sumukh:
What exactly does done mean here - that there is a submission and the assignment is inactive? 

It will already turn gray when the assignment is inactive.

Some3141:
Sorry- I meant submitted. Basically, that way a student can just quickly scroll through and get a sense of whether or not they submitted.

Sumukh:
:+1:

alvinwan:
Gotcha. Green added and partial updated accordingly.

PR #360, opened 2015-02-03
Changed logging level from exception to warn.

### LINKED ISSUE #358: Warn instead of Error (in logs) when client version is incorrect
Can we change the logging level for incorrect client version from Error to Warn? Right now, finding errors in the server logs is tricky because there are so many incorrect client version errors.

moowiz:
Fixes #358.


Sumukh:
LGTM. Merging.

PR #372, opened 2015-02-10
Bug/moowiz/api staff modification

moowiz:
Fixes some pylint stuff, and fixes some the `add_staff` method.

### DIFF
--- a/server/app/api.py
+++ b/server/app/api.py
@@ -1043,8 +1043,8 @@ class SubmissionAPI(APIResource):
 
     def win_rate(self, obj, user, data):
         """
-        Gets the win_rate for the submission. This method will be removed shortly.
-
+        Gets the win_rate for the submission.
+            This method will be removed shortly.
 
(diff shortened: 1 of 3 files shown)


server/app/api.py
```diff
@@ -1268,8 +1269,7 @@ def add_staff(self, course, user, data):
 
         if data['staff_member'] not in course.staff:
             user = models.User.get_or_insert(data['staff_member'].id())
-            course.staff.append(user.key)
-            course.put()
+            Participant.add_role(user, course, constants.STAFF_ROLE)
```
Some3141: I think we also need to add to course.staff unless we have an @property method?

Some3141:
LGTM. There's some travis bugs.

server/app/api.py
moowiz: It's a property method. It didn't used to be.

moowiz:
Merging.

PR #374, opened 2015-02-12
Adds user merging.

moowiz:
Also adds some tentative uniqueness checks.

### DIFF
--- a/server/app/models.py
+++ b/server/app/models.py
@@ -10,6 +10,7 @@ import datetime
 from app import app
 from app.constants import STUDENT_ROLE, STAFF_ROLE, VALID_ROLES
 from app.exceptions import *
+from app import utils
 from flask import json
 from flask.json import JSONEncoder as old_json
 
@@ -348,6 +349,8 @@ class User(Base):
         """Ensure that a user can be accessed by at least one email."""
         if not self.email:
             raise BadValueError("No email associated with " + str(self))
+
+        utils.check_user(self.key.id())
 
 class Course(Base):
     """Courses are expected to have a unique offering."""
--- a/server/app/models.py
+++ b/server/app/models.py
@@ -648,9 +648,8 @@ class Submission(Base):
     assignment = ndb.ComputedProperty(lambda x: x.backup.get().assignment)
     server_time = ndb.DateTimeProperty(auto_now_add=True)
 
-    def mark_as_final(self):
-        """Create or update a final submission."""
-        assignment = self.backup.get().assignment
+    def get_final(self):
+        assignment = self.assignment
         group = self.submitter.get().get_group(assignment)
         submitter = self.submitter
         if group:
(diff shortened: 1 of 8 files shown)


server/app/models.py
```diff
@@ -658,15 +660,50 @@ def mark_as_final(self):
...
         else:
-            final = FinalSubmission(assignment=assignment, submission=self.key)
+            group = self.submitter.get().get_group(self.assignment)
+            final = FinalSubmission(
+                assignment=self.assignment, submission=self.key)
             if group:
                 final.group = group.key
         final.put()
 
+    def resubmit(self, user_key):
```
Some3141: If I recall correctly, there was another model with a submit method- can't you just use that? (It's like SubmitNDBImplementation or something).

server/app/utils.py
```diff
@@ -264,15 +283,128 @@ def assign_submission(backup_id, submit):
...
+    # Re-submit submissions
+    S = ModelProxy.Submission
+    subms = S.query(S.submitter == dup_user_key).fetch()
+    for subm in subms:
+        subm.resubmit(user_key)
+
+    dup_user = get_dup_user()
+    # Change email
+    # Right now, I just append the hash of their key, which should be
+    # something they can't guess... Maybe make it more secure??
```
Some3141: I think the only way to auth through an email is to have Google tell us that something is their email. Can't you just have the email be like 'blah@berkeley.edu@hash_key"? That'll never be a valid email.

server/tests/unittests/test_user.py
```diff
@@ -39,5 +44,115 @@ class UserPermissionsUnitTest(PermissionsUnitTest):
...
+        self.assertNumGroups(self.user, self.assign, 0)
+
+    def test_user_unique_final_single_user(self):
+        self.check_user_final(self.user)
+        self.assertNumUsers(self.user, 1)
+        self.assertNumFS(self.user, self.assign, 0)
+        self.assertNumGroups(self.user, self.assign, 0)
+
+    def test_call_functions(self):
+        user_obj = models.User(email=['test@example.com'])
```
Some3141: Why is this important? Let's say that the check_user thing calls unique_final_submission 1000 times- it should still be okay, I think?

Some3141:
I think I'd feel better with some more tests- add in 2 users with email addresses 'dummY@admin.com' and 'dummy@admin.com' and see if they get merged. Also, gradually include groups that they're part of, final submissions for assignments and stuff as well.

server/app/models.py
moowiz: There's some stuff in there, like `late_flag`, adding roles, converting messages too and from json, that we don't need.

server/app/utils.py
moowiz: Oh, that's fine. What hash key should we use?

server/tests/unittests/test_user.py
moowiz: The idea is that it tests to make sure that these functions are actually called. Not how many times. I can change it to not test how many times it gets called if you want.

server/app/utils.py
Some3141: I'm fine with just having their email twice- it doesn't need to be unguessable, right? It should also be possible to retreive these old users just in case.

server/tests/unittests/test_user.py
Some3141: Then maybe just assert it's not equal to 0 then?

server/app/models.py
Some3141: Hmm... I'm torn. On one hand, I think there's a way to write this so that we don't duplicate code. On the other hand, I'm _really_ hesitant of touching the submission system since that's probably our most critical thing. Maybe add a TODO in here?

server/tests/unittests/test_user.py
moowiz: Sure.

server/app/utils.py
moowiz: Good idea!

moowiz:
@papajohn Can you take a look at this?

server/app/models.py
moowiz: This function is pretty simple... I feel like there's a lot of stuff we have to do to make the submission process itself do what we want, which is actually (I believe) more complicated than this. But, we could use that if you want.

Some3141: I'm not worried about bugs in this function- I'm more concerned with other things that can happen on submission (e.g. autograding) that may or may not happen here. I think for now, this is good, but in the future, we should get the actual submission system to have a hook that does what we want it to do. It's not very high priority, so I'm fine with just leaving a TODO for the future.

Some3141:
Awesome, tested with Max.

PR #388, opened 2015-02-21
Adds basic analytics framework

sharadmv:
Adds a rudimentary map-reduce framework for analytics on data stored in NDBs. The basic idea is that you specify an entity to map over (User, FinalSubmission, etc.), provide a map function over entities, a reduce function over the map result and the rest will be handled for you. You can also provide filters for the map, which are specified via a triple `("<field>", "<comparator>", "<value>")`.

Map results are currently stored in NDB permanently, but it's a quick fix to delete them after the job is done.

The results of the job will be stored in NDB.
Here's a sample code snippet that works. It calculates the average email length.

```python
from analytics import Job

def mapper(entity):
    return map(len, entity.email)

def reducer(map_result):
    total, count = 0, 0
    for result in map_result:
        for length in result:
            total += length
            count += 1
    return float(total) / count

filters = []

j = Job(models.User, mapper, reducer, filters=filters)
j.start()
```

The job will run in the background, using the `deferred` library. The results will be dumped into an NDB AnalyticsDump object. Here the example output for this job (in JSON).

```json
{
  "data": {
    "created": "2015-02-20 20:51:25.064035", 
    "id": 4556376185503744, 
    "map_count": 16, 
    "map_result": [
      [
        16
      ], 
      [
        20
      ], 
      [
        21
      ], 
      [
        20
      ], 
      [
        21
      ], 
      [
        20
      ], 
      [
        20
      ], 
      [
        20
      ], 
      [
        15
      ], 
      [
        20
      ], 
      [
        21
      ], 
      [
        20
      ], 
      [
        21
      ], 
      [
        20
      ], 
      [
        20
      ], 
      [
        16
      ]
    ], 
    "result": {
      "result": 19.4375
    }, 
    "status": "done"
  }, 
  "message": "success", 
  "status": 200
}
```


server/app/analytics/job.py
```diff
@@ -0,0 +1,93 @@
...
+    result = ndb.JsonProperty()
+    status = ndb.StringProperty()
+
+    def initialize(self):
+        self.map_result = []
+        self.map_count = 0
+
+    def update_status(self, status):
+        self.status = status
+        self.put()
```
moowiz: We might wanna change this to `put_async`? Cause we do this a lot? Not sure if that would be bad for any sort of guarantees though.

moowiz:
Looks interesting!

Can haz tests? :D

server/app/analytics/job.py
sharadmv: Right now, everything runs in serial, so it shouldn't be a problem

server/app/analytics/mapper.py
```diff
@@ -0,0 +1,117 @@
...
+        self.to_delete.append(entity if isinstance(entity, ndb.Key) else entity.key)
+
+    def update(self, entity):
+        self.to_put.append(entity)
+
+    def map(self, entity):
+        """Updates a single entity.
+
+        Implementers should return a tuple containing two iterables (to_update, to_delete).
+        """
```
moowiz: Add `raise NotImpelementError` or something :)

server/app/analytics/job.py
```diff
@@ -0,0 +1,93 @@
...
+        self.job_dump = job_dump
+        self.job_mapper = job_mapper
+        self.set_filters(filters)
+
+    def map(self, entity):
+        try:
+            map_result = self.job_mapper(entity)
+        except:
+            raise deferred.PermanentTaskFailure()
+        self.job_dump.map_result.append(map_result)
```
moowiz: Is it wise to always save the result of the map? It could generate a lot of data... right?

moowiz: Oh, whoops, didn't read the description.

Maybe we should make this deletion easier? Or make it optional to not output map results? If there's a common MapReduce job we run, I can imagine not wanting to have to save the output a lot. And that would save on datastore stuff.

sharadmv: Yea, so I'm saving the results right now for debugging purposes and I'll add the option to make it optional to store those results.

sharadmv: Also, hopefully most common jobs will be heavily filtered, so not too much data is generated.

server/app/analytics/job.py
sharadmv: Actually I take that back, there are issues with finished jobs having the correct state. Going to revert this.

moowiz:
LGTM

Some3141:
Okay, let's merge. I talked to @papajohn about having a DSL or arbitrary Python and I think we came to the conclusion that the correct way to go was not having that. If someone wants to run a job, they should do it through the analytics API, which means that we can check permissions and stuff on a per-job basis.

PR #408, opened 2015-03-03
Relaxes analytics permissions for admins

sharadmv:


server/app/analytics/job.py
```diff
@@ -135,9 +135,13 @@ def update_status(self, status):
...
     def _can(cls, user, need, obj, query):
         if need.action == "index":
+            if user.is_admin:
+                return query
             return query.filter(AnalyticsDump.owner == user.key)
         if need.action == "create":
             return True
         if need.action == "get":
+            if user.is_admin:
+                return True
```
Some3141: Can we just say at the top "if user.is_admin: return True"?

sharadmv: No, we need to handle the index case separately since we return `query`.

Some3141: Oh, didn't see that.

Some3141:
LGTM.

PR #414, opened 2015-03-12
Changed get_course_info to count instead of fetch all.

moowiz:


Sumukh:
Pending changes:

- This breaks the current sorting (lines 207 and 231) 
` all_backups.sort(key=lambda x, y: int(-5*(int(x.server_time > y.server_time) - 0.5)))`

- The frontend needs to adapt to the new changes.

Some3141:
We really only care about sending the first 10 file contents, so if we can do this without reading everything, it's still an optimization. (Not sure if it'll be enough)

Sumukh:
A few students can't use the web site at all because of this call timing out. It'd be great if we can get this working soon - we're pretty close!

moowiz:
Does the frontend still need to adapt to these changes?

moowiz:
@Some3141 can you shed some light on this sorting? I'm confused why we can't just use a [order](https://cloud.google.com/appengine/docs/python/ndb/queries#order) in the queries...

Sumukh:
@moowiz  re: frontend - Yes - but it'll be much easier to do once the call works.

Some3141:
@moowiz - I think I might not have realized I could order by time.

Also, is this ready to merge?

server/app/models.py
```diff
@@ -202,10 +229,14 @@ def get_submissions(self, assignment, num_submissions=10):
         all_subms.sort(lambda x, y: int(-5*(int(x.server_time > y.server_time) - 0.5)))
 
         for subm in all_subms[:num_submissions]:
-            subm.messages = None
+            subm.messages = None # What is this here for???
```
Some3141: So we don't send the messages back to the server.

Some3141:
Okay, everything looks good to me.

Sumukh:
@Some3141 - This isn't ready to merge. You can test by starting the server and trying to view the dashboard. 

- [ ] This breaks the current sorting (lines 207 and 231) 
all_backups.sort(key=lambda x, y: int(-5*(int(x.server_time > y.server_time) - 0.5)))

- [ ] The frontend needs to adapt to the new changes.

Sumukh:
LGTM.

PR #418, opened 2015-03-13
Made final submissions not be created after the due date for an assignment.

moowiz:
@Some3141


papajohn:
How will people submit late?

moowiz:
They can still submit, and it'll show up as a `Submission`. This change only affects if something gets chosen as a `FinalSubmission`.

If they want to pick a late submission as their actual submission, they should either
1. Talk to the course staff
2, Select it on the frontend (coming soon! @Sumukh )

papajohn:
Just as long as staff have some reasonable way to promote a submission to
final, this change sounds good.

moowiz:
There actually isn't... you can go in and manually edit the database, but that's a little clunky.

We should add some sort of API endpoint or... something. @Some3141 what do you think?

Sumukh:
^ I think that was part of the plan - but that endpoint doesn't exist yet - the front-end has the UI there to promote submissions (clicking the flags)
image


Another thing we'd need on the front-end is to be able to display more than the last 10 submissions.

moowiz:
How does the frontend promote the submissions?

Sumukh:
It doesn't yet - it just has the UI there, so when the endpoint is there - it can easily use it.

moowiz:
Oh, I see. I'll add the endpoint soon then.

moowiz:
@Sumukh I added the endpoint. Just do a post on `api/v1/final_submission` with the submission as an argument, and it'll pick that one to be your new final submission.

Some3141:
I'll assign it to Sumukh for now.

@Sumukh - after you do the frontend change, reassign this issue to me and I'll code review everything.

Sumukh:
Should be ready to review. Please take note of the permissions change in FinalSubmission. 

Before this is merged - the API/Model should return an error if a student attempts to make a FinalSubmission after the lock_date.

server/app/models.py
```diff
@@ -654,16 +654,11 @@ class Submission(Base):
     assignment = ndb.ComputedProperty(lambda x: x.backup.get().assignment)
     server_time = ndb.DateTimeProperty(auto_now_add=True)
 
+
     def get_final(self):
         assignment = self.assignment
         # I have no idea why this works... need it to pass tests
```
Some3141: Do we need this comment still? XD

server/app/models.py
```diff
@@ -1108,6 +1103,11 @@ def backup(self):
 
     @classmethod
     def _can(cls, user, need, final, query):
```
Some3141: We should say that you can't touch the final submission at all after the lock date unless you're an admin.

Some3141:
Okay, it looks good to me- especially nice job on the tests. Just a couple of small comments and then we can merge.

server/app/models.py
papajohn: Course staff need to be able to edit final submissions as well. For this
system to be usable outside of 61A, there has to be no regular activity
that requires admin access.

-John

Sumukh:
@Some3141 - Can you take a look at this again - The only thing left is making sure that the permission are right.

Some3141:
Okay, yeah. I think this looks okay. It can't be automerged though.

@moowiz - can you resolve the merge conflicts?

moowiz:
Resolved the conflicts. Merging once Travis says its ok.

PR #422, opened 2015-03-30
Submission endpoint accepts revisions

Sumukh:
This PR enables support for assignment revisions - it still needs to be tested.

When an assignment is in the revision period (toggled manually) - we will allow late submissions from students who have a graded final submission. If so, we tack the new submission to a new attribute of the final submission.

Todo in this PR: 
- [x] Acknowledge revision on student dashboard, 

Future work:
- [ ] queues indicate which submissions have been revised, 
- [ ] grading interface support viewing two submissions 
- [ ] grading view should just diff between the two submissions. (Maybe?)

### DIFF
--- a/server/app/api.py
+++ b/server/app/api.py
@@ -1107,8 +1107,9 @@ class SubmissionAPI(APIResource):
         due = valid_assignment.due_date
         late_flag = valid_assignment.lock_date and \
                     datetime.datetime.now() >= valid_assignment.lock_date
+        revision = valid_assignment.revision
 
-        if submit and late_flag:
+        if submit and late_flag and not revision:
             # Late submission. Do not allow them to submit
             logging.info('Rejecting Late Submission', submitter)
             return (403, 'late', {
--- a/server/app/api.py
+++ b/server/app/api.py
@@ -1109,12 +1109,22 @@ class SubmissionAPI(APIResource):
                     datetime.datetime.now() >= valid_assignment.lock_date
         revision = valid_assignment.revision
 
-        if submit and late_flag and not revision:
-            # Late submission. Do not allow them to submit
-            logging.info('Rejecting Late Submission', submitter)
-            return (403, 'late', {
-                'late': True,
-                })
+        if submit and late_flag:
+            if revision:
+              # In the revision period. Ensure that user has a previously graded submission.
+              fs = user.get_final_submission(assignment)
+              if fs is None or fs.submission.score is None:
+                return (403, 'late', {
+                  'late': True,
+                  'message': "No Previous Submission",
+                  })
+            else:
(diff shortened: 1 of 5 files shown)


papajohn:
Who's the best person to review this change? Stephen, perhaps? (It's fine if the answer is me.)

Sumukh:
Should be ready to merge/deploy. I've tested it locally. These changes will only come into effect once the "revision" property of the assignments is enabled - so we'll be able to test before announcing it. 

Either @papajohn or @moowiz can review - but I'll assign this to @papajohn for now.

server/app/api.py
```diff
@@ -1107,13 +1107,23 @@ def submit(self, user, assignment, messages, submit, submitter=None):
...
         if submit and late_flag:
-            # Late submission. Do not allow them to submit
-            logging.info('Rejecting Late Submission', submitter)
-            return (403, 'late', {
-                'late': True,
-                })
+            if revision:
+                # In the revision period. Ensure that user has a previously graded submission.
+                fs = user.get_final_submission(valid_assignment)
+                if fs is None or fs.submission.get().score == []:
```
papajohn: Should we log this case as well with `logging.info` so that we can see how many submission failures are happening?

server/app/models.py
```diff
@@ -234,6 +234,13 @@ def get_course_info(self, course):
                 assign_info['final']['backup'] = \
```
papajohn: All these long lines would go away if this block started with 

`final = assign_info['final']`

server/app/models.py
```diff
@@ -672,10 +682,14 @@ def get_final(self):
...
-
         if final:
-            final.submitter = self.submitter
-            final.submission = self.key
+            assignment = self.assignment.get()
+            if assignment.revision:
+                # Follow resubmssion procedure
+                final.revision = self.key
+            else:
+              final.submitter = self.submitter
```
papajohn: Indent 4 spaces.

papajohn:
Looks good to me! I made a couple of comments, but please just merge at your leisure.

Sumukh:
Everything seems to work - only catch seems to be that memcache needs to be flushed when the revision property is toggled

@moowiz ?

Some3141:
Where are you toggling it from?

moowiz:
I can review it today (I believe)

moowiz:
Are you sure it needs to be flushed? That doesn't really make sense to me...

@Some3141 is correct; if you toggle the revision property from the datastore admin page, then it doesn't update the memcached entity. If you make an API call, and say, do a PUT that changes the revision value, then it'll update the memcached entity.

Sumukh:
Yeah - I was changing it in the datastore.

PR #426, opened 2015-04-01
quantity paramter to get_submsission and get_backups

Sumukh:
Tested locally and it works. The frontend doesn't use this yet.


Some3141:
Bump the GAE version to fix travis. Also, it LGTM.

PR #428, opened 2015-04-01
Cross-browser compatibility issues

alvinwan:
- fixed line numbers for old submission view, in case old links are redirected there
- fixed Mozilla Firefox box-sizing issue (using -moz- CSS prefix)


Sumukh:
:+1:  Deployed.

PR #430, opened 2015-04-03
Stop logging an error on every api request 

Sumukh:
Silences the USING ACCESS TOKEN error that is polluting the logs.


Some3141:
This is probably my bad...

Some3141:
LGTM.

PR #437, opened 2015-05-13
Submission by time

Sumukh:
API endpoint to get a students most recent submission before a certain time. 

I'm not a huge fan of having to fetch 100 submissions and hope that they have file contents - let me know if you see an easier way to do this.

### DIFF
--- a/server/app/models.py
+++ b/server/app/models.py
@@ -180,6 +180,25 @@ class User(Base):
                 FinalSubmission.assignment==assignment_key,
                 FinalSubmission.submitter==self.key).get()
 
+    def get_submission_before(self, assignment_key, before_time):
+        """Get this users last submission before some server time"""
+        if isinstance(assignment_key, Assignment):
+            assignment_key = assignment_key.key
+        group = self.get_group(assignment_key)
+        if group and self.key in group.member:
+            return Submission.query(
+                Submission.assignment==assignment_key,
+                Submission.group==group.key,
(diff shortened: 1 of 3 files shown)


server/app/models.py
```diff
@@ -180,6 +180,34 @@ def get_final_submission(self, assignment_key):
                 FinalSubmission.assignment==assignment_key,
                 FinalSubmission.submitter==self.key).get()
 
+    def get_submission_before(self, assignment_key, before_time):
+        """Get this users last submission before some server time"""
+        if isinstance(assignment_key, Assignment):
+            assignment_key = assignment_key.key
```
Sumukh: There's a little bit of code duplication here (between the normal `get_submission_before`

server/app/models.py
```diff
@@ -180,6 +180,34 @@ def get_final_submission(self, assignment_key):
...
+                Submission.server_time < before_time).order(
+                  -Submission.server_time
+                ).fetch(100)
+        else:
+            subms = Submission.query(
+                Submission.submitter ==self.key,
+                Submission.assignment == assignment_key,
+                Submission.server_time < before_time,
+                ).order(
+                  -Submission.server_time
```
Sumukh: This `-` is worth investigation - it  might actually do the opposite of what we want in production. This works in testing because we don't have > 100 submissions.

Sumukh:
Merging temporarily for grading. This still needs to be code reviewed

PR #450, opened 2015-06-17
Editing Assignment through Dashboard

alvinwan:
"Edit" button and functionality added.

site-edit-assign

### DIFF
--- a/server/app/api.py
+++ b/server/app/api.py
@@ -652,8 +652,7 @@ class AssignmentAPI(APIResource):
                 'course': KeyArg('Course', required=True),
                 'max_group_size': Arg(int, required=True),
                 'due_date': DateTimeArg(required=True),
-                'templates': Arg(str, use=lambda temps: json.dumps(temps),
-                                 required=True),
+                'templates': Arg(str),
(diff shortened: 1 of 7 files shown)


Sumukh:
Can you also add fields for the lock date and the resubmission boolean?

alvinwan:
I just added the lock date and revision fields. What does it mean to have a template for an assignment? An HTML file?

Sumukh:
We need templates to render a diff between a students submission and the starting file that we provide (i.e a hog.py without student code”) 

It’s worth noting that there can be multiple template files (scheme project for example) 

-- 
Sumukh Sridhara

On June 17, 2015 at 3:59:29 PM, Alvin Wan (notifications@github.com) wrote:

I just added the lock date and revision fields. What does it mean to have a template for an assignment? An HTML file?

alvinwan:
I added the upload field; it allows multiple files.

alvinwan:
The changes to the Assignment API for templates have been reverted. I'm stumped.

Sumukh:
Ok merging - but we should figure out templates in #451

Issue #445, opened 2015-05-29
Program or web interface to download all submissions for an individual student


papajohn:
Requested by @albert12132 (though I don't think this functionality is currently used/needed regularly).

moowiz:
Theoretically should be fairly easy to write a program to do this. Maybe ~2 hours of work. If you want it done at some point, just let me know and I could whip it up, I believe.

soumyabasu:
I already have a script that does very similar things and @jathak is working on a PR in ok-client that ports that over (https://github.com/Cal-CS-61A-Staff/ok-client/pull/77)

Thanks a lot for offering though!

Issue #444, opened 2015-05-29
Program or web interface to download all final submissions for an assignment


papajohn:
We need a robust & standard way to download all final submissions for an assignment. (I know @soumyabasu has some kind of script, but let's make sure that the script is sufficiently documented so that exporting submissions is straightforward.

jathak:
I talked with @soumyabasu about adding this to the ok-client. It could then also be used by students to download their latest backup from the server (this might make sharing between partners easier as well). I have a bit of code written for this, but I'd want to see the existing script before proceeding further to make sure I implement it properly.

papajohn:
Adding batch download to `ok-client` sounds like a good idea to me. We need a way to restart a batch download when it's only half-complete and something goes wrong.

jathak:
@soumyabasu, could you upload your script somewhere so I can reference it when implementing this? Thanks!

soumyabasu:
Sent you all of the scripts via email.

The long term solution for using the API is to understand this file: https://github.com/Cal-CS-61A-Staff/ok/blob/master/server/app/api.py

jathak:
I've added this on [this branch](https://github.com/jathak/ok-client/tree/enhancement/jathak/submission-export) where the `--submissions` option overrides the client's normal behavior and instead downloads all  final submissions for the provided assignment (the `--course` option can be used to specify a course ID, though the client will prompt the user to select a course if this isn't provided). It should also be able to resume downloading if something goes wrong.

I don't have automated tests for this yet and I'm not sure how to write them since they depend on a server. I'm also not sure how well this works in relation to partners on an assignment.

soumyabasu:
Nice work- it looks like the right shape. One thing that might get annoying is that we'll have the list of *every* course on ok being displayed all the time. I think your code is laid out well enough though to handle pretty easily.

To test, you should mock out the API calls to the server. Maybe something like [1] might be helpful. Also, you should think of a general way to do this in case someone wants to add a feature like yours and needs to mock out requests to ok-server unless @albert12132 has already done this?

[1] https://pypi.python.org/pypi/requests-mock

soumyabasu:
Oh, one thing- the server might return a permission error, so you should look at the response code and display the right error message. (Imagine if a student tries to download everything)

albert12132:
I'll take a look at this either tonight or tomorrow night. In the meantime,
feel free to make a PR so that we can comment on the code. All of the
current server communication code is untested, so there's not really a good
[...]

PR #453, opened 2015-06-18
Search

### LINKED ISSUE #305: Web-based search for submissions and students by email, submission time, etc.
Putting an issue here to remind me. This may become more necessary for regrade requests and pulling submissions for specific students. There's currently no way to do this with the frontend.

alvinwan:
Search, available through the admin panel "<strike>All</strike> Submissions":

<strike>*Doesn't actually work yet. The code runs without errors and the tokenization/parsing is correct, but I may need to check on KeyArgs or something like that.</strike> Functional!

Of course, tags can be mixed and matched in any way; here are two demos:

site-submissions-search

site-submissions-search2

> **Note**: The date tag only works when -onlyfinal is false (or not specified at all); basically, the FinalSubmission table doesn't have a "server_time" field. Right now, the -onlyfinal tag actually changes the model that the search queries. Instead, it should probably add a filter that checks for final. I can/will file an issue when this is merged.

**Description**

All searches are composed of flags, operators, and their associated value. A flag is denoted by one hyphen, a flag by two, and a string in double quotations or no punctuation at all. Dates, for example, or any other name with hyphens must be enclosed by double quotations. Missing comparison operators are automaticaly assumed to be --equal.

Available flags: user (email), date (YYYY-MM-DD), assignment (name), onlyfinal (true/false)
Available operators: after, before, lt, gt, equal, eq (alias for equal)
Example: -user "test@example.com" -date --before "2015-06-22" -assignment Hog -onlyfinal true

**How it works**

1. Regex match, find all flags, operators, and arguments.
2. Convert all operators to their related functions, imported from operator.
3. Convert arguments to objects where necessary.
4. Compile FilterNodes, using operators on the proper model attribute and argument.
5. Create query using FilterNodes.
6. Query, and return.

fixes #305, #445, #444

### DIFF
--- a/server/app/api.py
+++ b/server/app/api.py
@@ -18,7 +18,7 @@ from app.constants import STUDENT_ROLE, STAFF_ROLE, API_PREFIX
 from app import models, app, analytics
 from app.codereview import compare
 from app.needs import Need
-from app.utils import paginate, filter_query, create_zip
+from app.utils import paginate, filter_query, create_zip, add_to_zip, start_zip, finish_zip
 from app.utils import add_to_grading_queues, parse_date, assign_submission
 from app.utils import merge_user
 
@@ -889,6 +889,15 @@ class SubmissionAPI(APIResource):
         :return:
         """
         
+    def data_for_zip(self, obj):
+        name = obj.submitter.email+'-'+str(obj.created)
+        messages = obj.get_messages()
+        if 'file_contents' not in messages:
+            raise BadValueError('Submission has no contents to download')
+        file_contents = messages['file_contents']
+        
+        return name, file_contents
+        
     def zip(self, obj, user, data):
         """ Grab all files in submission
         :param obj:
@@ -896,12 +905,7 @@ class SubmissionAPI(APIResource):
         :param data: 
         :return:
         """
-        messages = obj.get_messages()
-        if 'file_contents' not in messages:
-            raise BadValueError('Submission has no contents to download')
-        file_contents = messages['file_contents']
-
-        return self.zip_files(str(obj.created), file_contents)
+        return self.zip_files(*self.data_for_zip(obj))
 
     def zip_files(self, name, file_contents):
         """ Zip files
@@ -917,17 +921,17 @@ class SubmissionAPI(APIResource):
                 file_contents[key] = file_contents[key].encode('utf-8')
             except:
                 pass
-        zip = create_zip(file_contents)
-        return name, zip
+        zipfile = create_zip(file_contents)
+        return name, zipfile
 
-    def make_zip_response(self, name, zip):
+    def make_zip_response(self, name, zipfile):
         """
         Makes a zip response using a zip object.
         
         :param zip: 
         :return:
         """
-        response = make_response(zip)
+        response = make_response(zipfile)
         response.headers['Content-Disposition'] = (
             'attachment; filename=submission-%s.zip' % name)
         response.headers['Content-Type'] = 'application/zip'
@@ -1276,17 +1280,18 @@ class SearchAPI(APIResource):
         if data.get('all', False):
             start, end = SearchAPI.limits(data['page'], data['num_per_page'])
             results = results[start:end]
-        file_contents = {}
+        zipfile_str, zipfile = start_zip()
+        subm = SubmissionAPI()
         for result in results:
-            try :
+            try:
                 if isinstance(result, models.Submission):
                     result = models.Backup(key=result.backup)
-                name, zip = SubmissionAPI.zip(SubmissionAPI(), result, user, data)
-                file_contents[name] = zip
+                return subm.download(result, user, data)
+                name, file_contents = subm.data_for_zip(result)
+                zipfile = add_to_zip(zipfile, file_contents, name)
             except:
                 pass
-        subm = SubmissionAPI()
-        return subm.make_zip_response(*subm.zip_files('query', file_contents))
+        # return subm.make_zip_response('query', finish_zip(zipfile_str, zipfile))
 
     
     @staticmethod
--- a/server/app/api.py
+++ b/server/app/api.py
@@ -1222,6 +1222,7 @@ class SearchAPI(APIResource):
         scope = SearchAPI.translate(tokens)
         prime = SearchAPI.objectify(scope)
         query = SearchAPI.querify(prime)
+        print(query.fetch())
         return query.fetch()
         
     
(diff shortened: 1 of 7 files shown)


alvinwan:
@soumyabasu @Sumukh Could y'all take a look? We can search for submissions now!

server/app/api.py
```diff
@@ -1191,24 +1192,100 @@ class SearchAPI(APIResource):
             }
```
soumyabasu: I think there should be a default `num_per_page` and a default `page`. I think the default `page` should be the first page and the default `num_per_page` can either be a small number so the query goes fast (e.g. 10) or a really large number so you get _everything_ for most queries (e.g. 10000).

server/app/api.py
```diff
@@ -1191,24 +1192,100 @@ class SearchAPI(APIResource):
...
+        return dict(data={
+            'results': results,
+            'more': len(results) >= data['num_per_page']
+        })
         
+    
+    @staticmethod
+    def tokenize(query):
+        """ Parses each command for flag, op, and arg """
+        tokenizer = re.compile('-(?P<flag>[\S]+)\s+(--(?P<op>[\S]+)\s+)?"?(?P<arg>[\S][^"\s]+)"?"?')
```
soumyabasu: Can you have a comment as to what this regex specifically does? (Just to make modifications in the future easier)

server/app/api.py
```diff
@@ -1191,24 +1192,100 @@ class SearchAPI(APIResource):
...
+    }
 
     def index(self, user, data):
-        blocks = SearchAPI.blockify(data['query'])
-        tokens = SearchAPI.tokenize(blocks)
-        print(tokens)
+        tokens = SearchAPI.tokenize(data['query'])
+        scope = SearchAPI.translate(tokens)
+        prime = SearchAPI.objectify(scope)
+        query = SearchAPI.querify(prime)
```
soumyabasu: I think this might be cleaner/easier to understand if the querify method querified the tokens and called objectify, which called translate, etc.

soumyabasu:
A couple of stylistic comments, but outside of that, great work!

soumyabasu:
Also, can you add a "Download all" button or something that tosses all of the query results into a zipfile?

Sumukh:
One of the catches about searching for submission is some submissions that don't have code (analytics submissions) - there should be an option to only show submissions that contain code (i.e have file_contents with code) - (Perhaps this could be done by default) 

Overall looks good. For future improvement (interface wise) - instead of a custom query format we should make the search box be a form with a bunch of fields (i.e text boxes, dropdown, date picker). 

Re: FinalSubmission time - we can add a time field to the FinalSubmission model (as a computed property)

soumyabasu:
Also, this resolves #445 #444 #305

server/app/api.py
alvinwan: added

alvinwan:
@soumyabasu Adding

@Sumukh 
Containing code: I'll add an -onlywcode flag. It'll be true by default.
Input boxes: There's no limitation; instead, having a string query gives us greater flexibility. This way, to add a flag and/or an operator, you just append to the dictionary the API, instead of adding a new input field, controller->service argument, and API web_arg, on top of the dictionary. In the future, if someone would like to expand the search functionality, he/she would only need to understand python and the ~50 lines that make up the SearchAPI, instead of understanding the entire system and the JS front-end too.

alvinwan:
@Sumukh I could add a prettified form, but it would (1) take up more space above the fold, making scrolling necessary to see results and (2) convert fields to a string query anyways.

soumyabasu:
Also, from now on, all messages will have file_contents, so you don't need to worry about it.

papajohn:
Hi Alvin,

Can you tell us what happens next with this pull request? Is it ready? Do you want more people to review this? Could we launch it tonight so that Albert & Robert can demo it to their staff tomorrow?

alvinwan:
Hello Professor DeNero,

This branch will have completed and working zip functionality by 5 p.m. It seems like Soumya and Sumukh are good so far, but I'll ask them for a final code review then. This PR will be merged into master by 6 p.m. We can definitely launch this tonight.

alvinwan:
Zip functionality added. Option of downloading (1) number of submissions on the current page, (2) all submissions that match the search query in a zip file. Submissions are placed in directories containing the [user email]-[submission time]/

server/app/api.py
```diff
@@ -1180,35 +1221,157 @@ def post(self, user, data):
...
+    operators = {
+        'eq': op.__eq__,
+        'equal': op.__eq__,
+        'lt': op.__lt__,
+        'gt': op.__gt__,
+        'before': op.__lt__,
+        'after': op.__gt__
+    }
+    
+    flags = {
```
soumyabasu: What does the flags dictionary map onto? (You can either call it flags_to_X or toss it in a comment somewhere) It looks like flags goes from the name to the corresponding object.

server/app/utils.py
```diff
@@ -86,16 +87,35 @@ def create_api_response(status, message, data=None):
...
+def finish_zip(zipfile_str, zipfile):
+    return zipfile_str.getvalue()
+
+
+def add_to_zip(zipfile, file_contents, dir=''):
+    """
+    Adds files to a given zip file. Uses specified dir to store files.
+    """
+    for filename, contents in file_contents.items():
+        if filename.endswith('.py'):
```
soumyabasu: You should probably remove the print statements. Also, you don't want to differentiate between python files and other files.

soumyabasu:
Okay, done with my review.

server/static/partials/admin/submission.list.html
```diff
@@ -47,25 +52,25 @@ <h4 class="box-title" ng-if="submissions.length == 0" >No Submissions</h3>
...
           </td>
-          <td>
-              <div class="btn-group download-btn-group">
-                  <a ui-sref='submission.detail({submissionId:submission.id})' type="button" class="btn btn-flat btn-info">View</a>
-                  <a href='/api/v1/submission/{{submission.id}}/download' type="button" class="btn btn-flat btn-info">Download</a>
-                  <a ui-sref='submission.diff({submissionId:submission.id})' type="button" class="btn btn-flat btn-info">Diff</a>
-              </div>
-          </td>
+          <!--<td>-->
+              <!--<div class="btn-group download-btn-group">-->
```
Sumukh: If you are not planning on using this, you can delete these lines.

server/app/utils.py
alvinwan: Ahaha... that was my mistake. I was trying to figure out why python files weren't showing up (encoding).

soumyabasu:
Cool- merging as soon as Travis passes.

alvinwan:
@soumyabasu @Sumukh All changes made. Is this good to go?

soumyabasu:
Yup. I'll merge it as soon as travis passes.

alvinwan:
Sounds good.

PR #454, opened 2015-06-18
Partner Display and Reorder [Do Not Merge]

alvinwan:
Allows group members to switch and save order.

Front-end UI:

site-partner-switch


server/app/api.py
```diff
@@ -1431,7 +1431,13 @@ class GroupAPI(APIResource):
             },
         'exit': {
             'methods': set(['PUT', 'POST']),
+            },
+        'save_order': {
```
soumyabasu: You should probably call this "reorder" instead of "save_order".

server/app/api.py
```diff
@@ -1431,7 +1431,13 @@ class GroupAPI(APIResource):
...
+            'methods': {'PUT', 'POST'},
+            'web_args:': {
+                'order': Arg(dict, required=True)
```
soumyabasu: I was mentioning here-  you should take in a list of members here.

server/app/api.py
```diff
@@ -1498,6 +1504,15 @@ def exit(self, group, user, data):
...
         group.exit(user)
+        
+    def save_order(self, group, user, data):
+        """ Saves order of partners """
+        need = Need('save_order')
+        if not group.can(user, need, group):
+            raise need.exception()
+
+        self.order = data['order']
+        self.save()
```
soumyabasu: Okay, so the logic here would go something like this:

new_order = data['order']

if len(new_order) != len(group.member):
    # Tell the frontend that they messed up

for member in group.member:
     if member not in new_order:
           # Tell the frontend they messed up again.

group.member = new_order

server/app/models.py
```diff
@@ -1027,7 +1028,7 @@ def _can(cls, user, need, group, query):
             return False
         if action in ("get", "exit"):
             return user.key in group.member or user.key in group.invited
-        elif action in ("invite", "remove"):
+        elif action in ("invite", "remove", "save_order"):
```
soumyabasu: You'll also probably want to change this to "reorder"

soumyabasu:
Okay, that's the part with save_order. Great work on the frontend stuff though! :)

soumyabasu:
Is this ready to go?

server/app/api.py
```diff
@@ -1757,6 +1763,28 @@ def exit(self, group, user, data):
...
 
         group.exit(user)
+        
+    def reorder(self, group, user, data):
+        """ Saves order of partners """
+        need = Need('reorder')
+        if not group.can(user, need, group):
+            raise need.exception()
+
+        print(data)
```
Sumukh: print statement should be removed.

server/app/templates/student.html
```diff
@@ -17,7 +17,7 @@
         <script src="/static/vendor/jquery/dist/jquery.min.js"></script>
 
         <script src="/static/vendor/jquery-ui/jquery-ui.min.js"></script>
-        <script src="/static/vendor/raven-js/dist/raven.js"></script>
+        <script src="http://cdn.ravenjs.com/1.1.18/jquery,native/raven.min.js"></script>
```
Sumukh: note: should be `//cdn.raven` (might be fixed in another PR)

alvinwan:
@soumyabasu Not yet. I never figured it out; I'll continue working on it tonight or tomorrow.

alvinwan:
Reordering now works.

partner-reorder

@Sumukh @soumyabasu Ready. ... I removed all print's and console.logs this time. :P

soumyabasu:
The backend looks good to me.

@Sumukh - do you have any frontend comments? Otherwise, we're just waiting for the tests to pass.

server/app/templates/admin.html
```diff
@@ -19,7 +19,7 @@
         <link href="/static/css/main.css" rel="stylesheet" type="text/css" />
 
         <link href="/static/vendor/angular-loading-bar/build/loading-bar.css" rel="stylesheet"></link>
-        <link href="/static/vendor/sweetalert/dist/sweetalert.css" rel="stylesheet"></link>
+        <link href="/static/vendor/sweetalert/dist/sweet-alert.css" rel="stylesheet"></link>
```
Sumukh: I think this should be sweetalert.css - I updated the bower file so you may need to run bower install too see that.

server/app/templates/admin.html
```diff
@@ -45,7 +45,7 @@
         <script src="/static/vendor/angular-moment/angular-moment.js"></script>
         <script src="/static/vendor/showdown/compressed/Showdown.js"></script>
         <script src="/static/vendor/jquery.tablesorter/js/jquery.tablesorter.min.js"></script>
-        <script src="/static/vendor/sweetalert/dist/sweetalert.min.js"></script>
+        <script src="/static/vendor/sweetalert/dist/sweet-alert.min.js"></script>
```
Sumukh: ^  same idea.

server/app/templates/student.html
```diff
@@ -32,7 +32,7 @@
         <script src="/static/js/other/moment-timezone-with-data.js"></script>
         <script src="/static/vendor/angular-moment/angular-moment.js"></script>
         <script src="/static/vendor/showdown/compressed/Showdown.js"></script>
-        <script src="/static/vendor/sweetalert/dist/sweetalert.min.js"></script>
+        <script src="/static/vendor/sweetalert/dist/sweet-alert.min.js"></script>
```
Sumukh: same thing about sweetalert

Sumukh:
Frontend looks good - but I think sweetalert has been updated on master to be a newer version that doesn't use a dash. You will have to run bower install to get the latest version 

Once you fix that and it works - we should be good to merge.

PR #456, opened 2015-06-20
Export composition scores as CSV

LuiseV:
DO NOT MERGE


papajohn:
Hi @LuiseV and @alvinwan , summer 61A has started, and autograding is going to be integrated in the next couple of days. We'll need score export pretty soon. Do you think either of you will have time to keep going? If not, it's fine — I can help or find someone to help.

LuiseV:
I'm working right this moment! I hope to have the backend ready by tonight
or tomorrow. I've also tried to modify the JavaScript files - hopefully
correctly. At least I know that the API call goes through. Haven't hooked
it up to the html files though. Maybe you can do that, Alvin?

Best,
Luise
Den 24/06/2015 19.59 skrev "John DeNero" <notifications@github.com>:

LuiseV:
@papajohn  Also, currently, I am looking up all students enrolled in the course, and building the CSV from that. If a student didn't submit or if the submission isn't graded, there will still be a line for the student in the output file. Probably with score 0, and no info for message and grader. 
Another way to do it is to start with all final submissions, and create the CSV from those. If I do that, students who didn't submit a final submission simply won't appear in the CSV. 
Which way do you prefer?

papajohn:
I think it's best to have consistent rosters across assignments, which
means giving 0's for people who do not submit and always including everyone
enrolled.

alvinwan:
@LuiseV Do you still need help linking with the front-end?

LuiseV:
@alvinwan that would be amazing, thanks! As I mentioned, I think the only part that is missing is the html. I just pushed a functioning version, so the backend should pretty much done. 
Thanks!

alvinwan:
Sounds good. I'm going to resolve a regex bug first; I'll add and push right after.

LuiseV:
@alvinwan  And just as a heads up, this is what I did in the JavaScript files: 
- in admin/controllers.js: Created the function $scope.download_compScores() under AssignmentDetailCtrl which calls Assignment.download_composition_scores as specified in common/services.js
- in common/services.js: Under Assignment, specified URL for API call to Assignment.download_submission_scores. 
I know that the URL to the API call works, but I am not 100 % sure about the part in controllers.js. This is my first attempt with js ;)

alvinwan:
Sounds good. I'll take a look and ask if I have any questions about the API's response object.

alvinwan:
Actually, just let me know when it's ready, and I'll create a new branch, merge your changes in, and create another PR. I'm not sure how to checkout a branch from your fork.

*Doing the above ^ will also mark your branch as merged.

LuiseV:
@alvinwan Sounds good. . Let me know if you have any questions, or if I can do something to help out. I'm gonna call it a day soon (I'm on the East Coast, so a few hours ahead), but I can look at things tomorrow.

alvinwan:
Awesome - I'll create a new branch and PR.

LuiseV:
@papajohn @alvinwan  There seems to be a build error with my branch. I assume that you see the warning here as well. It is the call: `wget https://storage.googleapis.com/appengine-sdks/featured/google_appengine_$GAE_VERSION.zip -o gae_sdk.zip`. Do you know why?

alvinwan:
@LuiseV Because I'm unable to push directly to your branch, I'm investigating here #481. I'll let you know if/when I figure it out.

LuiseV:
@alvinwan Sounds good, thanks!
Perhaps either [this](http://stackoverflow.com/questions/5884784/how-to-pull-remote-branch-from-somebody-elses-repo) or [this](http://stackoverflow.com/questions/9153598/how-do-i-fetch-a-branch-on-someone-elses-fork-on-github) can tell you how to checkout and push to a fork next time? :)

alvinwan:
@LuiseV haha thanks for the links! I actually already tried adding a remote and checking out/tracking your branch; I can't push to your branch because of a permissions error, though. If you could grant me permissions, I can work off of your PR and branch instead.

alvinwan:
@LuiseV Even with push privileges, git is giving me errors about a detached head. This part of git is just beyond me; let me know when you're back online, and I can just give you a short snippet to plug in.

LuiseV:
@alvinwan Feel free to send me the snippet whenever you like. Thanks

alvinwan:
@LuiseV Was wondering: what types of objects can you get a CSV of composition scores for? An assignment? Just one submission? The ideal solution would be to accept a list of submission/backup IDs and generate a CSV from there. This way, we can place it in the submission search. If we do that, here are the changes I recommend:

- move the controller method to `SubmissionListCtrl`
- in `/static/partials/admin/submissions.list.html`, search for `<div class="box-tools" ...>` (line 34). After the other buttons in that div, add this - replacing callYourMethod with the proper method invocation:
```
<a href="#"><button class="btn btn-primary btn-sm" ng-click="callYourMethod();"> <i class="fa fa-download"></i>  Download</button></a>
```

LuiseV:
@alvinwan Currently, the composition score CSV is for an assignment object. So it automatically includes all students in the course that the assignment is under. Should it rather be for a list of submissions? @papajohn ?

alvinwan:
I think either works. Allowing a list of submissions - thus accessible by the search - would mean a more "feature-rich" CSV download (download all for assignment, all for a user, by date etc.) We should stick with what you've already done though, since we've got to deploy soon; no worries, I was just asking. Unless @papajohn says otherwise.

Sticking with what we have, I recommend moving the JS method to CourseAssignmentListCtrl, add a new table header like "Composition Scores", and inside the ng-repeat="assignment for assignments" block, add the download button I included in the last post. This will include your download button in the list of all assignments. The assignment details page is actually dysfunctional and useless, atm.

papajohn:
I think CSV export containing all of the most recent scores for a single
assignment is the most important use case (and I think that's what you
have). Let's not make it more complicated for now.

LuiseV:
This should be ready now @papajohn

server/app/models.py
```diff
@@ -1199,3 +1207,17 @@ def _can(cls, user, need, final, query):
     def _pre_put_hook(self):
         # TODO Remove when submitter is a computed property
         self.submitter = self.submission.get().submitter
+
+    def get_comp_score(self):
```
papajohn: I see two small issues with this design.

(1) I think it's easy to generalize this method to get all scores, instead of just composition. Remove the `score.tag == 'composition'` check and add a `tag` field to each response. So, this function should be called `get_scores` and it should return a bunch of tagged scores, each with a message and grader. 

(2) I'm confused why the `submitter` email is returned. When a student works with a partner, she may have a `FinalSubmission` for which she is not the submitter. That person still needs to be assigned a grade.

server/app/models.py
```diff
@@ -1199,3 +1207,17 @@ def _can(cls, user, need, final, query):
...
+        """ 
+        Return a list of the format [student, score, message, grader] 
+        if the submission has a composition score. Otherwise None. 
+        """
+        subm = self.submission.get()
+        comp_scores = [score for score in subm.score if score.tag == 'composition']
+
+        if len(comp_scores) > 0:
+            comp_score = comp_scores[0]
```
papajohn: Instead of choosing the first, it should choose the most recent.

server/app/models.py
```diff
@@ -1199,3 +1207,17 @@ def _can(cls, user, need, final, query):
...
+            return [self.submitter.get().email[0], \
```
papajohn: End-of-line slash is not necessary after a comma in a list literal.

server/app/seed/__init__.py
```diff
@@ -110,6 +110,11 @@ def make_seed_submission(assignment, submitter, final=False):
                 'submit': final
             }
 
+        score = models.Score(
```
papajohn: I'd add a tag, just so the data look like the ones in a real deployment.

server/app/seed/__init__.py
```diff
@@ -216,6 +243,16 @@ def make_queue(assignment, submissions, asignee):
         models.Participant.add_role(s.key, course.key, STUDENT_ROLE)
         students += [s]
 
+    # Luise
```
papajohn: I think this comment can go.

server/app/seed/__init__.py
```diff
@@ -216,6 +243,16 @@ def make_queue(assignment, submissions, asignee):
...
+    for i in range(0, 9):
```
papajohn: `range(9)`

server/static/js/admin/controllers.js
```diff
@@ -480,7 +480,7 @@ app.controller("CourseListCtrl", ['$scope', 'Course',
           $window.swal('Error', 'Could not delete assignment.', 'error')
          });
       }
-
+
```
papajohn: Please remove extra white space.

papajohn:
Looking great! We're going to put more scores than just composition in the system over the summer, so let's generalize a bit before merging. In particular, every project will have `composition` and `Partner 0` and `Partner 1` tagged scores. Rather than hard-coding the options, it would be great to have any tagged scores exported into the `csv` file and marked with their tag.

papajohn:
With the generalization, I expect this change will satisfy #484

alvinwan:
Oo, yeah. That's brilliant... @LuiseV. All you have to do is copy-and-paste the HTML button you used for composition score CSV downloads, on the front-end. If you need params for any reason, add the proper `web_arg` to the dictionary of methods and append to the query string to the API call. Ping me if you need help, but it looks like you've got the hang of it. Nice.

LuiseV:
@papajohn and @alvinwan,  I am currently working on fixing a bug in the downloads. Basically, the problem is that data_for_score in api.py loops over all students, and get_scores in the FinalSubmission model loops over all members in the current student's group. That means that Partner A and B will both be counted twice for instance - once as a student, and onces as a group member of the other. In data_for_scores, I now skip students that have already been seen in groups. It improves some parts but doesn't fix the case where Partner B is the submitter, but Partner A is handled first by the download methods. Partner A will be added to the CSV with an empty submission first, and then later with her real score, when Partner B's group is processed. I'm still trying to figure out how to get rid of the first write. 
It shouldn't be so that all assignments are forced to be either a group or individual assignment, right? That would simplify things. As far as I can see, however, there seems to only be a *max* group size for an assigment.

LuiseV:
@papajohn The bug is fixed, and score export should be working. Please let me know if there is anything you would like me to change!

server/app/models.py
```diff
@@ -1199,3 +1214,27 @@ def _can(cls, user, need, final, query):
...
+
+    def get_scores(self):
+        """ 
+        Return a list of lists of the format [[student, score, message, grader, tag]]
+        if the submission has been scored. Otherwise an empty list. 
+        If the submission is a group submission, there will be an element 
+        for each combination of student and score.
+        """
+        # TODO: get the most recent score for each tag. 
+        # Question: will all scores have a grader? In particular the scores from the autograder.
```
papajohn: I think we can't assume that every score will have a grader.

server/app/seed/__init__.py
```diff
@@ -264,10 +320,25 @@ def make_queue(assignment, submissions, asignee):
...
+    # Make team 1's submission final and score it. 
+    final3 = make_final_with_group(group3_subm, assign, group_members[5], g3)
+    score_seed_submission(final3, 1, "Awesome job, group 3!", staff[8]) 
+    subms.append(group3_subm)
+
     # Make this one be a final submission though.
-    subm = make_seed_submission(assign, group_members[1], True)
-    subm.put()
-    subms.append(subm)
+    # subm = make_seed_submission(assign, group_members[1], True)
```
papajohn: I think these commented-out lines can go.

papajohn:
Excellent! Merging.

server/app/models.py
LuiseV: Ok, then I will have to change this a bit. The current version breaks when there is no grader, but it should be an easy fix.

LuiseV:
Awesome. A few things I/we might want to do in the futues are to add an option to filter scores based on tags. I think that would be relatively simple with the current implementaiton. 
Also, in my current implementation, we get the scores for students and groups  by calling the methods scores_for_group_members and scores_for_student_or_group in the Assignment model. It might be more logical, if I implement these methods under the User and Group models respectively instead so score lookup would be based on students rather than assignments.

soumyabasu:
Can you make issues for these? They sound like really great ideas, and I don't want them to be lost.

LuiseV:
Sure, just did! I can work on both!

PR #458, opened 2015-06-20
Search Permissions

alvinwan:
Temporary fix for permissions issues with the search.
- must pass courseId with all queries
- check if current user is staff in course

Notes:
- doesn't mean search is restricted to course
- temporary fix

@Sumukh ready.


Sumukh:
I'm going to merge this - but in the future there should be tests for API related changes - especially when there are permissions invloved.

PR #460, opened 2015-06-20
Fixes for Student front-end

alvinwan:
- adds offering info to course info blob
- fixes glitchy hover

@Sumukh ready.


Sumukh:
Looks good. Merging.

We should add an active/inactive flag on courses so we filter inactive/not enrolled courses on the frontend

PR #472, opened 2015-06-21
Delete Assignment

alvinwan:
Changes
- new feature: delete assignment
- create assignment page based on course
- edit assignment page based on course
- both views return to course assignments page
- correct permissions checks and tests

@Sumukh ready.


server/app/api.py
```diff
@@ -757,6 +760,9 @@ def invite(self, obj, user, data):
         err = models.Group.invite_to_group(user.key, data['email'], obj.key)
         if err:
             raise BadValueError(err)
+        
+    def delete(self, obj, user, data):
+        obj.key.delete()
```
Sumukh: You should add a permissions check here.

server/static/js/admin/controllers.js
```diff
@@ -29,9 +29,23 @@ app.controller("AssignmentModuleController", ["$scope", "Assignment",
 
 app.controller("AssignmentListCtrl", ['$scope', '$http', 'Assignment',
   function($scope, $http, Assignment) {
+    $scope.reloadView = function() {
     Assignment.query(function(response) {
       $scope.assignments = response.results;
     });
+    }
+    $scope.delete = function(assign) {
+     Assignment.delete({
```
Sumukh: Perhaps have a warning to make sure. 
You can do that with something like this

```
$window.swal({
  title: "Are you sure?",
  text: "You will not be able to recover this imaginary file!",
  type: "warning",
  showCancelButton: true,
  confirmButtonColor: "#DD6B55",
  confirmButtonText: "Yes, delete it!",
  closeOnConfirm: false,
  html: false
}, function(){
/// Deletion code here. 
  swal("Deleted!",
  "Your imaginary file has been deleted.",
  "success");
});
```

server/static/js/admin/controllers.js
```diff
@@ -364,14 +378,28 @@ app.controller("CourseListCtrl", ['$scope', 'Course',
...
+    $scope.reloadView = function() {
+       Course.assignments({
+        id: $stateParams.courseId
+       },function(response) {
+         $scope.assignments = response
+       });
+     }
+
+     $scope.delete = function(assign) {
+       Assignment.delete({
```
Sumukh: Ideally we'd avoid repeated code - but it's ok. 

Same thing about the deletion (warning)

alvinwan: Good point. I'm going to delete the "All assignments" page and all links to it. Migrating towards course-centric admin now.

Sumukh:
Please add permissions + permissions tests to (tests/unittests/test_assignment.py).

The user need here is delete. You'll have to add the check to the API and the actual permissions needed to delete in models.py

server/app/api.py
alvinwan: That's right. I forgot you already told me to. x)

alvinwan:
@Sumukh It looks like delete already has tests and permissions checks, so I went back - looks like the BaseAPI has a delete method. However, this is only accessible to admin. I overrode it and allowed staff to delete as well.

Sumukh:
Did you make changes? I don't see any new commits. 

Also the model will still have to be changed. The base API works with the HTTP delete method - not with a POST.

alvinwan:
I just pushed, but wait, I'm calling a new delete method in the Assignment API. Am I supposed to call the default? The default only allows admin, and I don't want to change that.

soumyabasu:
I think we should relax the permissions then. The "owner" of an assignment
is the instructor for the course not an admin.

On Sunday, June 21, 2015, Alvin Wan <notifications@github.com> wrote:

> I just pushed, but wait, I'm calling a new delete method in the Assignment
> API. Am I supposed to call the default? The default only allows admin, and
> I don't want to change that.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Cal-CS-61A-Staff/ok/pull/472#issuecomment-113959955>.
>


-- 
Sent from Phone

Sumukh:
Looking good. 

You can actually get rid of the delete method (it's almost exactly the same as the BaseAPI method). The base method just defers to the models permissions. And now that the model accounts for delete permissions, we can just do that. 

Then you'd just make the angular service submit a DELETE request instead of POST.  

There should be a test to ensure an anon user cannot delete an assignment.

alvinwan:
Ohh. That's convenient. I'll call the default then. And, I'll add in the user test as well.

alvinwan:
@Sumukh I'm having some trouble. I'm using the Angular service's default delete method. However, server is responding "Method not allowed". Should I just leave the newly-added delete method?

Sumukh:
Can you push up the code? I'll take a look 
Do you have
```
        'delete': {
            'methods': set(['DELETE'])
        }
```
as a method in the methods list?

alvinwan:
haha Yeah, I eventually figured out I needed that. x) Thanks

alvinwan:
@Sumukh Ready to go. Added tests for anon users and all assignment views reoriented to course.

Sumukh:
Looks good. Left a minor comment - feel free to fix that if you want and I'll merge once travis comes in.

alvinwan:
Fixed. Why don't don't we use set literals though?

Sumukh:
I'm not sure why it was done that way. That being said, I prefer using `set([...])` rather than `{ }`

PR #473, opened 2015-06-21
Autograde UI [DO NOT MERGE]

alvinwan:
Sumukh's suggestions

- [x] Grade Final Submissions button -- *in course assignments next to each assignment*
- [x] Provide zip file URL with the starter code -- *part of assignment creation/edit*
- [x] Customize the grading script (which is just bash) here’s where you could download additional tests and do any additional setup. -- *part of assignment creation/edit*
- [ ] Hit done - you’ll be sent to a queue status page (here’s where you could cancel the queue, see the status etc) - the grades will pour in.

### DIFF
--- a/server/app/api.py
+++ b/server/app/api.py
@@ -663,7 +663,11 @@ class AssignmentAPI(APIResource):
                 'templates': Arg(str, use=lambda temps: json.dumps(temps),
                     required=True),
                 'revision': Arg(bool),
-                'lock_date': DateTimeArg()
+                'lock_date': DateTimeArg(),
+                'autograding_enabled': Arg(bool, default=True),
+                'grading_script_file': Arg(str),
+                'zip_file_url': Arg(str),
+                'access_token': Arg(str)
                 }
         },
         'put': {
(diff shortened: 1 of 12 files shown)


alvinwan:
@Sumukh Is there supposed to be one "autograde" button for all assignments? I placed on "auto grade" button for each assignment in the course assignments list.

albert12132:
An autograde button for individual submissions would be useful too, so that
staff members can facilitate regrade requests.

Sumukh:
In addition I need three/four fields on the assignment. 

- Autograding enabled (boolean)
- Grading Script file (as a string) 
- Zip File URL  (string)
- Access Token field (string)

Could you add these fields to the Assignment model and to the assignment creation/editing pages?

alvinwan:
@albert12132 Good idea. Will add.
@Sumukh On it.

alvinwan:
@Sumukh Is the grading script "file field" actually a string containing the file's contents?

Sumukh:
@alvinwan  - yes but its been implemented in the other branch. 

Closing because this is already in the other branch. Let's work in enhancement/sumukh/autograding_v2

alvinwan:
@Sumukh sounds good.

PR #483, opened 2015-06-26
Enrollment APi

alvinwan:
moved enrollment method to a "GeneralAPI" in api.py


soumyabasu:
Can we make a Participant API and have /enroll be a function of that?

PR #488, opened 2015-07-01
Student front-end fix

alvinwan:
assets issue with sweet-alert


Sumukh:
I think it has to go the other way (no dashes) with v1 of sweet alert.

alvinwan:
Ah I see. I'll fix the directory structure in your new branch.

PR #494, opened 2015-07-02
Generate composition grading queues

### LINKED ISSUE #441: Web interface to create queues for composition grading
I thought this button existed at one point, but couldn't find it just now... I think a natural place for such a button would be the staff-facing assignment-edit page.

alvinwan:
fixes #441 

basic checklist/'*' input:
generate-basic-queue

If you'd like to assign all final submissions to two misbehaving readers:
generate-csv-queue

old:
queue-generation

@Sumukh @soumyabasu Ready. This took a bit more code, so I'll wait until you guys give the green light, to merge.


alvinwan:
(for the wiki, ignore)

adin-tutorial-assignments
adin-tutorial-assignments2
adin-tutorial-assignments3

server/app/api.py
```diff
@@ -439,6 +439,18 @@ def enrollment(self):
                 })
         return json.dumps(data)
 
+    def check(self, emails, course, role):
```
Sumukh: What is this used for? The ok-client enrollment?

server/app/api.py
```diff
@@ -1987,6 +2010,72 @@ def new_entity(self, attributes):
...
+            subm.queue = queues[i].key
+            subm.put()
+            i = (i + 1) % len(staff)
+            
+        return queues
+
+    def check_permissions(self, user, data):
+        course = data['course'].get()
+
+        if user.key not in course.staff and not user.is_admin:
```
Sumukh: This method should probably return True/False and leave it up to the caller what to do.

server/app/api.py
```diff
@@ -1987,6 +2010,72 @@ def new_entity(self, attributes):
...
+                'course': KeyArg('Course', required=True),
+                'assignment': KeyArg('Assignment', required=True)
+            }
+        }
+    }
+    
+    def generate(self, user, data):
+        """ Splits up submissions among staff members """
+        
+        self.check_permissions(user, data)
```
Sumukh: See check_permission comment.

server/app/api.py
```diff
@@ -1987,6 +2010,72 @@ def new_entity(self, attributes):
...
+        course, assignment = data['course'].get(), data['assignment'].get()
+        userify = lambda parts: [part.user.get() for part in parts]
+        
+        staff = userify(models.Participant.query(
```
Sumukh: We don't necessarily want all staff grading composition. We'd need a selection of staff. The front-end could display a list of staff and the user could select the users to choose. 

Perhaps the create queues button goes a page `/assignId/queues/new` choosing what staff members get assigned. That page can then include the userIDs selected in it's post.

server/app/api.py
```diff
@@ -1987,6 +2010,72 @@ def new_entity(self, attributes):
...
+            models.Participant.role == STAFF_ROLE,
+            models.Participant.course == course.key).fetch())
+
+        students = userify(models.Participant.query(
+            models.Participant.role == STUDENT_ROLE,
+            models.Participant.course == course.key).fetch())
+
+        subms, queues = [], []
+
+        for student in students:
```
Sumukh: Wouldn't this double submit into a queue if a student is part of a group? I think it'd be better to pull from FinalSubmissions.

server/static/js/admin/controllers.js
```diff
@@ -223,6 +223,92 @@ app.controller("AssignmentEditCtrl", ["$scope", "$window", "$state", "$statePara
...
+            $window.swal('Error', 'Queues could not be generated.', 'error');
+        });
+    }
+    
+    $scope.reloadAssignment();
+    $scope.reloadQueues();
+}])
+
+// obsolete for now, may need for more composition queue functionality
+app.controller("AssignmentQueueGenerateCtrl", ["$scope", "$window", "$state", "$stateParams", "Assignment", "Course", "Queues",
```
Sumukh: Ah I see - this is for the future when we can assign to specific staff.

Sumukh:
Made some comments - still have to test locally.

server/app/api.py
alvinwan: Oh, this is for a theoretical "generate composition queues" based off of a set of user emails. I originally had users input a CSV, so I needed this check.

server/app/api.py
alvinwan: Mhm. Should I keep the old CSV version down below? (queue.generate.html) Or should I go with a set of checkboxes?

alvinwan:
@Sumukh Ready. Gif in initial post updated.

Issue #116, opened 2014-09-04
Handle issues gracefully on the frontend.


moowiz:
Right now, when the API errors on the front end, the exception is swallowed up. 

We should show this to the user.

alvinwan:
Error messages are now accessible with err.data.message (or response.data.message).

Here's an example of error message being passed successfully.

![error-messages-fix](https://cloud.githubusercontent.com/assets/2068077/8505272/675c384c-2196-11e5-81da-415542ada95f.gif)

PR #507, opened 2015-07-03
Error Message Clarifications

alvinwan:
With #116 now fixed, we can have more informative errors. Taken from a comment on that issue, here's an example of error messages now working:

error-messages-fix

For posterity, here's how error messages work:
1. Backend throws a `BadValueError(message)`.
2. That message is returned as a part of a JSON object's `message` field.
3. Frontend makes the entire JSON object accessible, if the status code is anything other than `200` (success).

That JSON object would be stored in any `response.data`. In short, a message thrown in a `BadValueError` is now accessible via `response.data.message`.


alvinwan:
@Sumukh Ready.

PR #521, opened 2015-07-07
Adds mark as final functionality

alvinwan:
mark-as-final

### DIFF
--- a/server/app/api.py
+++ b/server/app/api.py
@@ -2108,27 +2108,33 @@ class FinalSubmissionAPI(APIResource):
     The API resource for the Assignment Object
     """
     model = models.FinalSubmission
-
+    contains_entities = False
+    
     methods = {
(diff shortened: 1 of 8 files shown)


server/app/api.py
```diff
@@ -2108,31 +2108,48 @@ class FinalSubmissionAPI(APIResource):
...
     methods = {
         'get': {
         },
         'index': {
         },
-        'post': {
+        'mark_backup': {
+            'methods': set(['POST']),
             'web_args': {
-                'submission': KeyArg('Submission', required=True)
```
Sumukh: Wait - this may affect the client right?

alvinwan: For some reason, the front end 'submission' object was actually a backup. You have a point though. I'll restore the `post` method but leave this in.

PR #530, opened 2015-07-10
Cron Job config

alvinwan:
When deployed with GAE, script will launch a daily backup of all users and final submissions.

[Reference](https://cloud.google.com/appengine/articles/scheduled_backups)

@soumyabasu @Sumukh Ready.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ install:
   - pip install -r server_requirements.txt
 
 script:
-  - FLASK_CONF=TEST coverage run --source=server $CMD --sdk_location $GAE_SDK --quiet
+  - FLASK_CONF=TEST coverage run $CMD --sdk_location $GAE_SDK --quiet
 
 before_script:
   - wget https://storage.googleapis.com/appengine-sdks/featured/google_appengine_$GAE_VERSION.zip -o gae_sdk.zip
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ python:
   - "2.7"
 
 install:
+  - pip install pip --upgrade
   - pip install -r server_requirements.txt
 
 script:
--- a/cron.yaml
+++ b/cron.yaml
@@ -1,5 +1,5 @@
 cron:
 - description: Ok Daily Backup
   url: /_ah/datastore_admin/backup.create?name=2015DailyBackup&kind=Submissionv2&kind=Userv2&filesystem=gs&gs_bucket_name=ok_2015_backups
-  schedule: every 24 hours
+  schedule: every day 17:00
   target: ah-builtin-python-bundle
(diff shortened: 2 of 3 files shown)


Sumukh:
Backing up FinalSubmissions seems like a good idea as long as we keep in mind it doesn't contain the actual submission.

How expensive (time & $$$) wise is it to backup the submissions table.

alvinwan:
I realized old backups are for Submissionv2 and not FinalSubmission v2. Whoops. Changed.

Backups as recent as July 4 take a minute or less; start and end times on the datastore admin are identical. As for $$$, not sure, but atm, it definitely won't surpass our daily limit. The latest blobstore backup incurred no charge, and I can't find the GCS charge on the bill.

Sumukh:
BackupV2 is what actually contains student code - lets backup FinalSubmission and Submission daily and do Submissions a little more infrequently. 

It'd be great if we can schedule backups around 3-6 am PST since that's there's the least load.

soumyabasu:
Also, another thing to note is that this is an alpha feature, so I'm not 100% sure if we should deploy this in production just yet. I wouldn't mind just having this in a PR until they officially release this feature.

Another thing- do you guys know if the backups are smart backups? It'd be nice if we could backup the submissions table every 12 hours or so, but that's only feasible if it does it intelligently (only backing up keys that have changed from the old backup).

alvinwan:
@Sumukh Sounds good. FinalSubmissionsv2 added, and cron job scheduled for 5 p.m. every day.

@soumyabasu I could be wrong, but I think that the "Alpha release" notice is for the new Datastore Admin application and not the cron job configuration itself. Another [reference for the same cron job configuration](https://cloud.google.com/appengine/docs/python/config/cron) doesn't have an Alpha version warning. The only difference is that the second URL is for the old admin view. (I could be wrong) 

I think the backup is "dumb". It's just a backup script; nothing special about it that would suggest "diff backups". Although, that would be nice.

You make a good point though. Maybe we should have one weekly backup for Backupv2? The daily backups only backup Userv2, Submissionv2, and FinalSubmissionv2?

Sumukh:
Looks good. I changed the timestamp to be be 3AM local time. Merging and deploying.

Sumukh:
This is an alpha release of the backup script - that being said I'm willing to try it considering our current backup strategy. 

https://cloud.google.com/appengine/articles/scheduled_backups

>
Scheduled Backups
Alpha
This is an Alpha release of Datastore Administration. This feature might be changed in backward-incompatible ways and is not recommended for production use. It is not subject to any SLA or deprecation policy.

alvinwan:
Oops, sorry. Didn't see your latest comment when I deleted the branch.

My mistake, and I didn't see that last part either.... mhm.

PR #536, opened 2015-07-12
Status Checks

alvinwan:
remaining tests and modifications to `user.status` commented


Sumukh:
Changes to participant.status should also be removed. 

I'd be ok with just removing the code instead of commenting it out.

alvinwan:
*Status fields removed.

soumyabasu:
Is this good to go? It's causing our build to break.

alvinwan:
Yep, it's good to go if you two are good with it.

PR #537, opened 2015-07-12
Restores coveralls check

alvinwan:
by removing `--source` flag

@soumyabasu @Sumukh Can we merge this one? Since it's not a good idea to merge/deploy the cron job script, I decided to cherry-pick this commit out from that branch first.

*Travis is failing, because the commits removing `user.status` tests are not included.


Sumukh:
Sounds good

Issue #484, opened 2015-06-28
CSV export for autograded scores


albert12132:

Sumukh:
I tried to export scores from an assignment and Google App Engine cut the process off (it exceeded the timeout of 60 seconds). 

Perhaps we should perform this as a deferred task (https://cloud.google.com/appengine/articles/deferred)  and store/email the CSV somewhere (but probably not in the database)

cc: @LuiseV @alvinwan

LuiseV:
Yeah, I see how that could take too long. I'm without my laptop for the weekend, but can check out the deferred library when I get home, and then work on the issue

alvinwan:
Mhm. This could be a problem for queue generation and zip-file downloads too. I'll take a look.

Convert to deferred task:
- [ ] query download zip
- [ ] queue generation

papajohn:
You could just write it to Google Cloud Storage when someone asks for the
CSV, then provide a download link when it's ready.

https://cloud.google.com/appengine/docs/python/googlecloudstorageclient/

alvinwan:
@LuiseV The Google Cloud Storage API will require a bucket name; you can use `2015_backup`. Just created it.

LuiseV:
Thanks, @alvinwan .

LuiseV:
@alvinwan Do I have access to the bucket? I get "Failed to load" when I try to access it through https://console.developers.google.com/storage/2015_backup/. Is there a different way for me to access it? Or can you give me access for testing purposes?

Sumukh:
@LuiseV - I added your berkeley email to a new bucket created for grades (ok_grade_bucket)

 https://console.developers.google.com/project/ok-server/storage/browser/ok_grades_bucket/

-- 
Sumukh Sridhara

On July 9, 2015 at 6:37:41 PM, LuiseV (notifications@github.com) wrote:

@alvinwan Do I have access to the bucket? I get "Failed to load" when I try to access it through https://console.developers.google.com/storage/2015_backup/. Is there a different way for me to access it? Or can you give me access for testing purposes?

LuiseV:
Thanks, @Sumukh !

LuiseV:
@Sumukh , it says I don't have sufficient permissions to view the page when I follow your link. I am logged in with by Berkeley email.

Sumukh:
whoops - this link should work 

https://console.developers.google.com/storage/browser/ok_grades_bucket 

-- 
Sumukh Sridhara

[...]

PR #545, opened 2015-07-13
Score export to Cloud Storage - deferred task

LuiseV:
Files written to GCS in developer mode don't show up in the real bucket. The files can instead be accessed through https://localhost:8080/_ah/gcs/ok_grades_bucket/<file_name> when testing locally. 
Fixese #484


soumyabasu:
@alvinwan @Sumukh - can you guys take a look at this PR? I'm not too familiar with this feature.

alvinwan:
I'll check it out n the morning.

server/app/api.py
```diff
@@ -672,7 +672,7 @@ def get_backups(self, obj, user, data):
         return obj.get_backups(data['assignment'], data['quantity'])
 
     def get_submissions(self, obj, user, data):
-        return [subm.submission for subm in obj.get_submissions(data['assignment'], data['quantity'])]
+        return obj.get_submissions(data['assignment'], data['quantity'])
```
alvinwan: Mhm. This could actually break master. I made the correction above to fix student and admin bugs with submission views. Each submission has a `backup` property, so if you really need to, you should use `[subm.backup for sum in get_submissions(...)`.

server/app/api.py
```diff
@@ -1821,7 +1763,8 @@ def get_courses(self, course, user, data):
     def get_students(self, course, user, data):
         query = models.Participant.query(
             models.Participant.course == course.key,
-            models.Participant.role == 'student')
+            models.Participant.role == 'student',
+            models.Participant.status != 'inactive')
```
alvinwan: Try merging master in again? This is old code that can break ok (that was reverted).

server/app/utils.py
```diff
@@ -131,6 +133,51 @@ def create_csv(content):
...
+        gcs_file = gcs.open(gcs_filename, 'w', content_type='text/csv', options={'x-goog-acl':'project-private'})
+        gcs_file.write(contents)
+        gcs_file.close()
+    except Exception as e:
+        logging.exception("ERROR: {}".format(e))
+        try:
+            gcs.delete(gcs_filename)
+        except gcs.NotFoundError:
+            logging.info("Could not delete file " + gcs_filename)
+    logging.info("Created file " + gcs_filename)
```
alvinwan: Nice. Good job figuring out GCS.

server/app/api.py
LuiseV: @alvinwan These and the above are not changes I have made - they are probably from merges. When I merge with master, it says that my branch is already up-to-date though. Are the two places you pointed out the only two that should be changed? If so, I can do it manually.

alvinwan: Mhm. I was guessing something along those lines. There are actually quite a few; I can go through tomorrow, and pull them out one-by-one.

LuiseV: @alvinwan I really hope that there is some way of solving this that doesn't involve you going through it all. If I can do it, please let me know. I guess I can just look in the master repo on GitHub. 
I can se tons of people having the same issue with "Already up-to-date" even when all files aren't the same, but I still haven't found a solution that works. 
If it is just a question of whether you or I find the changes manually, let me know - then I'll do it tomorrow.

alvinwan: I'll make a first pass to remove incompatibilities; can you make a second pass to make sure that I don't break anything?

server/app/models.py
```diff
@@ -733,7 +744,6 @@ class Score(Base):
     score = ndb.IntegerProperty()
     message = ndb.TextProperty() # Plain text
     grader = ndb.KeyProperty(User) # For autograders, the user who authenticated
-    server_time = ndb.DateTimeProperty(auto_now_add=True)
```
alvinwan: @Sumukh @LuiseV Did one of you guys add this line? I've been reapplying old commits and am not sure if this line should be deleted.

server/app/models.py
```diff
@@ -1246,11 +1274,13 @@ def get_scores(self):
...
+            mem_email = member.get().email[0]
+            grader_email = None
+            
             for score in self.submission.get().score:
                 all_scores.append([email,
                         score.score,
                         score.message,
-                        score.grader.get().email[0],
+                        grader_email,
                         score.tag])
```
alvinwan: @Sumukh @LuiseV Ditto. See message above.

alvinwan:
@LuiseV I've finished reapplying old commits and making changes where needed. This is really great.

@Sumukh LGTM. Waiting for your review. Ignore Travis; I'll have to fix it in another PR.

server/app/models.py
Sumukh: I added this line - we should keep it.

alvinwan: Added. I'm assuming you made the changes specified in the next comment too. I'll re-apply those.

server/app/models.py
Sumukh: I don' think this was me - also this would seem to set the grader_email to none ?

Sumukh:
Alright - I'll take a look at this tonight. Thanks @LuiseV & @alvinwan ! :+1:

server/app/models.py
alvinwan: Yeah, not sure what the rationale behind it was. Not related to Luise's code, so I removed the change.

server/app/models.py
LuiseV: @alvinwan  I did not add this line.

server/app/models.py
LuiseV: @alvinwan  I added this because Prof. DeNero said we couldn't depend on autograded scores having a grader attribute. Now it seems like master has been changed so that also autograded scores have a grader. So my changes aren't necessary now, but make the code able to handle scores without grader if those might exist in the future.

LuiseV:
Thanks, @alvinwan and @Sumukh !

server/app/models.py
alvinwan: I see. That makes sense.

alvinwan:
Since I don't want to be the only unappreciative one on this thread...

Thanks @LuiseV and @Sumukh :P

server/app/constants.py
```diff
@@ -4,3 +4,4 @@
 STUDENT_ROLE = 'student'
 STAFF_ROLE = 'staff'
 VALID_ROLES = [STUDENT_ROLE, STAFF_ROLE]
+GRADES_BUCKET = 'ok_grades_bucket'
```
Sumukh: This eventually won't be a constant but probably a property of a class. This is fine for now.

Sumukh:
This looks good! I'm going to test more after the merge. Travis is failing but on a known issue (with a fix in #546) 

The diff looks pretty self-contained to exporting scores so we should be safe to deploy.

Sumukh:
Just deployed it and it worked for composition grades! :grinning: 

This is great! 

In the future we could have ok download the file programmatically and serve it so viewers don't need admin privileges. (or just send an email with it attached once it's done)

LuiseV:
Great!
I can definitely work on downloading and serving the file programatically. The only thing is that it might take a bit before the file is written, so it would be served with some dealy - which might not matter that much. Email is definitely an option as well! 
Another possibility is to just change the access permission. Right now it is project private which means that project owners and editors have owner permission ('read'+'write'), and team members have 'read'. Either class staff members could all be team members (perhaps not the best idea to mix OK and the class like that). It is also possible to give access to certain Google Group accounts or specified individual google accounts (could e.g. be all emails for users with 'STAFF_ROLE'). 
The easiest is probably just to serve it though :) 

Another thing to consider (not related to permission) is if we should really write the file every time. If the scores have been uploaded to gcs previously, we could have the download button give the user the option of rewriting the file or just downloading the existing one. In that way, you only have to write it again if scores have been updated.

PR #546, opened 2015-07-15
Fix Travis

alvinwan:
@soumyabasu @Sumukh Travis is dying because a fix for `mark_as_final` was never mirrored in the tests. It's a short and small update, cherrypicked from `cleanup/soumya/tests`.

We should merge to get that green Travis button back.


soumyabasu:
LGTM. Merging now.

Issue #523, opened 2015-07-08
Deferred Tasks


alvinwan:
Convert to deferred task:
~~- [ ] query download zip~~
~~- [ ] queue generation~~
~~- [x] CSV generation (@LuiseV)~~

to fix #484

LuiseV:
Also, score export should be on that list. 
@alvinwan  We will have the deferred task write to GCS, right?

PR #550, opened 2015-07-18
Deferred Tasks

alvinwan:
- [x] query download zip

~~- [ ] queue generation~~ (doesn't seem to be an issue atm)

#523 #484


alvinwan:
@Sumukh @soumyabasu Ready.

server/assign_queues.py
```diff
@@ -1,30 +0,0 @@
-from google.appengine.ext.remote_api import remote_api_stub
-from google.appengine.ext import ndb
-from app import models
-import getpass
-import sys
-import pickle
-
-def main():
```
Sumukh: :+1:  no more remote-api calls! :smile:

server/static/js/admin/controllers.js
```diff
@@ -533,6 +533,29 @@ app.controller("SubmissionListCtrl", ['$scope', '$stateParams', '$window', 'Sear
...
+        $window.swal({
+          title: 'Success',
+          text:'Saving submissions to ' + filename +
+          '\n Zip of submissions will be ready in Google Cloud Storage ok_grades_bucket in a few minutes',
+          type: 'success',
+          confirmButtonText: 'View zip',
+          cancelButtonText: 'Not now',
+          showCancelButton: true},
+          function() {
+            $window.location = 'https://console.developers.google.com/storage/browser/ok_grades_bucket/';
```
Sumukh: Could we just give them the direct url to the file and avoid redirecting them? Not everyone will have privileges to view the file. Either way not a big deal because we'll come up with a better way to proxy files through ok-server. 

I believe the format is : 
`https://console.developers.google.com/m/cloudstorage/b/ok_grades_bucket/o/{FILENAME}`

LuiseV: I think I looked into that when I did the score export. As far as I could tell, the above URL requires the user to have read access to the object (https://cloud.google.com/storage/docs/cloud-console#_downloadingdata under Accessing Google Developers Console). By default only project owners, editors and team members have read access. 
It might be worth trying out though!

alvinwan: I can add it in; it's actually just the filename appended to the URL in the JS.

Looks like Luise is right. For that matter though, only project owners, editors, and team members can access the exports atm.

server/assign_queues.py
alvinwan: xD

alvinwan:
@Sumukh Is this good to go?

Sumukh:
Sure.

PR #551, opened 2015-07-21
Fixes seed problesm

alvinwan:
`add_to_grading_queues` was acting up. It was just the difference between keys and non-keys.

@Sumukh @soumyabasu ready.


server/app/utils.py
```diff
@@ -319,9 +319,13 @@ def add_to_grading_queues(assign_key, cursor=None, num_updated=0):
     seen = set()
     for queue in queues:
         for subm in queue.submissions:
-            seen.add(subm.get().submitter.id())
+            try:
```
Sumukh: Instead of a try/except - it should be possible to see if subm is an instance of a particular class (models.Submission?) - (we do this occasionally in models.py/api.py) 

If that's not possible, then I guess a try/except is fine - it's more a style/preference thing to avoid try/excepts.

alvinwan: haha It's definitely possible to convert it; I can do it. On a side note, I thought Python was EAFP?

alvinwan: Fixed. The app was whining about a circular import, so I had to test for `ndb.Key` instead.

alvinwan:
@soumyabasu @Sumukh Is this good to go? I'll wait for you guys to merge, if it's good to go.

PR #552, opened 2015-07-21
Clarified partner reordering lock

alvinwan:
Wasn't clear why group members couldn't be reordered - clarified that (see gif)

ok-reorder-partners-lock

@soumyabasu @Sumukh REady.


Sumukh:
Front-end changes look good to me. It looks like this PR depends on #551 being merged first so once that's merged we can merge this.

Sumukh:
Looks good. The diff seems a bit out of date. @alvinwan

PR #555, opened 2015-07-26
Added Partner Info to Submission Download

alvinwan:
@Sumukh @soumyabasu Ready.

@jeffreylu017 Tada. (see gif below). Students without a group just won't have a `group.json`

get member info

### DIFF
--- a/server/app/utils.py
+++ b/server/app/utils.py
@@ -513,18 +513,10 @@ def scores_to_gcs(assignment, user):
 
 def add_subm_to_zip(subm, Submission, zipfile, result):
     """ Adds submission contents to a zipfile in-place, returns zipfile """
-    json_pretty = dict(sort_keys=True, indent=4, separators=(',', ': '))
     try:
         if isinstance(result, Submission):
             result = result.backup.get()
         name, file_contents = subm.data_for_zip(result)
-        group_files = backup_group_file(result, json_pretty)
-        if group_files:
-            for file in group_files:
-                add_to_file_contents(file_contents, *file)
-        add_to_file_contents(file_contents,
-                   'submission_meta.json',
-                   str(json.dumps(result.to_json(), **json_pretty)))
         return add_to_zip(zipfile, file_contents, name)
     except BadValueError as e:
         if str(e) != 'Submission has no contents to download':
--- a/server/app/utils.py
+++ b/server/app/utils.py
@@ -557,7 +557,7 @@ def backup_group_file(backup, json_pretty={}):
 
 def make_zip_filename(user, now):
     """ Makes zip filename: query_USER EMAIL_DATETIME.zip """
-    outlawed = [' ', '.', ':', '@']
+    outlawed = [' ', '.', ':', '/', '@']
     filename = '/{}/{}'.format(
         GRADES_BUCKET, 
         '%s_%s_%s' % (
--- a/server/app/utils.py
+++ b/server/app/utils.py
@@ -513,13 +513,18 @@ def scores_to_gcs(assignment, user):
 
 def add_subm_to_zip(subm, Submission, zipfile, result):
     """ Adds submission contents to a zipfile in-place, returns zipfile """
+    json_pretty = dict(sort_keys=True, indent=4, separators=(',', ': '))
     try:
         if isinstance(result, Submission):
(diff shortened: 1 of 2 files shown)


server/app/utils.py
```diff
@@ -517,12 +517,31 @@ def add_subm_to_zip(subm, Submission, zipfile, result):
...
+    file_contents[file_name] = file_content
+
+# TODO(Alvin): generalize, cleanup everything about zip
+def backup_group_file(backup):
+    """ Returns group information: group_[group ID], group JSON """
+    G = ModelProxy.Group
+    group = G.lookup(backup.submitter, backup.assignment)
+    if group:
+        json_data = group.to_json()
+        order = {chr(97+i): u['email'][0]
```
Sumukh: might want to leave a comment that the chr(97+idx) is to get A/B instead of 0/1.

alvinwan: Sure thing. Adding.

Sumukh:
What's the behavior on a bulk download? Going to be a group_members.json file in ever group on a bulk download?  

We might want to just standardize this into something like a meta.json that includes other information like who submitted it, what time it was submitted, partners in the group etc. If that's not much more work - I think it's worth it. 

It might also be useful to have this is in the exported grades. (but that can wait for later)

server/app/utils.py
```diff
@@ -517,12 +517,33 @@ def add_subm_to_zip(subm, Submission, zipfile, result):
...
+        # 97+i converts 0, 1, 2, 3 to ascii codes corresponding to a, b, c...
+        # chr converts ascii code to an ascii char
+        order = {chr(97+i): u['email'][0]
+                 for i, u in enumerate(json_data['member'])}
+        return 'group_members_%s.json' % group.key.id(), str(json.dumps(order))
+
+
 def make_zip_filename(user, now):
     """ Makes zip filename: query_USER EMAIL_DATETIME.zip """
     outlawed = [' ', '.', ':', '@']
```
Sumukh: '/' should also be outlawed right?

Sumukh:
@jeffreylu017 - was this what you had in mind/would work for you?

alvinwan:
`group_members.json` is present in any submission folder that has a group.

I originally did that, actually. I just spit out the entire JSON, but it's incredibly ugly and unreadable... if you say so though, it's an easy add. I'll add in a `group_meta.json` and a `submission_meta.json`.

Sumukh:
json.dump has the following optional arguments which make it a little prettier and easier to read `sort_keys=True, indent=4, separators=(',', ': ')`

alvinwan:
Oh, thanks. Shoulda read the docs.

soumyabasu:
I'm not 100% sure if we should convert to Partner A and Partner B since that's a 61A naming convention. We should just give back the group ordering in a nice way (I think 0/1 is fine actually) and the course staff should make the conversion into whatever they want.

alvinwan:
@soumyabasu Mhm. Fair enough; any group with more than 26 members would start to have gibberish for the keys in `group_members.json` anyhow.

server/app/utils.py
alvinwan: Oh, yes. Oops.

That'd only happen if someone had a `/` in his/her email address though. Added it anyhow.

alvinwan:
Final export results. @soumyabasu @Sumukh Are we good to go?

metas

jeffreylu017:
Sorry I'm late to the party.

Will this work if I export this using the OK client (`python3 ok --export --latest`)?  I'm unable to unzip the file downloaded from the website, so I've been downloading submissions through the terminal.

The json file looks great.  I'm flexible with any format, and I can just grep the information I need.

alvinwan:
Let me check; not sure what endpoint the client uses.

alvinwan:
@jeffreylu017 For download, you'll have to use Safari for now. Not sure why it downloads differently, sorry.

`python3 ok --export --latest` doesn't call the search download endpoint, so it won't include these meta files.

alvinwan:
@albert12132 @jeffreylu017 Once sumukh/soumya approve and merge in the two PRs I made, zip downloads will work on all browsers... this entire time, I forgot to `zipfile.close()`.

PR #556, opened 2015-07-27
Fixed zip download

alvinwan:
one-liner

@soumyabasu @Sumukh Ready.

Zip downloads weren't working... because I didn't close... I removed the `with` block at some point and that broke things.


Sumukh:
Whoops. Accidentally closed it

LGTM.

PR #563, opened 2015-08-03
deferred invokes self if timeout

### LINKED ISSUE #557: Submission Downloads too big
Jeffrey

> When trying to download OK submissions from a search query, I just get the message:
>Success

>Saving submissions to /ok_grades_bucket/query_greynotgray-berkeley-edu_2015-07-31-21-18-53-167770.zip
Zip of submissions will be ready in Google Cloud Storage ok_grades_bucket in a few minutes

>Along with a link to view the zip (https://console.developers.google.com/storage/browser/ok_grades_bucket/query_greynotgray-berkeley-edu_2015-07-31-21-18-53-167770.zip).  But after a few minutes, when I click the link, I just get a message that just says "Failed to load".  Although for Albert, the message he gets is "no objects in this bucket".  I think for me, I might be having a permissions error, but we don't have a way to download submissions for project 3.

Sumukh

>I remember seeing a timeout error on the deferred write to the bucket in the logs. That seems like it could be the culprit. 

Idea: split up zip generation into subprocesses, download actually spawns one task per page of download

alvinwan:
I'm not actually sure if this will fix the problem

@Sumukh @soumyabasu Could you guys take a look? At some point or another, I can integrate with `mapper.py` and somehow generalize `mapper.py` for more than analytics.

Until then, the function that writes to gcs will invoke itself on timeout, while it's querying the database and writing to the string buffer. Caveat: if it times out while writing contents to a zip file... it dies.

fixes #557

### DIFF
--- a/server/app/api.py
+++ b/server/app/api.py
@@ -1433,7 +1433,8 @@ class SearchAPI(APIResource):
         self.check_permissions(user, data)
 
         now = datetime.datetime.now()
-        deferred.defer(subms_to_gcs, SearchAPI, SubmissionAPI, models.Submission, user, data, now)
+        deferred.defer(subms_to_gcs, SearchAPI, SubmissionAPI, 
+                       models.Submission, user, data, now)
         
         return [make_zip_filename(user, now)]
 
--- a/server/app/utils.py
+++ b/server/app/utils.py
@@ -26,6 +26,7 @@ import cloudstorage as gcs
 from app import app
 from app.constants import GRADES_BUCKET
 from app.exceptions import BadValueError
+from google.appengine.runtime import DeadlineExceededError
 
 # TODO Looks like this can be removed just by relocating parse_date
 # To deal with circular imports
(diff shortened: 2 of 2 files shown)


Sumukh:
Hmm. Yeah - I'm not convinced this will work either. 

How about instead we use the autograder server for this. The autograder server would just receive just a list of submission/backup id's and the autograder can go in and generate a ZIP using the API. OK already does a similar task to perform autograding - so this would be autograding without the grading. 

I'm a little concerned with bringing OK features into the autograder but I think there's a way that this can be done cleanly.

The way we did this previously was run a script locally - we'd essentially be automating the scripts.

alvinwan:
Sure, I took a look at the code but am still not sure how you managed to run the autograder without timeout errors. Can I leave this to you?

Sumukh:
I'm going to close this PR and open up a new issue to support exporting via the auto grading. This will involve work on the ok server and on the autograder side.

alvinwan:
Sounds good.

PR #568, opened 2015-08-11
Fixes query issue

alvinwan:
using cursors

@Sumukh ready.


server/app/utils.py
```diff
@@ -563,16 +564,29 @@ def make_zip_filename(user, now):
...
-    for result in results:
-        zipfile = add_subm_to_zip(subm, Submission, zipfile, result)
+    subm, i = SubmissionAPI(), 0
+    for result in query:
+        zipfile, i = add_subm_to_zip(subm, Submission, zipfile, result), i + 1
+        if i == constants.BATCH_SIZE:
+            start_cursor = query.cursor()
+            deferred.defer(subms_to_gcs, SearchAPI, SubmissionAPI, Submission,
+                           user, data, datetime, start_cursor)
+            return
```
Sumukh: Ok - it seems like this function will get `BATCH_SIZE` submissions from a query and then when it hits 25 - it will create a new deferred job with the cursor. I think this can work - but we'd be creating a lot of cursors.

Sumukh:
It seems like this will work. I do want to run another idea by you and get your thoughts: 

Since the deferred job has 10 minutes to run (which seems like plenty) but queries have < 1 minute or so - we should be able to keep it all inside one deferred job - as long as we handle queries well. 

We'd run the query like so 
` results, next_curs, has_more = Search.query(data).fetch_page(25)`

All we would do have to do is something along the lines of:
```
has_more = True 
while has_more:
   results, next_curs, has_more = Search.query(data).fetch_page(25)
   for res in results:
     # add to zip 
```

There might even be a way to do it with the iter method. (https://cloud.google.com/appengine/docs/python/ndb/queries#iterators) 

Let me know if you want to try something like that or just want to run with the existing approach of creating more deferred jobs. I'd prefer to try to just keep it in one deferred job - but it's your call.

alvinwan:
Oo I see. Sorry, misunderstood you the first time around. Sure.

alvinwan:
@Sumukh ready.

server/app/utils.py
```diff
@@ -563,16 +564,17 @@ def make_zip_filename(user, now):
...
-    """
-    results = SearchAPI.results(data)
+def subms_to_gcs(SearchAPI, SubmissionAPI, Submission, user, data, datetime,
+                start_cursor=None):
+    """Writes all submissions for a given search query to a GCS zip file."""
     zipfile_str, zipfile = start_zip()
-    subm = SubmissionAPI()
-    for result in results:
-        zipfile = add_subm_to_zip(subm, Submission, zipfile, result)
+    subm, has_more, next_cursor = SubmissionAPI(), True, None
```
Sumukh: Super Nitpicky (feel free to ignore) - the ordering of the variables should probably be consistent between 571 & 574

Sumukh:
Looks good - will merge/deploy/test on the same query soon.

server/app/utils.py
alvinwan: xD will do.

PR #569, opened 2015-08-13
hide sensitive info from json

Sumukh:
Hide autograder fields from json responses. I thought this was already implemented but apparently not. 

This shouldn't break anything because ok sends the relevant information to the autograder - not the other way around.

### DIFF
--- a/server/app/api.py
+++ b/server/app/api.py
@@ -2124,4 +2124,3 @@ class AnalyticsAPI(APIResource):
         return (201, 'success', {
             'key': job.job_dump.key.id()
         })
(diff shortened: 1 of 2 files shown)


Sumukh:
Merging to deploy this so future autograders are kept hidden.

As usual, feel free to comment. 

This approach isn't the best to protect sensitive fields and I'm working to make it the autograder need only one field that doesn't have to be hidden. 

In the future I think the only way to truly keep something secret is to keep it in the datastore and use permissions or not expose it programmatically.

PR #576, opened 2015-08-30
New Autograder Format 

Sumukh:
This PR makes autograding simpler from OK's perspective - it just has to store a key for the assignment (provided by the autograding service) 

I will merge this in when I get continuous autograding working.


Sumukh:
Merging. It now supports autograding upon submit (and emailing notifications).

PR #579, opened 2015-08-31
Assignment Detail

alvinwan:
- adds assignment statistics on an assignment detail view in the admin panel
- clicking on the assignment name in the list of assignments now leads to assignment detail

### DIFF
--- a/server/app/api.py
+++ b/server/app/api.py
@@ -795,10 +795,7 @@ class AssignmentAPI(APIResource):
         'queues': {
             'methods': set(['GET']),
         },
-        'submissions': {
-            'web_args': {
-                'final': BooleanArg()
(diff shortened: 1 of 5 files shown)


server/app/api.py
```diff
@@ -889,6 +891,14 @@ def queues(self, obj, user, data):
 
         return models.Queue.query(models.Queue.assignment == obj.key).fetch()
 
+    def statistics(self, assignment, user, data):
+        """Returns assignment statistics"""
```
Sumukh: How are permissions handled? 
We could add permissions checking here - Need('Staff') (see line 888 above)

alvinwan: Just added permissions check.

Issue #575, opened 2015-08-28
User API returns too much data


knrafto:
e.g. GET https://ok-server.appspot.com/api/v1/user/knrafto@berkeley.edu?course=5165212546105344&format=json takes around 13s and returns 25kB of data

Sumukh:
:+1: - This would be a big performance win for students. 
Here's the trace from app engine 
![image](https://cloud.githubusercontent.com/assets/882381/9559025/b1e6bb3a-4da1-11e5-8a99-e6bdf3a8bb05.png)

knrafto:
That doesn't seem right, that says it took only 97ms

Sumukh:
Yeah - I was a little surprised. This one is sounds more realistic (Note the 300+ RPCs!) 

![image](https://cloud.githubusercontent.com/assets/882381/9559057/5e11da3e-4da2-11e5-9ff9-a1cffe312804.png)

knrafto:
v2 is no longer being maintained.

PR #583, opened 2015-09-03
Bug/knrafto/optimize get course info

knrafto:
Tried to simplify some queries and parallelize using ndb tasklets. ~2x faster, ~2x less DB queries. Related to #575.

The local server now has a lot more assignments and submissions so I could get a good benchmark.

Point of contention: All submissions are now treated as submissions, rather than only submissions with files. I'm not clear on when a submissions wouldn't have files, or what happens when they don't. But it doesn't simplify things a whole lot.

#### Before
before cold
#### After
after cold
#### Before
before warm
#### After
after warm

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,9 +44,12 @@ To run the server locally:
 
 ```bash
 $ source activate_server.sh 
+$ cd server
 $ bower install  # to install frontend CSS/JS libraries
 $ ./start_server
 ```
+
+The server will listen on http://localhost:8080.
 
 Deploying
 ---------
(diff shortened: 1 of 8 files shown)


Sumukh:
:tada: This looks great. Thanks for making the PR. I'll take a look and will try to figure out a way to test this with production data (just to make sure there aren't weird issues with the old data)

knrafto:
Can you benchmark on the remote servers too? It's probably way different from doing it locally.

Sumukh:
I deployed this to staging.  https://staging-dot-ok-server.appspot.com

```
no matching index found.
The suggested index for this query is:
- kind: Assignmentv2
properties:
- name: course
- name: due_date
direction: desc
``` 

This should be fixable with a new index. Attempting a fix now.

Sumukh:
Great. Going to merge to master and deploy.

PR #591, opened 2015-09-09
autograder with support for auto-promotion, running in test mode

Sumukh:
... hide autograder errors from students on the frontend

Do not merge.

### DIFF
--- a/server/app/api.py
+++ b/server/app/api.py
@@ -683,7 +683,7 @@ class UserAPI(APIResource):
         return obj.get_backups(data['assignment'], data['quantity'])
 
     def get_submissions(self, obj, user, data):
-        return [subm.submission for subm in obj.get_submissions(data['assignment'], data['quantity'])]
+        return obj.get_submissions(data['assignment'], data['quantity'])
 
     def merge_user(self, obj, user, data):
         """
--- a/server/app/api.py
+++ b/server/app/api.py
@@ -40,7 +40,7 @@ from app.constants import STUDENT_ROLE, STAFF_ROLE, API_PREFIX, AUTOGRADER_URL
 from app import models, app, analytics, utils
 from app.needs import Need
 from app.utils import paginate, filter_query, create_zip, add_to_zip, start_zip, finish_zip, scores_to_gcs, subms_to_gcs, make_zip_filename, submit_to_ag
-from app.utils import add_to_grading_queues, parse_date, assign_submission
+from app.utils import assign_staff_to_queues, parse_date, assign_submission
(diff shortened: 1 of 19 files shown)


server/app/utils.py
```diff
@@ -609,28 +609,74 @@ def submit_to_ag(assignment, messages, submitter):
...
-      subm_ids[fsub.submission.id()] = fsub.submission.get().backup.id()
-
-    data = {
+def autograde_subms(assignment, user, data, subm_ids):
+    ag_data = {
         'subm_ids': subm_ids,
         'assignment': assignment.autograding_key,
-        'access_token': data['token']
+        'access_token': data['token'],
+        'testing': data['testing']
```
knrafto: It looks like this `testing` flag is not actually used anywhere...

server/app/api.py
```diff
@@ -790,6 +791,8 @@ class AssignmentAPI(APIResource):
             'methods': set(['POST']),
             'web_args': {
                 'grade_final': BooleanArg(),
+                'testing': BooleanArg(),
```
knrafto: It looks like this testing flag is not actually used anywhere...

server/app/api.py
```diff
@@ -865,6 +868,11 @@ def autograde(self, obj, user, data):
       if 'grade_final' in data and data['grade_final']:
         #Collect all final submissions and run grades.
         deferred.defer(autograde_final_subs, obj, user, data)
+    
+        if 'promote_backups' in data and data['promote_backups']:
+            # Force promote backups and run autograder
+            defrred.defer(promote_student_backups, obj, True, user, data)
```
knrafto: Typo `defrred`

server/app/api.py
```diff
@@ -865,6 +868,11 @@ def autograde(self, obj, user, data):
...
+
```
knrafto: Should promoting backups happen before the autograding starts?

knrafto: Oh, promote_student_backups also runs the autograder on those backups. It may be cleaner to separate this into two steps: promotion and grading

server/app/api.py
Sumukh: Indeed. At some point I'll add this to this flag to the frontend (the flag tells the autograder to not send grades back)

Actually, we should just have a hide grades feature.

server/app/api.py
Sumukh: I agree. Ideally they'd be two different endpoints. I'll file an issue for this later - but for now I'm going to run with this design. Autopromotion for now is something that is tied into autograding.

PR #592, opened 2015-09-12
defer queueing, remove old queue endpoint

Sumukh:


Sumukh:
Everywhere where we have a query over FinalSubmissions we have to do a deferred job - since that query alone takes more than what appengine provides for a deadline. 

It would be nice to have a `@deferred` decorator or something like it (combined with the permissions decorator).

Sumukh:
Merging despite the decreased test coverage  :disappointed: - so that we can generate queues.

PR #601, opened 2015-09-30
Cleanup zip and csv downloads and exports

knrafto:
Tested in staging:
* Exporting all Hog submissions
* Exporting scores for Hog
* Downloading my own submission for Maps

### DIFF
--- a/server/app/api.py
+++ b/server/app/api.py
@@ -43,7 +43,7 @@ from app.needs import Need
 from app.utils import paginate, filter_query, add_to_zip
 from app.utils import scores_to_gcs, subms_to_gcs, make_zip_filename, submit_to_ag
 from app.utils import add_to_grading_queues, parse_date, assign_submission, assign_staff_to_queues
-from app.utils import merge_user, backup_group_file, add_to_file_contents
+from app.utils import merge_user, backup_group_file
 from app.utils import autograde_final_subs, promote_student_backups
 
 from app.exceptions import *
--- a/server/app/api.py
+++ b/server/app/api.py
@@ -1009,7 +1009,7 @@ class SubmissionAPI(APIResource):
         try:
             user = obj.submitter.get()
             name = user.email[0]+'-'+str(obj.created)
-        except IndexError, AttributeError:
+        except (IndexError, AttributeError):
             name = str(obj.created)
 
         name = name.replace('.', '-').replace(' ', '_')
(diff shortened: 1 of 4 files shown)


server/app/api.py
```diff
@@ -1071,7 +1040,34 @@ def download(self, obj, user, data):
...
+            name, files = self.data_for_zip(obj)
+            write = start_response('200 OK', [
+                ('Content-Disposition', 'attachment; filename=submission-%s.zip' % name),
+                ('Content-Type', 'application/zip')
+            ])
+            stream = ResponseFile(write)
+            with zf.ZipFile(stream, 'w') as zipfile:
+                add_to_zip(zipfile, files)
+            return ''
+        return make_response(zip_response)
```
knrafto: I don't actually know that this streams the response instead of buffering it all in memory. Will investigate

knrafto:
Now `-onlyfinal` works too :)

Sumukh:
Cool - I'll take a look. I was under the impression that onlyfinal was working before?

Sumukh:
Another super useful feature is something to proxy reads/writes to google storage. (aka let non-admins access files in the bucket. AFAIK this is possible by generating a signed url from the cloud storage api - but I'm not sure how it works.

knrafto:
This cleanup was happening concurrently with the other PR, and I guess they diverged a bit.
I'll create an issue for the proxying

server/app/api.py
knrafto: I think it streams: https://www.python.org/dev/peps/pep-0333/#the-write-callable is under the section "Buffering and Streaming"

Sumukh:
Did you mean `-onlybackup`? I think all it needs is an index 
```
- kind: Backupv2
properties:
- name: assignment
- name: server_time
direction: desc
```

Sumukh:
NVM. Just updated the indexes and downloading/searching by onlybackups works.

knrafto:
So, a search query could return a FinalSubmission, a Submission, or a Backup, depending on what flags are used. Are there any others?

knrafto:
`-onlybackup` works? This should still run into issue #604

knrafto:
Unfortunately, downloading *all* Hog backups OOMs :( oh well

Sumukh:
Closing due to master being on v3

PR #607, opened 2015-10-07
Lock Date Revision Enhancements + AG single submission

### LINKED ISSUE #606: Submission search table should handle final submission
It looks like the submission search isn't built to handle final submissions so most of the features of the table will fail when -onlyfinal is True

Sumukh:
Prevent from students from submitting something well beyond the lock date and fix #606. Submission search now searches FinalSubmissions. We can also now autograde single submissions


Sumukh:
Tests passed (and new tests were added). Just successfully performed a regrade using the autograder & ok.

PR #614, opened 2015-11-01
Fix broken links in assignments and README

jvperrin:
Some links are broken when they are supposed to be external links but are instead interpreted as relative links. For instance on the most recent homework, the link in the assignment detail (shown below) is interpreted as going to `https://ok-server.appspot.com/cs61a.org/hw/hw08`, where it should go to `http://cs61a.org/hw/hw08` instead.

ok-broken-link

To fix this, I just added a function to check if a url string starts with `http`, and if it does, it is already absolute, but if not, `http://` is added so that the link becomes absolute. I also made some small changes to fix a broken link in the README with the same relative/absolute problem and specify that the Google App Engine SDK to download is for Python.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Installation
 
 To install:
 * Clone this repo
-* Install [Google App Engine SDK](https://developers.google.com/appengine/downloads) and add it to your `$PATH`. You can do `brew install google-app-engine` on a mac.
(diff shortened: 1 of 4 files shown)


Sumukh:
Looks good to me. Thanks!

jvperrin:
Thanks!

Sumukh:
FYI: Was looking at the logs and saw a lot of frontend errors because `url` was null. I've pushed a quick fix in 1d9dad9923cb1d708d91c3d4105f7af18059c002

PR #615, opened 2015-11-01
Fix submission score link styling

jvperrin:
I added an underline and a pointer cursor to the submission score links to make it more obvious how to see submission feedback. It turns out that if a link element is missing the `href` attribute, it doesn't change the cursor, but if it has a `href` it will change it, so I added a javascript `href` that does nothing to get the cursor to behave properly. Also, Angular doesn't like block elements within paragraph tags, so I had to remove the div from the original paragraph tag to get it to be positioned correctly.

Here's the end result:

ok-py-submission-links


alvinwan:
@jvperrin LGTM. Merging.

PR #620, opened 2015-11-07
Fix DST. 

Sumukh:
"Fix DST" is a bold statement - but it should be better than what we had before which was a manual offset by 8 hours. It normalizes times on API (both incoming/outgoing times). 

This will require all ok-server developers to update pytz - I've update the requirements. 

`pip install --update pytz`

### DIFF
--- a/server/app/models.py
+++ b/server/app/models.py
@@ -27,9 +27,10 @@ Specification: https://github.com/Cal-CS-61A-Staff/ok/wiki/Models
 import datetime
 import itertools
 import logging
+import pytz
 
 from app import app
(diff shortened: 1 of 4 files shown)


server/app/models.py
```diff
@@ -58,15 +59,11 @@ def default(self, obj):
 
 app.json_encoder = JSONEncoder
 
-def convert_timezone(utc_dt):
-    """Convert times to Pacific time."""
-    # This looks like a hack... is it even right? What about daylight savings?
```
Sumukh: :fearful:

papajohn:
Thanks for fixing!

PR #621, opened 2015-11-07
Fix DST issues (Potentially)

Sumukh:
Being a little less bold with the claim on this PR (see #620) 

Deployed to staging and it seems to indeed fix DST related issues (for the frontend/backend). We use timezone comparisons a lot and I think I've hit most of them with this new logic. Autograding should be accounted for as well as it uses the submission logic 

If anyone has better ideas I'd love to hear them.

In the meantime, I'm just going to leave this here http://gizmodo.com/why-daylight-saving-time-is-pointless-5892438

### DIFF
--- a/server/app/utils.py
+++ b/server/app/utils.py
@@ -12,6 +12,7 @@ import itertools
 from os import path
 from app import constants
 import requests
+import pytz
 
 try:
     from cStringIO import StringIO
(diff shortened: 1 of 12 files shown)


server/app/utils.py
```diff
@@ -49,6 +50,8 @@ def parse_date(date):
         date = datetime.datetime.strptime(
             date, "%Y-%m-%d %H:%M:%S")
 
+    logging.debug("utils.parse_date being used")
```
Sumukh: I'm not sure if this method is called anywhere - it doesn't look like it - but it is imported in places.

Sumukh:
Merging so that we have this for future assignments - but it'd be great to have this be code reviewed.

PR #622, opened 2015-11-11
Add Revision ID and Submission ID to grade export

Sumukh:
Grade Export now includes the submission ID and revision ID

(this is the actual ID of the submission, not the backup)


Sumukh:
I forgot to change these, just added them in 2947b36
```
        if not member:
          return [["Unknown-"+str(self.member[0]), 0, None, None, None, None, None]]

        return [[member.email[0], 0, None, None, None, None, None]]
```

PR #625, opened 2015-11-18
Auto Promotion of Backups

Sumukh:
This fix has been deployed onto staging and seems to work. 

TODO:
 - Promote more than just the last backup. Attempt to hunt for the backup closest to the due-date. 
 - Lower Priority: Also consider submissions first if any exist (to handle edge cases where lock_dates/due_dates where off) 
- Write tests for auto promotion (coveralls fell below 89%)

### DIFF
--- a/server/app/utils.py
+++ b/server/app/utils.py
@@ -696,9 +696,12 @@ def promote_student_backups(assignment, autograde=False, user=None, data=None):
         fs = student.get_final_submission(assignment)
         if not fs:
             recent_bcks = student.get_backups(assignment.key, 1)
+            # TODO: Also get submissions that weren't marked as final for some reason
             if len(recent_bcks) > 0:
                 try:
-                    new_sub = force_promote_backup(recent_bcks[0].key.id())
+                    # TODO: Get a bunch of backups and choose to closest to due date
+                    closest_backup = recent_bcks[0]
+                    new_sub = force_promote_backup(closest_backup.key.id())
                     newly_created_fs.append(new_sub.id())
                 except ValueError as e:
                     error_msg = "{email} only had backups past the due date".format(email=student.email[0])
--- a/server/app/utils.py
+++ b/server/app/utils.py
@@ -693,16 +693,18 @@ def promote_student_backups(assignment, autograde=False, user=None, data=None):
     newly_created_fs = []
     for participant in enrollment:
         student = participant.user.get()
-        fs = student.get_final_submission(assignment)
+        fs = student.get_final_submission(assignment.key)
+        group = student.get_group(assignment.key)
         if not fs:
-            backup_q = student.backups(fs.group, assignment)
-            elgible_backups = backup_q.filter(Backup.server_time <= assignment.due_date)
+            backup_q = student.backups(group, assignment.key)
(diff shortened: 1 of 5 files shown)


server/app/utils.py
```diff
@@ -695,9 +695,18 @@ def promote_student_backups(assignment, autograde=False, user=None, data=None):
...
+            recent_bcks = student.get_backups(assignment.key, 1)
+            # TODO: Also get submissions that weren't marked as final for some reason
+            if len(recent_bcks) > 0:
+                try:
+                    # TODO: Get a bunch of backups and choose to closest to due date
+                    closest_backup = recent_bcks[0]
+                    new_sub = force_promote_backup(closest_backup.key.id())
+                    newly_created_fs.append(new_sub.id())
+                except ValueError as e:
+                    error_msg = "{email} only had backups past the due date".format(email=student.email[0])
```
Sumukh: Currently a more accurate message is that the most recent backup is past the due date.

knrafto:
Can you add a `WHERE time < due_date` to get the most recent backup before the due date?

knrafto:
```
group = student.get_group(assignment)
# will be a Backup or None
student.backups(group, assignment).filter(Backup.client_time <= assignment.due_date).get()
```

Sumukh:
We can indeed do that & that's probably what's going to happen. There just wasn't a method in the objects for it in the code. 

Lock Dates add a bit of complication - but I think it's totally fair to say that in the absence of a Final Submission - auto promotion chooses the final backup before the deadline.

knrafto:
I added the `User.backups(group, assignment)` method when implementing the `get_course_info` stuff. It returns an NDB query, so you can add filters and `get` instead of `fetch`

Sumukh:
Ah - get_backups is using User.backups but I didn't think about just using backups directly. Thanks

knrafto:
`raise ValueError("Cannot change submission after due date")` looks like it'll get thrown no matter what if we try to promote backups after the deadline

The other two `raise ValueError`s should never happen (i.e. getting them would be a bug, so you shouldn't catch that exception)

Sumukh:
This code in models.py is also used for --submit which is why there is code for replacing an existing submission as well.  

As for the last exception - that's intended behavior (especially since this method is used for the mark as final button on the site and --submit) 

We could create another method I suppose just for promoting an arbitrary backup. (This could also generalize into a promote backup feature). 

As a more general note: I'm not a huge fan of placing a lot of our logic into the user model but that's for another time.

knrafto:
https://github.com/Cal-CS-61A-Staff/ok/pull/625/files#diff-bb50eb2e4a13c0bf26e75a3a8442adb2R704
So it looks like if a ValueError gets raised in the try block here, it indicates a bug, so you shouldn't try to catch the exception here.

Is `promote_student_backups` idempotent?

Sumukh:
@knrafto - I see - I've removed the try/catch (we needed it before when we just got an arbitrary backup - but no longer) 

Aside from tests - is there anything you'd like to see made in this PR before merging?

knrafto:
LGTM

Sumukh:
Thanks. Merging despite coveralls failure (will add tests soon!)

PR #627, opened 2015-11-24
Get All Scores/Handle Partners in Grade Export

Sumukh:
Grade export should get scores from all submissions (mainly for Scheme EC) but this could help in cases where students change the submission after the grade has been inputted. 

Partner Scores are limited to Final Submissions only at the moment (it made sense to put that logic there versus elsewhere)

### DIFF
--- a/server/app/models.py
+++ b/server/app/models.py
@@ -1321,8 +1321,11 @@ class FinalSubmission(Base):
                   # Individual group:
                   if group_size == 1 and is_partnered:
                         # Score is an average of all partnered parts of project
-                        other_partner_parts = [sc for sc in subm_scores if 'partner' in sc[4]]
-                        average_score = int(sum([s[1] for s in other_partner_parts])/len(other_partner_parts))
+                        other_partner_parts = [sc for sc in subm_scores if 'partner' in sc[4].lower()]
+                        if not other_partner_parts:
+                            average_score = score[1]
+                        else:
+                            average_score = int(sum([s[1] for s in other_partner_parts])/len(other_partner_parts))
                         score[1] = average_score
                   elif is_partnered:
                         expected_letter = chr(97 + position) # 0 -> a, etc
--- a/server/app/models.py
+++ b/server/app/models.py
@@ -1325,7 +1325,6 @@ class FinalSubmission(Base):
                         average_score = int(sum([s[1] for s in other_partner_parts])/len(other_partner_parts))
                         score[1] = average_score
                   elif is_partnered:
-                        # Only supports two member groups for now.
                         expected_letter = chr(97 + position) # 0 -> a, etc
                         if score[4].lower().strip() != "partner {}".format(expected_letter):
                             continue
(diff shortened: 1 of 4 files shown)


Sumukh:
Test Coverage (Coveralls) is back in the green!

server/app/models.py
```diff
@@ -237,12 +237,12 @@ def get_backups(self, assignment, num_backups=10):
...
 
     def submissions(self, group, assignment):
-        """A query that fetches all backups for a group and assignment."""
+        """A query that fetches all submissions for a group and assignment."""
         members = self.members(group)
         return Submission.query(
             Submission.submitter.IN(members),
             Submission.assignment == assignment
-        ).order(-Backup.server_time)
+        ).order(-Submission.server_time)
```
Sumukh: That looked wrong ...  `-Backup.server_time` might be a relic of the days when submissions were backups.

knrafto: My bad, copy-paste error. Thanks!

server/app/models.py
```diff
@@ -415,13 +415,23 @@ def scores_for_assignment(self, assignment):
...
-            Format: [['STUDENT', 'SCORE', 'MESSAGE', 'GRADER', 'TAG', 'SUBM_ID', 'REVISION_ID']]
+            Format: [['STUDENT', 'SCORE', 'MESSAGE', 'GRADER', 'TAG', 'SUBM_ID', 'REVISION_ID', 'IS_FINAL']]
         """
         fs = self.get_final_submission(assignment.key)
         scores = []
         if fs:
             scores = fs.get_scores()
-        return (scores, True) if scores else ([[self.email[0], 0, None, None, None, None, None]], False)
+        existing_subms = [s[5] for s in scores]
+        submissions = self.get_submissions(assignment.key, num_submissions=50)  # Fetch all submissions as well
```
Sumukh: This could be a potential culprit of why this query is so resource intensive.

Sumukh:
The current way of looping over past submissions and exporting scores (L430 of models.py) seems to be the culprit of memory errors. I'll try different instance sizes to get an idea of how much memory it actually needs.

Sumukh:
The export seemed to have worked with a F4_1G instance on staging. I don't want to merge/deploy this since it would probably prevent most other exports from working.

knrafto:
> It would probably prevent most other exports from working

What do you mean?

Sumukh:
From my (limited) testing it seems like we'd have to deploy F4_1G instances to get exports that search non-final submission to finish (without running out of memory). Exports on smaller instance sizes ran out of memory. 

If that's the tradeoff, I'd rather not use that instance size for the primary version. 

If we can't figure out a way to do this export efficiently, I'd be fine if we didn't have this feature (getting scores from non-final submissions). I just needed this to pull some scores for scheme.

knrafto:
How big is `subm.export_scores()` typically?

Sumukh:
Closing - this was useful as a script but I don't think we should deploy this.

PR #628, opened 2015-11-26
Stream scores to GCS

knrafto:


knrafto:
Tested in staging with 256MB(?) instances, successfully exported scores for Scheme. I'm not sure what the output should look like, but it looks okay to me.

@Sumukh ready for review

Sumukh:
Thanks - this LGTM. I'm going to merge this with #627 and test in staging.

Sumukh:
Your changes look good but it doesn't seem to alleviate the problem with #627 running out of memory (meaning the issue is somewhere in there)

PR #634, opened 2016-01-12
Testing login endpoint

knrafto:
When the TESTING_LOGIN config value is True, `/login` will redirect to `/testing-login`, which just asks for your email and lets you in the back door. Tests can log in users by `POST /testing-login/authorized?name=Marty%20McFly&email=martymcfly@aol.com`.

Both `/testing-login` and `/testing` will 404 otherwise.


server/authenticators.py
```diff
@@ -1,114 +0,0 @@
...
-        return self.google.authorized_response()
-
-    def get(self, *args, **kwargs):
-        return self.google.get(*args, **kwargs)
-
-    def email(self, access_token):
-        return self.token_email(access_token)
-
-    @staticmethod
-    def token_email(access_token):
```
Sumukh: The API relies on having a token_email function to call on all requests. We can either put this token_email function in there (along with a trapdoor for testing-login) or add it directly to the `auth.py` controller and import it in api. 

Any preference for where this goes? (I'm leaning towards the latter option)

server/controllers/auth.py
```diff
@@ -0,0 +1,96 @@
...
+# Requires that TESTING_LOGIN = True in the config; production should NEVER have this on.
+@auth.route('/testing-login')
+def testing_login():
+    if not current_app.config.get('TESTING_LOGIN', False):
+        abort(404)
+    return render_template('testing-login.html', callback=url_for(".testing_authorized"))
+
+@auth.route('/testing-login/authorized', methods=['POST'])
+def testing_authorized():
+    if not current_app.config.get('TESTING_LOGIN', False):
```
Sumukh: Can we also have another check here - `or current_app.config.get('ENV') == 'PROD'`? 

The motivation is that only the prod config will have the credentials for the prod database - so even if we accidently deploy with the `TESTING_LOGIN` - it'll never leak on production. 

Another approach is to only allow fake logins on DEV or TEST environments and remove the TESTING_LOGIN config. The downside would be that we wouldn't be able to test google login locally - which I don't think is acceptable for now.

Sumukh:
Overall looks good. Let's resolve the two comments above and we should be set to merge. 

Here's the relevant part of api.py: https://github.com/Cal-CS-61A-Staff/ok/blob/rewrite/Sumukh/v3/server/controllers/api.py#L91

knrafto:
Simplified a little:

* Moved login_manager initialization to auth.py
* Use Flask-Login's request_loader to authenticate API requests (tested with enrollment API, works)
* Users' names aren't in the model and I couldn't get Google to return them, so I took them out
* I thought keeping the access token in the session was a little messy because it's not actually used anywhere besides a debug endpoint. So I added `/access-token`, where you can sign in again to get a Google access token. Does this need a new home?

server/controllers/api.py
```diff
@@ -151,8 +139,7 @@ def get(self, user):
         return {
             'version': API_VERSION,
             'url': '/api/{}/'.format(API_VERSION),
-            'documentation': 'http://github.com/Cal-CS-61A-Staff/ok/wiki',
-            'access_token': session['google_token']  # FOR TESTING ONLY
```
knrafto: Should I restore this?

Sumukh:
Looks good. I haven't had a chance to test it locally - but if you want to merge it - go ahead!

server/controllers/api.py
Sumukh: No need. It was marginally useful while working on the token auth but I don't think it's needed here. 
In an ideal world our users should never had to deal with their access_token

PR #635, opened 2016-01-13
Testing

knrafto:
Ported to use Flask-Testing. Tests now use a separate Postgres database (oktest). Examples of unit tests (test_user.py) and integration tests (test_main.py, test_auth.py)


Sumukh:
Looks much better! Thanks. 

Let's make sure travis has the right test running command. (see .travis.yml and Makefile) 
Also travis is failing because it can't import `secret_keys` 

```
Traceback (most recent call last):
  File "./manage.py", line 9, in <module>
    from server import create_app
  File "/home/travis/build/Cal-CS-61A-Staff/ok/server/__init__.py", line 10, in <module>
    from server.controllers.auth import auth, login_manager
  File "/home/travis/build/Cal-CS-61A-Staff/ok/server/controllers/auth.py", line 12, in <module>
    from server.secret_keys import google_creds
ImportError: No module named secret_keys
```

There are a variety of ways to fix this. The hacky way is a try catch but I'd like to avoid that. We could have a manage.py command that creates the secret key file (loads the creds from env variables)

Sumukh:
Travis Passes (Though we fail the coveralls check). Merging.

PR #636, opened 2016-01-17
Groups

knrafto:
Implements student group actions. Still need to log group events and make admin tools.

### DIFF
--- a/server/models.py
+++ b/server/models.py
@@ -27,10 +27,6 @@ class User(db.Model, UserMixin, TimestampMixin):
     alt_email = db.Column(db.String())
     active = db.Column(db.Boolean(), default=True)
 
-    def __init__(self, email, sid=None):
-        self.email = email
-        self.sid = sid
-
     def __repr__(self):
         return '<User %r>' % self.email
 
@@ -42,17 +38,9 @@ class User(db.Model, UserMixin, TimestampMixin):
         ).all()
 
     @staticmethod
-    def from_email(email):
-        """
-        Get a User with the given email address, or create one if no User with
-        this email is found.
-        """
-        user = User.query.filter_by(email=email).one_or_none()
-        if not user:
-            user = User(email)
-            db.session.add(user)
-            db.session.commit()
-        return user
+    def lookup(email):
+        """Get a User with the given email address, or None."""
+        return User.query.filter_by(email=email).one_or_none()
 
 
 class Course(db.Model, TimestampMixin):
--- a/server/models.py
+++ b/server/models.py
@@ -1,7 +1,9 @@
 from flask.ext.sqlalchemy import SQLAlchemy
-from sqlalchemy import UniqueConstraint
+from sqlalchemy import PrimaryKeyConstraint
 from sqlalchemy.dialects import postgresql as pg
 from sqlalchemy.ext.hybrid import hybrid_property
+from sqlalchemy.orm import backref
+from werkzeug.exceptions import BadRequest
 
 from flask.ext.login import UserMixin, AnonymousUserMixin
 from flask.ext.cache import Cache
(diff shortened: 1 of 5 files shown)


server/models.py
```diff
@@ -254,43 +247,136 @@ class GroupMember(db.Model, TimestampMixin):
...
+        if not self.assignment.active:
+            raise BadRequest('The assignment is past due')
+        member = GroupMember.query.filter_by(
+            user=user,
+            group=self,
+            status='pending'
+        ).one_or_none()
+        if not member:
+            raise BadRequest('{0} is not invited to this group'.format(user.email))
+        member.status = 'active'
```
Sumukh: Will this persist to the database if we don't have `db.session.commit()`?

knrafto: No. It says (in the sqlalchemy docs)[http://docs.sqlalchemy.org/en/latest/orm/session_basics.html#when-do-i-construct-a-session-when-do-i-commit-it-and-when-do-i-close-it] to do the commit somewhere else, so maybe it can go in the controllers. I'm fine with committing in the "public" methods (accept, decline) etc. though

server/models.py
```diff
@@ -254,43 +247,136 @@ class GroupMember(db.Model, TimestampMixin):
...
+
+    assignment = db.relationship("Assignment")
+
+    def size(self):
+        return GroupMember.query.filter_by(group=self).count()
+
+    def has_status(self, user, status):
+        return GroupMember.query.filter_by(
+            user=user,
+            group=self,
```
Sumukh: I'm not sure what the ORM does here since group is a relationship. I can find out in a bit. (mostly just curious about what the ORM is doing) 

The alternative is to do `group_id=self.id`

server/models.py
Sumukh: Putting them in the public methods sounds like a good idea.

server/models.py
```diff
@@ -98,16 +94,13 @@ def active(self):
...
                           nullable=False)
-    role = db.Column(db.Enum(*VALID_ROLES, name='role'), nullable=False)
-    course = db.relationship("Course", backref="participants")
+    role = db.Column(db.Enum(*VALID_ROLES, name='role'), default=STUDENT_ROLE, nullable=False)
 
-    def __init__(self, user, course_id, role=STUDENT_ROLE):
-        self.user = user
-        self.course_id = course_id
-        self.role = role
+    user = db.relationship("User", backref="participants")
```
Sumukh: Sidenote for merge: I called the backref `participations`, enrollments would be the ideal name - but we already have a User model method named enrollments. I'll sort it out during the merge

server/models.py
```diff
@@ -254,43 +247,136 @@ class GroupMember(db.Model, TimestampMixin):
...
     """A group is a collection of users who are either members or invited.
 
-    Members of a group can view each other's submissions.
+    Groups are created when a member not in a group invites another member.
+    Invited members may accept or decline invitations. Active members may
+    revoke invitations and remove members (including themselves).
 
-    Specification:
-    https://github.com/Cal-CS-61A-Staff/ok/wiki/Group-&-Submission-Consistency
+    A group must have at least 1 active member and at lease 2 participants.
```
Sumukh: s/lease/least

server/models.py
```diff
@@ -254,43 +247,136 @@ class GroupMember(db.Model, TimestampMixin):
...
+        ).one_or_none()
+        if member:
+            return member.group
+
+    @staticmethod
+    def invite(sender, recipient, assignment):
+        """Invite a user to a group, creating a group if necessary."""
+        if not assignment.active:
+            raise BadRequest('The assignment is past due')
+        group = Group.lookup(sender, assignment)
```
Sumukh: There should be a check to ensure that the user is enrolled in the same course.

Sumukh:
- [ ] Check that users are enrolled in course 
- [ ] db.session.commit() in the public methods (like invite) 

Once we have those two - we should be good to merge.

server/models.py
knrafto: There is, in `_add_member`

Sumukh: Ah I see it. thanks.  :+1:

PR #637, opened 2016-01-18
Enrollment (Admin) 

Sumukh:


server/controllers/admin.py
```diff
@@ -123,3 +129,55 @@ def assignment(cid, aid):
...
+        student = find_student.first()
+        if student:
+            students = Participant.query.filter_by(course_id=cid, role=STUDENT_ROLE,
+                user_id=student.id).paginate(page=page, per_page=1)
+        else:
+            flash("No student found with email {}".format(query), "warning")
+    if not students:
+        students = Participant.query.filter_by(course_id=cid,
+                role=STUDENT_ROLE).paginate(page=page, per_page=5)
+    staff = Participant.query.filter(Participant.course_id == cid,
```
Sumukh: This is a rather large controller method. A lot of this querying probably should become model methods.

server/controllers/api.py
```diff
@@ -338,18 +367,21 @@ def post(self, user):
         return {'id': score.id, 'backup': 1}
 
 
-class Enrollment(Resource):
+class Enrollment(PublicResource):
```
Sumukh: I'm not a huge fan of exposing this info externally. I plan on making this an authed method when I make some more client changes.

server/templates/staff/course/enrollment.html
```diff
@@ -0,0 +1,203 @@
...
+                          <li><a href="{{ url_for('.enrollment', cid=current_course.id, page=page) }}">{{ page }}</a></li>
+                        {% else %}
+                          <li class="active"><a href="#">Page {{ enrollments.page }} of {{ enrollments.pages }}</a></li>
+                        {% endif %}
+                      {% else %}
+                        <li><span class=ellipsis>…</span></li>
+                      {% endif %}
+                    {%- endfor %}
+                    {% if enrollments.has_next %}
+                      <li><a href="{{ url_for('.enrollment', cid=current_course.id, page=enrollments.prev_num) }}">»</a></li>
```
knrafto: Should be `next_num`

server/controllers/admin.py
```diff
@@ -123,3 +129,55 @@ def assignment(cid, aid):
     return render_template('staff/course/assignment.html', assignment=assgn,
                            form=form, courses=courses,
                            current_course=current_course)
+
+@admin.route("/course/<int:cid>/enrollment",
+             methods=['GET', 'POST'], defaults={'page': 1})
+@admin.route("/course/<int:cid>/enrollment/page/<int:page>",)
```
knrafto: `page` should be a query param since it's the same template - `/courses/1/enrollment/?page=5`

server/controllers/admin.py
```diff
@@ -123,3 +129,55 @@ def assignment(cid, aid):
...
+        Participant.enroll_from_form(cid, single_form)
+        flash("Added {email} as {role}".format(email=email, role=role), "success")
+
+    query = request.args.get('query', '').strip()
+    students = None
+    if query:
+        find_student = User.query.filter_by(email=query)
+        student = find_student.first()
+        if student:
+            students = Participant.query.filter_by(course_id=cid, role=STUDENT_ROLE,
```
knrafto: You could join with users and save a db query

server/forms.py
```diff
@@ -1,45 +1,57 @@
...
-class AssignmentForm(Form):
-    display_name = TextField(u'Display Name',
-                             [validators.required()])
-    name = TextField(u'Offering', filters=[strip_whitespace],
-                     validators=[validators.required()])
+
+class AssignmentForm(BaseForm):
+    display_name = StringField(u'Display Name',
+                               validators=[validators.required()]),
+    name = StringField(u'Offering', validators=[validators.required()])
```
knrafto: Should this be url-friendly?

server/forms.py
```diff
@@ -1,45 +1,57 @@
...
     due_date = DateTimeField(u'Due Date (Pacific Time)',
-                             filters=[convert_to_utc],
```
knrafto: `lock_date` is UTC but `due_date` isn't?
Making `lock_date` default to `due_date` + 1 day would save some confusion if someone else needs to create an assignment

server/forms.py
```diff
@@ -1,45 +1,57 @@
...
                              default=dt.datetime.now,
                              validators=[validators.required()])
     lock_date = DateTimeField(u'Lock Date (UTC)',
-                              filters=[convert_to_utc],
                               default=dt.datetime.now,
                               validators=[validators.required()])
     max_group_size = IntegerField(u'Max Group Size', default=1,
                                   validators=[validators.required()])
-    url = TextField(u'URL', [validators.optional()])
+    url = StringField(u'URL',
```
knrafto: What is this field for?

server/models.py
```diff
@@ -113,21 +119,70 @@ class Participant(db.Model, TimestampMixin):
...
 
-    @hybrid_property
-    def is_course_staff(self):
-        return self.role in STAFF_ROLES
-
     def is_staff(self, course):
         return self.course == course and self.role in STAFF_ROLES
 
+    @staticmethod
+    def enroll_from_form(cid, form):
```
knrafto: Use `@transaction` to make sure `db.session.rollback()` is called if an exception is thrown

server/forms.py
Sumukh: Fix the string to reflect that it's also in pacific time. (commit is coming in) 

That would be nice - but I'm going to file it as a `nice to have`

server/forms.py
Sumukh: It would be nice. Is there a way for us to validate that using the form class? No special characters?

server/forms.py
Sumukh: Linking to the assignment from the ok dashboard.

knrafto: Can't it be rendered from `url_for`?

Sumukh: I wasn't clear earlier - this is the course assignment page. `http://cs61a.org/proj/hog/` for example.

server/forms.py
knrafto: Regex `[0-9A-Za-z._-]+` or similar seems pretty reasonable

server/forms.py
knrafto: Ah, got it

Sumukh:
Changes made in this commit: https://github.com/Cal-CS-61A-Staff/ok/commit/325d55184665cce1d94e3b11b5f64c2ef63c4317 (On another branch unfortunately)

Once you are done looking at this - let's merge this branch and then the student frontend pr will display these edits.

PR #638, opened 2016-01-18
v3 Student Dashboards [WIP]

Sumukh:
Depends on #637 (Merge that PR first) 

Pages left: 
- [ ] Assignment Detail (submission history, group status) 
- [ ] Code view 

Pages Completed: 
- [x] Landing
- [x] Courses 
- [x] Course Overview (List Assignments) 

Other todo: 
- [ ] Assignments list doesn't fetch any info about the users submissions/group yet. It should. 
- [ ] Having the URLs be based off the course offering would be nice (and a small tweak for a later PR)


server/controllers/student.py
```diff
@@ -1,14 +1,65 @@
...
+    courses = {
+        'instructor': [e.course for e in enrollments if e.role in STAFF_ROLES],
+        'current': [e.course for e in student_enrollments if e.course.active],
+        'past': [e.course for e in student_enrollments if not e.course.active],
+        'num_enrolled': len(enrollments)
+    }
+    # Make the choice for users in one course
+    if len(enrollments) == 1 and auto_redir:
+        if courses['instructor']:
+            return redirect(url_for("admin.course", cid=enrollments[0].id))
```
knrafto: This redirect is better in `/` (not `/dashboard/`) so we can still get to the student dashboard easily

server/models.py
```diff
@@ -47,18 +57,20 @@ def __repr__(self):
...
-    def enrollments(self, roles=['student']):
-        return Participant.query.filter(
-            Participant.user == self.id,
-            Participant.role.in_(roles)
-        ).all()
+    def enrollments(self, roles=[STUDENT_ROLE]):
+        return [e for e in self.participations if e.role in roles]
+
+    def is_enrolled(self, course_id, roles=VALID_ROLES):
+        for enroll in self.participations:
```
knrafto: Should be a db filter, not a Python filter - `self.participations` will get really large

Sumukh: FWIW - I kept it this way because I noticed this resulted one less query to render the page.

server/controllers/student.py
Sumukh: I kind of agree - but it's fairly easy to access the student dashboard from the instructor panel. When instructor with in enrolled in only one course they wouldn't see anything but a link to the admin panel.

Sumukh: I agree. I'm going to mark this as a nice to have (#639)

server/controllers/student.py
```diff
@@ -1,14 +1,118 @@
...
+    if assgn:
+        course = assgn.course
+        group = Group.lookup(current_user, assgn)
+        if group:
+            # usr_ids = [u.id for u in group.users()]
+            # TODO : Fetch backups from group.
+            pass
+        backups = assgn.backups(current_user.id).limit(5).all()
+        subms = assgn.submissions(current_user.id).limit(5).all()
+        flagged = any([s.flagged for s in subms])
```
knrafto: This will be wrong if the flag is more than 5 submissions back

Sumukh: Right. Thoughts on how we should fix it: 
If we do an orderby on both flagged DESC & time DESC that should resolve it (while still keep it in one query). 

The simple solution is to do another query for backups with flagged=True

knrafto: Just pushed a final submission query in https://github.com/Cal-CS-61A-Staff/ok/pull/641 - we can just check if the final submission is flagged.

Sumukh: Just saw it. That looks good 
Relevant line: https://github.com/Cal-CS-61A-Staff/ok/commit/21c1764f82fc7063cce7f0929dc86d8f19c7b95d#diff-e07bdcf29922d0f33a2c70e8145df054R188

Sumukh:
This has been merged already - the new branch to reference is `v3/master`

PR #640, opened 2016-01-20
Submission code view

brianhou:
not ready for merge yet

https://db.tt/EPC1dUIA

- [ ] inline diff
- [ ] comments
- [ ] language inference
- [ ] Admin side view for code/comments

### DIFF
--- a/server/controllers/student.py
+++ b/server/controllers/student.py
@@ -100,25 +100,32 @@ def code(cid, aid, bid):
         backup = Backup.query.get(bid)
         if backup and backup.can_view(current_user, group, assgn):
             submitter = User.query.get(backup.submitter)
+            file_contents = [m for m in backup.messages
+                             if m.kind == "file_contents"]
+            if not file_contents:
+                flash("That code submission doesn't contain any code")
+                abort(404)
+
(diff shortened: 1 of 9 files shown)


server/controllers/student.py
```diff
@@ -100,15 +100,29 @@ def code(cid, aid, bid):
...
-                                m.kind == "file_contents"]
-            if file_contents:
-                files = file_contents[0].contents
-                return render_template('student/assignment/code.html', course=course,
-                        assignment=assgn, backup=backup, submitter=submitter,
-                        files=files)
-            else:
-                flash("That code submission doesn't contain any code")
+
+            diff = Diff.query.filter_by(assignment=aid, backup=bid).one_or_none()
```
Sumukh: We should only generate Diff objects for assignments that have templates. This might mean that we have a separate template just for diffs.

Sumukh:
Thanks for getting this started. 

We should only generate Diff objects for assignments that have templates. This might mean that we have a separate template just for diffs. We could insert a "What do the red lines/green lines" somewhere into that page.

knrafto:
Resolved by #669

PR #641, opened 2016-01-20
Group backups and submissions

### LINKED ISSUE #644: Group Lookup in student.py
Right now you can view submissions from groups you are invited to. 

That shouldn't be the case. Solution: Lookup with status = "accepted" or similiar.

knrafto:
Unifies backups and submissions and looks up submissions for groups. Fixes #644, #645


Sumukh:
We had some discussion about optimizing [this](https://github.com/Cal-CS-61A-Staff/ok/pull/641/files#diff-ead499da14d24ae498e025a69497a1a1R59) but we've tabled it. 

The other changes look fine.

PR #655, opened 2016-01-24
Database agnostic

### LINKED ISSUE #650: Migrate to MySQL (for now)
All this means is that we have to change json fields to text fields. 

The reasoning here: Google Cloud has a hosted MySQL service and it's easy enough to move to postgres if needed. If we get a hosted Postgres server - we can go back to that.

Sumukh:
Resolves #650 - or at least the motivation behind it. It wouldn't be a bad idea to test with an actual MySQL server. 

This allows use to use SQLite for testing - which is nice.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: python
 env:
 - SERVER_ENV=test
-addons:
-  postgresql: '9.4'
+# addons:
+#   postgresql: '9.4'
 python:
 - '3.3'
 - '3.4'
@@ -13,11 +13,9 @@ install:
 - pip install python-coveralls
 script:
 - make test
-before_script:
-- psql -c 'create database oktest;' -U postgres
+# before_script:
+# - psql -c 'create database oktest;' -U postgres
 after_success:
-- coveralls
-after_failure:
 - coveralls
 sudo: false
 cache:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,6 @@
 language: python
 env:
 - SERVER_ENV=test
-# addons:
-#   postgresql: '9.4'
 python:
 - '3.3'
 - '3.4'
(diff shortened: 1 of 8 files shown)


server/models.py
```diff
@@ -11,20 +10,21 @@
...
-    "ix": 'ix_%(column_0_label)s',
-    "uq": "uq_%(table_name)s_%(column_0_name)s",
-    "ck": "ck_%(table_name)s_%(constraint_name)s",
-    "fk": "fk_%(table_name)s_%(column_0_name)s_%(referred_table_name)s",
-    "pk": "pk_%(table_name)s"
-}
-
-metadata = MetaData(naming_convention=convention)
-db = SQLAlchemy(metadata=metadata)
+# convention = {
```
knrafto: Delete instead of commenting out?

server/settings/test.py
```diff
@@ -4,7 +4,7 @@ class TestConfig(LocalConfig):
     ENV = 'test'
     DEBUG_TB_INTERCEPT_REDIRECTS = False
 
-    SQLALCHEMY_DATABASE_URI = 'postgresql://postgres:@localhost:5432/oktest'
+    SQLALCHEMY_DATABASE_URI = 'sqlite:///../database.db'
```
knrafto: This should be different from the dev database. The URI "sqlite://" will give you a temporary in-memory database: https://pythonhosted.org/Flask-Testing/#testing-with-sqlalchemy

SETUP.md
```diff
@@ -27,3 +20,14 @@ Running App:
 $ ./manage.py server
 
 Open http://localhost:5000 in your browser of choice.
+
+Production DB: MySQL.
+
+Deprecated:
```
knrafto: Delete?

.travis.yml
```diff
@@ -13,12 +13,10 @@ install:
 - pip install python-coveralls
 script:
 - make test
-before_script:
-- psql -c 'create database oktest;' -U postgres
+# before_script:
+# - psql -c 'create database oktest;' -U postgres
```
knrafto: Delete?

knrafto:
LGTM

PR #663, opened 2016-01-30
Instantclick.io

Sumukh:
Uses instantclick.io

### DIFF
--- a/server/templates/staff/base.html
+++ b/server/templates/staff/base.html
@@ -76,8 +76,10 @@
 
       {% assets "adminlte_js" %}
           <script type="text/javascript" src="{{ ASSET_URL }}"></script>
-          <script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/instantclick/3.0.1/instantclick.min.js"></script>
-          <script data-no-instant>InstantClick.init(50);</script>
+          {% if not config['DEBUG'] %}
+            <script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/instantclick/3.0.1/instantclick.min.js"></script>
+            <script data-no-instant>InstantClick.init(50);</script>
+          {% endif %}
       {% endassets %}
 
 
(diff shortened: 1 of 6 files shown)


server/templates/staff/base.html
```diff
@@ -76,6 +76,8 @@
 
       {% assets "adminlte_js" %}
           <script type="text/javascript" src="{{ ASSET_URL }}"></script>
+          <script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/instantclick/3.0.1/instantclick.min.js"></script>
```
Sumukh: Will be bundled into our assets js file later.

knrafto:
Is there some JS library that's rendering the times? It makes the submissions table jump around when reloading (e.g. clicking a flag), and does weird things like this:
screen shot 2016-01-29 at 8 08 20 pm

Sumukh:
Yes - that's moment.js. It's a hack for now until I implement a jinja filter to convert times to PST

Sumukh:
This is now only active when debug mode is off. 
Filed an issue for the timestamps  (#667).

knrafto:
I'm worried that this won't be well tested and it's easy to mess up - for example, if we miss blacklisting a logout link, hovering over the link will mysteriously log the user out. I think we should hold off on this for now and come back to it when the app is more mature

alvinwan:
You could use Python's arrow for easier time conversions. Here is a sample you could probably just copy and paste with minor changes.
```
import arrow
utc = arrow.get(py_datetime)
utc.to('local')  # converts to US/Pacific time
utc.humanize() # for "3 hours ago"-like string
utc.format('YYYY-MM-DD HH:mm:ss')  # 2016-01-22 09:11:12
```
Arrow was, in fact, inspired by moment.js. http://crsmithdev.com/arrow/

knrafto:
Logging out should use POST not GET

Sumukh:
This fix is included in https://github.com/Cal-CS-61A-Staff/ok/tree/v3/sumukh/assorted

PR #665, opened 2016-01-30
Log group actions - resolves #649

### LINKED ISSUE #649: Group Logging
Audit Log: 
- On every action (create, invite, accept, decline, remove, delete) 
 - action type (str ENUM) 
 - sender 
 - target 
 - group_before (serialized the existing group - could be null) 
 - group_after (serialized - could be null)

knrafto:
Doesn't log creation and deletion, for ease of implementation. Instead, actions that start/end with a delete group will show up as a group with no members.

### DIFF
--- a/server/models.py
+++ b/server/models.py
@@ -9,6 +9,7 @@ from flask.ext.cache import Cache
 cache = Cache()
 
 import functools
+import contextlib
 import csv
 import json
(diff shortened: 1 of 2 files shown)


Sumukh:
LGTM. I like the use of the context manager.

PR #666, opened 2016-01-30
Hash IDs in URLs - resolves #659

### LINKED ISSUE #659: No ID numbers in URLs
Use something like [hashids](http://hashids.org/python/) to obfuscate the ID numbers

knrafto:
Not sure where to put the converter class. utils.py maybe?

### DIFF
--- a/server/__init__.py
+++ b/server/__init__.py
@@ -4,7 +4,7 @@ from flask import Flask
 from flask.ext.rq import RQ
 from hashids import Hashids
 from webassets.loaders import PythonLoader as PythonAssetsLoader
-from werkzeug.routing import BaseConverter
+from werkzeug.routing import BaseConverter, ValidationError
 
 from server import assets
 from server.models import db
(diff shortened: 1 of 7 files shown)


server/__init__.py
```diff
@@ -20,6 +22,17 @@
     debug_toolbar
 )
 
+class HashidConverter(BaseConverter):
+    # ID hashing configuration.
+    # DO NOT CHANGE ONCE THE APP IS PUBLICLY AVAILABLE. You will break every
```
Sumukh: Should we add a test to make sure that 
 - the hashid converter exists
 - that the config hasn't changed (by encoding/decoding a value) and comparing it's value.

Sumukh:
I agree that utils.py is a better place to put the class. The templates will have to use this to convert some IDs. (or we could make it a model method)

server/utils.py
```diff
@@ -1,15 +1,19 @@
...
-    return models.Assignment.query.filter_by(name=name).one_or_none()
+    def to_python(self, value):
+        numbers = self.hashids.decode(value)
+        if len(numbers) != 1:
+            raise ValidationError('Could not decode hash {} into ID'.format(value))
+        return numbers[0]
 
-def course_by_name(name):
-    return models.Course.query.filter_by(offering=name).one_or_none()
+    def to_url(self, value):
```
knrafto: This doesn't seem to be used anywhere

knrafto:
LGTM, merging

PR #668, opened 2016-01-31
Group Actions (Invite/Remove/Reject/Accept)

Sumukh:
Frontend forms & Routes for actions. 

This PR also handles CSRF protection for all forms. 
Assignment Dashboard: 
image

Confirm Removals:
image

Invite Actions:
image


server/__init__.py
```diff
@@ -22,6 +24,7 @@
 )
 
 
+
```
Sumukh: Random line? To Remove

server/controllers/student.py
```diff
@@ -89,13 +92,29 @@ def assignment(course, assign):
...
-    flagged = final_submission and final_submission.flagged
-    return render_template('student/assignment/index.html', course=course,
-            assignment=assign, backups=backups, subms=subms, flagged=flagged)
+    fs = assign.final_submission(user_ids)
+    group = Group.lookup(current_user, assign)
+    can_invite = assign.max_group_size > 1
+    can_remove = False
+    if group:
+        can_invite = len(group.members) < assign.max_group_size
+        can_remove = current_user in [m.user for m in group.members if m.status == 'active']
```
Sumukh: All that's needed here is to figure out if the current user is actually a member. `user_ids` doesn't do this since it always has the current_users id.

server/controllers/student.py
```diff
@@ -165,6 +184,76 @@ def flag(course, assign, bid):
             assign.flag(bid, user_ids)
         else:
             assign.unflag(bid, user_ids)
-        return redirect(next_url)
+        if is_safe_redirect_url(request, next_url):
```
Sumukh: Should prevent [open redirects](https://www.owasp.org/index.php/Open_redirect)

server/static/css/main.css
```diff
@@ -17,3 +17,12 @@ pre code  {
 .no-dash {
   border-bottom: none !important;
 }
+
+.top-align {
+    margin-top: -1em;
+}
+
+td.slim-cell {
```
Sumukh: I don't think this is used anymore in this PR - but I can imagine it being useful later.

server/controllers/student.py
knrafto: `can_remove = group.has_status(current_user, 'active')`

server/static/js/student.js
```diff
@@ -0,0 +1,19 @@
+jQuery(document).ready(function($){
+    $('.remove-confirm').on('click',function(e){
+        e.preventDefault();
+        var form = $(this).parents('form');
+        var info = form[0][1].value;
+        swal({
+            title: "Just double checking",
+            text: "Are you sure you want to remove " + info + " from the group?",
```
knrafto: If you try to leave, it shows "Are you sure you want to remove [your email] from the group?", which could be confusing.

knrafto:
The accept/reject buttons are aligned vertically for me:
screen shot 2016-01-31 at 12 28 02 pm

knrafto:
For the status column, "in group" and "invited" might be clearer than "active" or "pending" status

knrafto:
The accept/reject/remove buttons (that are just icons) should have tooltips explaining what they will do

server/templates/student/assignment/_tablehelper.html
```diff
@@ -1,4 +1,6 @@
-{% macro render_subms(subms, course, flagged, tname="Submissions", empty=True) %}
+{% import 'student/assignment/_formhelpers.html' as forms %}
+
+{% macro render_subms(subms, course, flagged, flag_form, tname="Submissions", empty=True) %}
```
knrafto: `flag_form` is missing when called from the "view all submissions" page

Sumukh:
Addressed your comments. 

The copy on the tooltips aren't great - but they should work. Feel free to change them if you have ideas for them. 

I deleted the mocks from the templates

Issue #646, opened 2016-01-23
Do not store Diffs


Sumukh:
If we have comments be associated to Backups directly, we don't have to store Diff objects. We should be able to compute these on the fly. 

That does mean comments are not allowed on the removal parts of a diff - but that's ok.

Issue #647, opened 2016-01-23
Diffs should actually be chunked


Sumukh:
Previous: 
```
-
+
- 
+ 
- 
+
```
Ideal:
```
-
-
- 
+
+
+
```

@brianhou - I think this was an issue with difflib

PR #669, opened 2016-01-31
Render code and diffs

knrafto:
Code and diffs look pretty. You can switch between the two with a button. The seeded before/after files are difflib.py from Python 2.7 and Python 3.5 respectively.


knrafto:
Resolves #646, #647

knrafto:
Ready for review!

server/assets.py
```diff
@@ -47,6 +47,7 @@
 
 student_css = Bundle(
     'css/student.css',
+    'css/highlight.css',
```
Sumukh: It seems like including highlight.css with the student bundle causes some alignment issues.  Do you also see this? 

Adding highlight to the common_css bundle seemed to fix this. I think it's fair to include highlight.css in common since it's shared between the staff/student view.

image

Sumukh:
Looks great!  Some comments: 

- The syntax highlighting CSS italicizes comments and docstrings: My person preference is to not italicize them. I believe Github highlighting also does not use italics for comments/strings. 
- The highlight.css file can be moved to common (and this also solved a layout issue with the overall page). I couldn't really figure out why it caused the layout issue (screenshot in comment above)
- It might be worth removing the `h1` header here. I'm not sure what value it provides for the user 
image 
 -  We could replace it with a h3 with the submission info. 
- On small screen sizes - the tables goes off the edge. 
image

server/templates/student/assignment/code.html
```diff
@@ -35,29 +36,31 @@
...
         <!-- start desktop/tablet version -->
         <div class="subcontent list">
           <div class="wrap">
-            <p> Recieved at <time datetime="{{ backup.created }}">{{ backup.created }}</time> from {{ submitter.email }} </p>
-            {% if files %}
-              {% for file, code in files.items() %}
-              {{ file }}
-                <pre><code class="python">{{code}}</code></pre>
-              {% endfor %}
+            <p> Recieved at <time datetime="{{ backup.created }}">{{ backup.created }}</time> from {{ backup.submitter.email }}
```
Sumukh: Typo in the original template - should be Received

Sumukh:
f65b0a9 fixes the issues I mentioned above. Feel free to merge if those changes look ok to you.

server/assets.py
knrafto: What exactly is misaligned here?

Sumukh: I outlined the parts in yellow. There's some extra padding that applies to the sidebar and a bit of the navbar. 

alignment

PR #683, opened 2016-02-05
Shorter routes

### LINKED ISSUE #675: Shorter routes

knrafto:
Resolves #675 for students.

Moved converters to a new file, converters.py.
I noticed there's a lot of repetition in the templates. It made it annoying to replace all the `url_for`s.


tests/test_auth.py
```diff
@@ -20,13 +20,3 @@ def test_testing_login_fail(self):
         app.config['TESTING_LOGIN'] = False
         response = app.test_client().get('/testing-login/')
         self.assert_404(response)
-
-    def test_restricted(self):
```
knrafto: Is there a good way to salvage this test?

server/converters.py
```diff
@@ -0,0 +1,53 @@
...
+def restricted_name_part(exceptions):
+    """Return a regex that matches a URL part except one of the words in the
+    exceptions list.
+    """
+    # (?!...) is a negative lookahead
+    return ''.join('(?!{}/)'.format(w) for w in exceptions) + name_part
+
+class OfferingConverter(BaseConverter):
+    regex = restricted_name_part([
+        'login',
```
Sumukh: Some other restricted names could be: 
`api`
`admin`
`auth` (not sure about this one - I'll consult the route table and update this) 

I'll add more of I think of them. I'm not sure if this

tests/test_auth.py
Sumukh: I think so. It just needs to incorporate some elements of frontend testing - checking if certain elements exist on the page.

server/converters.py
```diff
@@ -0,0 +1,53 @@
...
+        'logout',
+        'testing-login',
+    ]) + '/' + name_part + '/' + name_part
+
+class AssignmentNameConverter(BaseConverter):
+    regex = OfferingConverter.regex + '/' + restricted_name_part([
+        'enrollment',
```
Sumukh: I'm a little confused about why need to protect /enrollment specifically here. 

Is this a holdover from the current v2 design `/enrollment` route? That route has been moved to the API.

server/controllers/student.py
```diff
@@ -214,17 +187,12 @@ def group_remove(course, assign):
             group.remove(current_user, target)
         except BadRequest as e:
             flash(e.description, 'danger')
+    return redirect(url_for('.assignment', offering=assignment.course.offering, name=assignment))
```
Sumukh: The group actions redirects: `name=assignment` should be `name=assignment.name` 

I don't offering is actually necessary - it would just be added as a query parameter (and isn't used by the route anyway)

server/controllers/student.py
```diff
@@ -14,53 +14,56 @@
...
+        courses = {
+            'instructor': [e.course for e in enrollments if e.role in STAFF_ROLES],
+            'current': [e.course for e in student_enrollments if e.course.active],
+            'past': [e.course for e in student_enrollments if not e.course.active],
+            'all': all_courses,
+            'num_enrolled': len(enrollments)
+        }
+        # Make the choice for users in one course
+        if len(enrollments) == 1 and auto_redir:
+            return redirect(url_for(".course", course=enrollments[0].course.offering))
```
Sumukh: This errors out (no offering).

tests/helpers.py
```diff
@@ -39,6 +39,14 @@ def setup_course(self):
             max_group_size=4)
         db.session.add(self.assignment)
 
+        self.staff = User(email='okstaff@okpy.org', is_admin=True)
```
Sumukh: Unused right now, but could be used later.

Sumukh:
I addressed some of my comments in the commits above. 

 - [ ] Add API/Admin to restricted routes
-  [ ] Make admin course creation enforce that assignment names are formatted correctly (start with course offering)

Sumukh:
LGTM - We'll enforce this on the admin side later.

PR #688, opened 2016-02-14
Add Version Model+API

### LINKED ISSUE #681: Client versions

Sumukh:
Resolves #681
- Versions now have a model + API
- Submission/Backups now use the version value sent by the client instead of a hardcoded value
- Ok Client changes were made to account for the renaming of current_download_link to download_link 

Design changes from v2: We don't store previous version info. We never really used that information. 

- Entirely unrelated changes (didn't want to make a separate PR for these)
  - Enrollment API is protected by some permissions now (this requires some minor changes to ok-client)
  - login links to testing login in dev mode
  - Quick Links on testing-login to login to staff/student accounts

### DIFF
--- a/server/controllers/api.py
+++ b/server/controllers/api.py
@@ -31,7 +31,6 @@ from server.utils import encode_id
 import server.models as models
 
 
-
 endpoints = Blueprint('api', __name__)
 endpoints.config = {}
 
--- a/server/controllers/api.py
+++ b/server/controllers/api.py
@@ -231,13 +231,21 @@ class MessageSchema(APISchema):
         'created': fields.DateTime(dt_format='rfc822')
     }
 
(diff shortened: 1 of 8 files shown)


server/controllers/api.py
```diff
@@ -133,13 +137,22 @@ def name_to_assign_id(name):
...
+    if version:
+        return version.current_version, version.download_link
+    return None, None
+
 class Resource(restful.Resource):
     version = 'v3'
     method_decorators = [authenticate, check_version]
     # applies to all inherited resources
 
+    def __repr__(self):
```
Sumukh: This is so that flask-cache can memoize some routes that will be hit often/won't change.

knrafto:
LGTM, feel free to merge once the tests are fixed

PR #695, opened 2016-02-17
Adds download button to student code view

### LINKED ISSUE #670: Download links for files
There should be some button in the code view that downloads the file.

jathak:
This resolves #670.


knrafto:
Nice, works perfectly! LGTM

server/controllers/student.py
```diff
@@ -132,6 +132,18 @@ def code(name, submit, bid):
     return render_template('student/assignment/code.html',
         course=assign.course, assignment=assign, backup=backup, use_diff=use_diff,
         files_before=assign.files, files_after=backup.files())
+        
+@student.route('/download/<hashid:bid>/<file>')
```
Sumukh: I think this means that course id's can't start with download. I'm fine with adding download to the protected URLs list. 

Alternatively, the route could be `/<assignment_name:name>/submissions/<hashid:bid>/download/<file>`

knrafto:
Actually, could you add a test?
It would just be `response = self.client.get(<url>)` and testing the body matches the file

server/controllers/student.py
knrafto: I like that better too. You can copy the `code()` endpoint:
```
@student.route('/<assignment_name:name>/<bool(backups, submissions):submit>/<hashid:bid>/download/<file>')
```

Sumukh:
A few more ideas for tests: 
- Special unicode characters (smart quotes, characters like ❀ and other unicode characters etc).
- Make sure unauthenticated users/users that are not in the group cannot download file.

knrafto:
:+1:

tests/test_download.py
```diff
@@ -0,0 +1,62 @@
...
+            
+        db.session.add(self.backup)
+        db.session.add(self.message)
+        db.session.commit()
+        
+    def test_simple(self):
+        filename = "test.py"
+        contents = "x = 4"
+        self._add_file(filename, contents)
+        encoded_id = utils.encode_id(self.backup.id)
```
Sumukh: For future cleanup: seems like we can generalize this code (34-40) to a `_download` helper since it's repeated.

PR #696, opened 2016-02-18
If no pygments lexer, don't highlight file

### LINKED ISSUE #689: Pygments should have a default lexer
If pygments can't find an appropriate lexer by the filename - we should choose one. It currently will throw an exception and the code view page won't load. 

use case:  submitted file name is `data` instead of `data.txt`

knrafto:
Fixes #689. Scroll down to the bottom a submission to see a plaintext file that's not highlighted.

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -79,7 +79,13 @@ def seed():
     db.session.add(assign2)
     db.session.commit()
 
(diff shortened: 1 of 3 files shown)


Sumukh:
LGTM

PR #697, opened 2016-02-18
Show semester example

Timmy-G:
Currently just assumed offering string will be of format "..../..../sp16" or something of the sort. I went ahead and made it so anywhere that previously showed a course's display name now shows its display name in addition to the semester of its offering. 

Let me know if I should revert it in any of these places or have a less hacky way for me to determine the semester.


server/models.py
```diff
@@ -137,6 +136,17 @@ def __repr__(self):
...
+    @property 
+    def display_name_with_semester(self):
+        year = self.offering[-2:len(self.offering)]
+        if "fa" in self.offering[-4:len(self.offering)]:
+            semester = "Fall"
+        elif "sp" in self.offering[-4:len(self.offering)]:
+            semester = "Spring"
+        else:
+            semester = "Summer"
+        return self.display_name + " (%s 20%s)" % (semester, year)
```
Sumukh: " ({0} 20{1}).format(semester, year)

server/models.py
```diff
@@ -137,6 +136,17 @@ def __repr__(self):
     def by_name(name):
         return Course.query.filter_by(offering=name).one_or_none()
 
+    @property 
+    def display_name_with_semester(self):
+        year = self.offering[-2:len(self.offering)]
```
Sumukh: I think it'll be safer to split on `/` characters since those are required in the offering tag.

Sumukh:
Nice!

Some tests are still failing like `test_backup` (see travis). This is mostly because the test expects something like `cs61a` but it is getting `cs61a (spring 2016)` instead.

Timmy-G:
Okay I'll fix those things up soon! How do I run the tests myself without having to make a pull request?

knrafto:
`make test` will run tests

tests/test_api.py
```diff
@@ -65,7 +65,7 @@ def _test_backup(self, submit):
             'course': {
                 'id': course.id,
                 'offering': course.offering,
-                'display_name': course.display_name,
+                'display_name': course.display_name_with_semester,
```
knrafto: The API should send the actual display name.

Timmy-G:
I've fixed the issues mentioned (other than the tests). For some reason when I run make test I get the error: `make: docker-compose: No such file or directory` before running any tests.

server/models.py
```diff
@@ -137,6 +136,17 @@ def __repr__(self):
...
+    @property 
+    def display_name_with_semester(self):
+        year = self.offering[-2:]
+        if "fa" in self.offering[-4:]:
+            semester = "Fall"
+        elif "sp" in self.offering[-4:]:
+            semester = "Spring"
+        else:
+            semester = "Summer"
+        return self.display_name + " ({0} 20{1})".format(semester, year);
```
Sumukh: Semicolon?

Sumukh:
LGTM. Thanks!

server/models.py
Timmy-G: LOL too much C and C++.... So many projects right now

PR #698, opened 2016-02-18
Disable Travis email notifications

knrafto:
Seems kind of silly if we have Slack notifications. They're just noise.


Sumukh:
I think this does mean that people not on our slack that submit a PR won't get emails - but I'm fine with that.

PR #700, opened 2016-02-18
Render full diffs (all context lines)

knrafto:
Along with a tabbed switcher. Needed for composition grading.


Sumukh:
LGTM. I added a 404 for diffs against assignments that don't have templates (and similarly, the template doesn't link to diffs if no templates exist)

PR #702, opened 2016-02-25
Scores & Comments Model

Sumukh:
Adding a scores & comment models to support migrating comments & scores from V2.

### DIFF
--- a/server/models.py
+++ b/server/models.py
@@ -60,6 +60,10 @@ class Json(types.TypeDecorator):
 def ok_blob(element, compiler, **kw):
     return "BLOB"
 
+@compiles(mysql.MEDIUMTEXT, 'sqlite')
+def ok_text(element, compiler, **kw):
+    return "TEXT"
+
 class JsonBlob(types.TypeDecorator):
     impl = mysql.MEDIUMBLOB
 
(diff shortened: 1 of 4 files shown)


server/models.py
```diff
@@ -630,3 +637,34 @@ class Version(Model):
...
+class Score(Model):
+    id = db.Column(db.Integer, primary_key=True)
+    grader_id = db.Column(db.ForeignKey("user.id"), nullable=False)
+    assignment_id = db.Column(db.ForeignKey("assignment.id"), nullable=False)
+    backup_id = db.Column(db.ForeignKey("backup.id"), nullable=False, index=True)
+
+    kind = db.Column(db.String(255), nullable=False)
+    score = db.Column(db.Float, nullable=False)
+    message = db.Column(mysql.MEDIUMTEXT)
+    public = db.Column(db.Boolean, default=True)
```
knrafto: Does this mean we'll need a giant UPDATE over every score in order to release scores?

Sumukh: For now I don't plan on having this property change at all - some scores will only be visible during a score export. (It's useful to hide scores if you are just testing auto-grading or you don't want there to be two places students can check grades)

We could build out the ability to do updates if it's needed to hide grades and publish.

knrafto:
LGTM, feel free to merge

PR #708, opened 2016-03-02
Editing assignments as admin

Timmy-G:


server/controllers/admin.py
```diff
@@ -124,6 +126,17 @@ def assignment(cid, aid):
     if assgn.course != current_course:
         return abort(401)
 
+    assgn.name = form.name._value()
```
Sumukh: What happens if it's not a POST (or a valid submission?) - We want to validate that they submitted the form (and that the forms are valid). 

1. Method is POST. 
2. form.validate_on_submit()  (before you create the form from `assgn`) 

Take a look at `new_assignment`.

server/controllers/admin.py
```diff
@@ -124,6 +126,17 @@ def assignment(cid, aid):
...
+    assgn.display_name = form.display_name._value()
+    assgn.due_date = datetime.strptime(form.due_date._value(), '%Y-%m-%d %H:%M:%S')       
+    assgn.lock_date = datetime.strptime(form.lock_date._value(), '%Y-%m-%d %H:%M:%S')
```
Sumukh: I don't think you have to do this if you use form.validate_on_submit() and form.populate_obj 

Also note to self: These times should be converted to UTC sometime (but that's not related to this PR)

server/controllers/admin.py
knrafto: You only have to use `form.validate_on_submit()`. It'll check the request method is POST for you. 
Docs: https://flask-wtf.readthedocs.org/en/latest/quickstart.html#validating-forms

Sumukh:
Great. I think we are going to add some cleanup regrading time conversion on forms to this PR (@Sumukh) 

We should also have tests for editing assignments.

server/controllers/admin.py
```diff
@@ -117,13 +119,17 @@ def assignment(cid, aid):
     # Convert TZ to Pacific
     assgn.due_date = convert_to_pacific(assgn.due_date)
     assgn.lock_date = convert_to_pacific(assgn.lock_date)
```
knrafto: WTForms should handle time conversion

Sumukh: Any ideas here? I tried thinking of a few solutions but none of them seemed right. The best one I have so far is a custom DateTimeField renderer (in _formhelpers) that calls `utils.local_time`

I'm planning on leaving this for later.

Sumukh:
Merged via #736

PR #709, opened 2016-03-03
Fixes admin enrollment forms

jathak:
Single and batch enrollment from the admin page should now work

### DIFF
--- a/server/models.py
+++ b/server/models.py
@@ -316,46 +316,65 @@ class Enrollment(Model):
             db.session.add(usr)
         db.session.commit()
         role = form.role.data
-        Enrollment.create(cid, [usr.id], role)
+        info = {
+            'id': usr.id,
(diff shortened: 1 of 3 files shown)


knrafto:
Thanks! Could you add some tests so it doesn't break again?

server/models.py
```diff
@@ -316,46 +316,65 @@ def enroll_from_form(cid, form):
...
+            'class_account': form.secondary.data,
+            'section': form.section.data
+        }
+        Enrollment.create(cid, [info], role)
 
     @staticmethod
     @transaction
     def enroll_from_csv(cid, form):
-        new_users, existing_uids = [], []
+        enrollment_info = [];
```
Sumukh: Nit: semi-colon

Sumukh:
LGTM

PR #719, opened 2016-03-24
Randomly generate seed data

### LINKED ISSUE #686: Enhance Seed Data
Improve seed data

knrafto:
Resolves #686. This should make it easier to spot issues that we couldn't with our naive seed data.

Some issues
* Some parameters could be played with to make the data more realistic
* Generating data takes around 10 seconds, which could be annoying

We should randomly generate data for tests, too.


Sumukh:
Did you want to resolve the TODOs in `generate.py` in this PR or leave those for later? 

Thoughts on making a utils folder? (`generate.py` could go in there). I'd like to avoid the extremely large utils.py we had in v2.

knrafto:
I'm working on comments now, because I'll need them. The rest are pretty low priority, so I think it's okay to leave them for later. We can make an issue about it.

I agree that utils.py should be broken up, but I'm not sure that generate.py should go in there. It's not used by the main app, just by manage.py.

knrafto:
Just kidding, I'll add the comment generation later, because I'm not sure if it's good without a way to view it. This should be ready for review

server/generate.py
```diff
@@ -0,0 +1,187 @@
...
+        display_name = random.choice([
+            'Hog', 'Hog Contest', 'Maps', 'Ants', 'Scheme'
+        ])
+    else:
+        display_name = '{} {}'.format(
+            random.choice(['Homework', 'Lab', 'Quiz']),
+            str(random.randrange(15)).zfill(2))
+    name = course.offering + '/' + display_name.lower().replace(' ', '')
+
+    last_night = datetime.datetime.now().replace(
```
Sumukh: I think we want `.utcnow()` - `.now()` uses the computer's local timezone.

The benefit of keeping the seed database in UTC is that we can be consistent in our TZ handling code.

Sumukh: `datetime.datetime.now().date()` will just give you the date

Sumukh:
The code looks good. The only concern is now that the names are random - it becomes a bit harder to login as a staff member to test staff only features.

* The testing login could have a login as random staff member 
* Always generate a specific staff member (somewhat defeats the point of randomness)

server/templates/testing-login.html
```diff
@@ -8,22 +8,17 @@
             Login for Testing
         </div>
         <div class="login-box-body">
-            {% for user in ['okstaff@okpy.org', 'student0@okpy.org'] %}
+            <p class="login-box-msg">Impersonate a random user</p>
```
Sumukh: Wrap in a if block in the case when there are no users. 

`{% if random_user %}`

knrafto:
Random staff member is a bit ill-defined - how about a random admin?

Sumukh:
Random admin works with me.

PR #720, opened 2016-03-28
Comment viewing and editing

knrafto:
Needs some tests, but the rest can be reviewed.

Generated comments appear on submissions.

### DIFF
--- a/server/controllers/student.py
+++ b/server/controllers/student.py
@@ -6,6 +6,7 @@ from werkzeug.exceptions import BadRequest
 
 import functools
 
+from server import highlight
 from server.constants import VALID_ROLES, STAFF_ROLES, STUDENT_ROLE
 from server.extensions import cache
 from server.forms import CSRFForm
@@ -135,7 +136,7 @@ def code(name, submit, bid):
         return abort(404)
     return render_template('student/assignment/code.html',
         course=assign.course, assignment=assign, backup=backup, diff_type=diff_type,
-        files_before=assign.files, files_after=backup.files())
+        files=highlight.diff_files(assign.files, backup.files(), diff_type))
 
 @student.route('/<assignment_name:name>/<bool(backups, submissions):submit>/<hashid:bid>/download/<file>')
 @login_required
--- a/server/controllers/student.py
+++ b/server/controllers/student.py
@@ -4,9 +4,10 @@ from flask.ext.login import login_user, logout_user, login_required, \
     current_user
 from werkzeug.exceptions import BadRequest
 
+import collections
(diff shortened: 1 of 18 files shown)


server/static/js/student.js
```diff
@@ -14,4 +25,138 @@ jQuery(document).ready(function($){
...
+            confirmButtonText: "Yes, I'm sure!",
+            closeOnConfirm: true,
+        }, function() {
+            $.ajax({
+                url: '/comments/' + comment.data('id'),
+                type: 'DELETE'
+            }).done(function () {
+                comment.fadeOut(function () { removeComment(comment) });
+            }).fail(function () {
+                swal('Oops, something went wrong. Try again?');
```
Sumukh: If you want an error icon: `swal(msg, 'error')` should do the trick.

server/controllers/student.py
```diff
@@ -237,3 +248,31 @@ def group_respond(name):
...
+        message=request.form['message'])
+    db.session.add(comment)
+    db.session.commit()
+    return render_template('student/assignment/comment.html', comment=comment)
+
+@student.route('/comments/<hashid:comment_id>', methods=['PUT', 'DELETE'])
+@login_required
+def edit_comment(comment_id):
+    comment = models.Comment.query.get(comment_id)
+    if not comment or comment.author != current_user:
```
Sumukh: Allowing admins to delete comments would be nice. 

I believe the `user` model has a `is_admin` property

Sumukh:
Thanks! I did a quick scan but I'll take a more detailed look later.

What do you think about moving comments to the API instead of giving its own prefix `/comment`? 
Putting it in the API seems somewhat appropriate for what the route does.

server/highlight.py
```diff
@@ -79,3 +100,18 @@ def format_range_unified(start, stop):
                 yield from insert(j1, j2)
             elif tag == 'equal':
                 yield from equal(i1, i2, j1, j2)
+
+def diff_files(files_before, files_after, diff_type):
```
knrafto: This can be cached, keyed on (assignment_id, backup_id, diff_type). I removed the cache on the frontend because the comments may change.

knrafto:
TODO: separate from student view so it can be used for admin as well

Sumukh:
> What do you think about moving comments to the API instead of giving its own prefix /comment? 
Do you recall what we decided about this? 

I'm going to be working off of this for composition grading (so if you think it can be merged - go for it)

Sumukh:
Merged via #736. 

TODO: separate from student view -> Has been done.

PR #738, opened 2016-06-19
Use Alpine Linux for Docker Images

Sumukh:
Using Alpine Linux reduces the size of the built images. 

Before (using Ubuntu as the base): ~800mb (~700mb from `python:3.5`)
After: 174.9mb (71.46 MB from `python:3.5-alpine`) 

Downsides: Some packages/dependencies are harder to get installed. I had to switch libraries for email formatting because installing `lxml` for python3 was not trivial. 

 If we _really_ need a package we can't get going on Alpine, we can switch back to `python:3.5`.


Sumukh:
Keeping a record of how this switch has affected the project: 

- Got rid off the LXML requirement from premailer (a good thing) 
- 😖 Alpine Linux/MUSL's C's implementation of strftime doesn't support the same zero padding formatting. `time.strftime("%-I")` will just return an empty string if the dash is used. (fixed in another PR/commit)

PR #752, opened 2016-07-10
List of queues by staff member

### LINKED ISSUE #750: Assignment grading queue view should list aggregate stats
Instead of just presenting all of the grading tasks left for an assignment, displaying queue statistics is a better idea. 

- `admin/course/1/assignments/1/queues`
  - Currently: Displays all grading tasks (grouped by staff & completeness)
  - Should: Display statistics per staff member (sorted by incomplete, ideally user sortable) with a link to jump into their queue. (This might have to be paginated)  

- `admin/course/1/assignments/1/queues/hashid(user ID 123)`
  - Currently: Doesn't exist
  - Should: Display queue for this user 

- `admin/course/1/assignments/1/queues/all`
  - Currently: Doesn't exist
  - Should: Do what `assignments/1/queues` does right now 

- Minor Enhancement: 
 Empty State for Grading Queue should be clearer about the tasks left. 

Something like this would be nice (from the wordpress Empty State)

Sumukh:
Resolves #750 

image


server/models.py
```diff
@@ -942,22 +946,65 @@ class GradingTask(Model):
...
+    @classmethod
+    def get_staff_tasks(cls, assignment_id):
+        """ Return list of namedtuple objects that represent queues.
+            Only uses 1 SQL Query.
+            Attributes:
+                - grader: User, assigned grader
+                - completed: int, completed tasks
+                - remaining: int, ungraded tasks
+        """
+        tasks = (db.session.query(cls,
```
Sumukh: It would be ideal if this only selected the grader_id and the two counts. It might also resolve the error from ONLY_GROUP_BY

PR #754, opened 2016-07-11
Keep existing flags of group members on accept

Sumukh:
When a user accepts an invite - their flagged submissions are unflagged and they "join" the group and take on whatever submission the group already had.

I think this is better than unflagging all submissions for the entire group.

### DIFF
--- a/server/models.py
+++ b/server/models.py
@@ -743,8 +743,7 @@ class Group(Model):
             raise BadRequest('{0} is not invited to this group'.format(user.email))
         with self._log('accept', user.id, user.id):
             member.status = 'active'
-        for member in self.assignment.active_user_ids(user.id):
-            self.assignment._unflag_all([member])
+        self.assignment._unflag_all([user.id])
 
     @transaction
     def decline(self, user):


Sumukh:
TODO: Review our group consistency with https://github.com/Cal-CS-61A-Staff/ok/wiki/Group-&-Submission-Consistency

Merging this PR violates: 
- Whenever a new active member is added to a group, the group final submission is updated to the most recent submission of any active member of the group.

I think this was before the idea of flagged submissions. The best analog to flagged submission is to the most recently flagged submission - but that doesn't seem ideal for the existing group members

knrafto:
Keeping the group's flagged submission seems more reasonable. Maybe we should ask some students to see what behavior they want.

How does this work if a new group is formed when a user invites another? Is the inviter's flagged submission kept?

knrafto:
Bump - I think this is a good idea. The wiki is pretty horribly out of date, so I don't think it's a blocker that it doesn't mention flagged submissions. The wiki is accurate if nobody has flagged a submission.

I created #788 to remind us to update it soon.

knrafto:
Perhaps some tests?

Sumukh:
Tests are a good idea. I'll merge this in once I add tests.

Sumukh:
I'm just going to make the tests an issue and merge this PR.

PR #761, opened 2016-07-18
PhantomJS tests for student routes 

Sumukh:
Accidently included #760 in this PR. Either that has to be merged first or those commits should be removed.


knrafto:
Hey, integration tests! Awesome :)

Sumukh:
It seems like the PhantomJS tests aren't being included in the test coverage. It needs more investigation - but it's not a blocker for this PR.

PR #763, opened 2016-07-22
Redirect to backup view if given a submission (instead of 404)

Sumukh:
Currently: `okpy.org/assign/backup/<id>` will 404 if `<id>` is a submission

It would nice to be able to provide a URL for a backup without having to know if it's a submission or not first. 

This PR's behavior: 

- If the route matches the backup's submission status: No change. 
- If the route is `/.../submission/<id>` and the backup isn't a submission: redirect to `/backup/id` 
- If the route is `/.../backup/<id>` and the backup is a submission, redirect to `/submission/id` 

Another option is to provide a single endpoint that always works regardless of submission status: 
 - `/.../code/<id>`


brianhou:
What happens if it's neither a submission nor a backup? (Like if someone makes a typo.)

Sumukh:
It would 404 (this PR maintains that). 

The behavior listed above assumes that we can find the row in the database and if the user has permission to see it.

Sumukh:
On reflection: having a `/.../code/<id>` that redirects to `/.../submit/<id>` or `../backup/<id>`  might just be better - this can be a stop-gap for now.

PR #772, opened 2016-07-30
Catch invalid HashIds in converter

### LINKED ISSUE #766: Internal server error on invalid hash IDs
https://sentry.cs61a.org/cs61a/ok-server/issues/69/

500 on http://ok.cs61a.org/cal/cs61a/su16/proj01/submissions/xxxxxx/. Should be an easy fix, maybe we can catch the ValueError [here](https://github.com/Cal-CS-61A-Staff/ok/blob/master/server/converters.py#L22)

Sumukh:
Resolves #766 

The earlier errors in Sentry were because we threw ValueErrors and didn't catch those in the converter.

### DIFF
--- a/server/controllers/api.py
+++ b/server/controllers/api.py
@@ -336,7 +336,11 @@ class ScoreSchema(APISchema):
 
     def add_score(self, user):
         args = self.parse_args()
-        backup = models.Backup.query.get(decode_id(args['bid']))
+        try:
+            bid = decode_id(args['bid'])
+        except (ValueError, TypeError):
+            restful.abort(404)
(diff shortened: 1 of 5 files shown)


server/controllers/api.py
```diff
@@ -336,7 +318,11 @@ def __init__(self):
 
     def add_score(self, user):
         args = self.parse_args()
-        backup = models.Backup.query.get(decode_id(args['bid']))
+        try:
+            bid = decode_id(args['bid'])
+        except (ValueError, TypeError):
```
Sumukh: There's probably a better way to handle this - especially since this route uses reqparse.

PR #781, opened 2016-08-08
Web Submission

Sumukh:
TODO: 
- [x] Run migration
- [x] Write PhantomJS Tests


Sumukh:
I'm think that the coveralls coverage percentage is incorrect - when I run it locally I get a coverage is 67%

PR #789, opened 2016-08-14
Serve JS/CSS libraries from CDNs

### LINKED ISSUE #723: JS and CSS assets should not be in git
We shouldn't have to copy a JS library into our repo to use it.

Options:
* Fetch all external libraries from CDNs from the client.
* Flask-Assets bundles will fetch the libraries from CDNs the first time the bundle is used. We may need to write a custom webassets resolver: https://webassets.readthedocs.org/en/latest/generic/custom_resolver.html

knrafto:
Resolves #723.

Most libraries are served from [cdnjs](https://cdnjs.cloudflare.com). pygal.js and list.fuzzysearch.js aren't on there, so I left them in the repo.

I also modified the templates so that assets are only loaded once anywhere, from one of `base.html`, `student/base.html`, `staff/base.html`, or `about/base.html`.


knrafto:
Okay, now ready for review

knrafto:
Just noticed https://github.com/Cal-CS-61A-Staff/ok/commit/76ec826a65648f8e998e6ee792dfe58df690e04b. What changes were made?

Sumukh:
I'll look at this in a bit. 






On Sun, Aug 14, 2016 at 2:46 PM -0700, "Kyle Raftogianis" <notifications@github.com> wrote:












Just noticed 76ec826. What changes were made?

Sumukh:
re: 76ec826 - I changed how it labels pages. It's not a critical fix - I think a good solution would be to serve list.js from the CDN and host pagination plugin ourselves. 

Before it was just ... 12, 13, 14 ...  but after the change it was ... Page 12 of 50, 13, 14  ... 

It doesn't appear like the published versions provide an interface to change the name, so I just edited the source. There's an open PR on https://github.com/javve/list.pagination.js/pull/13 to improve that behavior, but that's also not a CDN (the commit mentions forking the pagination library and merging in those changes)

FYI: Not really related to this PR, but the static assets are being served via [Google Cloud CDN](https://cloud.google.com/cdn/). (I've temporarily disabled it though) 

I still haven't looked at this PR but I will soon.

server/templates/staff/course/enrollment.html
```diff
@@ -178,3 +166,28 @@ <h3 class="box-title">Add/Edit Participant</h3>
 
                 <!-- </body> do not close body in template-->
                 {% endblock %}
+
+{% block page_js %}
+<script>
+  $(document).ready(function() {
```
knrafto: You can remove the `$(document).ready` wrapper. I added that only because list.js wasn't loaded at the time this script was run, but it should be now.

Sumukh:
I made some changes to allow for one off resources for pages using a `page_js` block (inside of the `js` block).

The other changes look good to me. 

My favorite part of this PR: image

PR #790, opened 2016-08-17
Groups API 

Sumukh:
Adds group & assignment endpoints to the API. 

Used by the collaborative editor. 

Also: 
- Changed other endpoints to avoid needing to know the database ID. 
- Cleanup
- Documentation updates

### DIFF
--- a/documentation/api.md
+++ b/documentation/api.md
@@ -207,10 +207,6 @@ curl "https://ok.cs61a.org/api/v3/assignment/cal/cs61a/sp16/lab01/group/email@ex
 {
     "code": 200,
     "data": {
-        "assignment": {
-            "course": { ... },
-            "name": "cal/cs61a/sp16/lab01"
-        },
         "members": [
             {
                 "created": "2016-08-16T20:54:28",
@@ -252,7 +248,6 @@ access_token | None | (Required) Access Token of staff member
 Parameter | Type | Description
 ---------- | ------- | -------
 members | List | List of members if available. If there is no group, this value is equal to `[]`
-assignment | Dictionary | The assignment name and the course info.
 
 See example output for fields.
 `member['status']` is either "pending" or "active"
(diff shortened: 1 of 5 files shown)


server/models.py
```diff
@@ -266,9 +266,10 @@ def can(cls, obj, user, action):
             return False
         if user.is_admin:
             return True
+        is_staff = user.is_enrolled(obj.course.id, STAFF_ROLES)
         if action == "view":
-            return user.is_authenticated
-        return user.is_enrolled(obj.course.id, STAFF_ROLES)
+            return is_staff or obj.visible
```
Sumukh: The student endpoint doesn't do a `.can` check, so this won't affect the behavior of the assignment page on the student dashboard.

Sumukh:
@knrafto - feel free to review if you want to. I'm going to merge later today if not.

documentation/api.md
```diff
@@ -152,10 +152,115 @@ See example response.
...
+See example output for other fields.
+
+## Group Info
+>><h4> Example Response </h4>
+> ```
+curl "https://ok.cs61a.org/api/v3/assignment/cal/cs61a/sp16/lab01/group/email@example.com"
+{
+    "code": 200,
+    "data": {
+        "assignment": {
```
knrafto: We shouldn't send assignment info, because this will send the file templates as well. The client should make a separate request for the assignment info and cache that. In general, I'd like to keep the endpoints lightweight to avoid the get_course_info fiasco.

server/controllers/api.py
```diff
@@ -611,12 +634,69 @@ def get(self, name=None):
...
+    Used by: Collaboration/Scripts
+    """
+    model = models.Group
+    schema = GroupSchema()
+
+    @marshal_with(schema.get_fields)
+    def get(self, user, name, email):
+        assign = models.Assignment.by_name(name)
+        target = models.User.lookup(email)
+        default_value = {'assignment': assign, 'members': []}
```
knrafto: Ditto, should not send assignment

knrafto:
+1 for using names instead of database keys. LGTM except the extraneous assignment info

PR #796, opened 2016-08-20
OAuth Provider (for Login via Ok)

### LINKED ISSUE #794: Single sign-on
Many apps use the OK api, and you currently have to remember to use the right Google email for all of them. This is very easy and to mess up. Instead, they should only have to log in through OK.

Apps that could use this:
* ok web interface
* ok-client
* office hours queue
* collaborative editor
* lab assistant manager

Sumukh:
ok-oauth-2

This creates a standard OAuth flow for client apps to request login via OK. An example client is provided in documentation. 

There are two scopes currently: `email` or `all`. 
- `all` is like a google access token - It gives all access
- `email` only provides support for the `/api/v3/user/` endpoint. This is useful so apps can use this as a form of identity without getting all permissions. 

Resolves #794 

TODO before merge:
- [x] Add CSRF for OAuth routes
- [x] Add admin side form to create clients
- [x] Ability to log in as someone else from confirmation screen
- [x] Proper login flow if user is not already logged in


knrafto:
Not forcing reauthentication will make it much harder to use the wrong email. We have a whitelist of clients and a second confirmation page, so I think it'll be alright.

knrafto:
I'm pulling my hair out trying to figure out why it forces you to log in again. I think Flask is just ignoring the session cookie on `oauth/authorize`, or the browser is not sending it? I'm not sure.

Sumukh:
I know why. I was also pulling my hair out when building this

Use two different IPs for logging in. Login toOk with local host, access the example with 127.0.0.1

Flask overwrites the session otherwise

knrafto:
Oh my god >_<

There should also be a way to log in as a different from the confirm screen, just in case

knrafto:
Is `_formhelpers.html` a copy of `staff/_formhelpers.html`? Why?

server/controllers/oauth.py
```diff
@@ -0,0 +1,111 @@
...
+    db.session.add(tok)
+    db.session.commit()
+    return tok
+
+@oauth.route('/oauth/authorize', methods=['GET', 'POST'])
+@oauth_provider.authorize_handler
+@login_required
+def authorize(*args, **kwargs):
+    # Only CSRF protect this route.
+    if not use_testing_login():
```
Sumukh: This should be more specific. something like if not csrf_disabled:

Sumukh:
Yeah - it's a copy. I've removed it now.

I moved it there because I thought we could consolidate the file there - but I'll just use the original file.

Sumukh:
One last thing before merging: Generate the alembic migration and run it on prod.

Sumukh:
The DB migration has been run.

PR #800, opened 2016-08-23
Course upload/edit forms

### LINKED ISSUE #735: Create a new course via Web Interface (v3)
The admin view doesn't have an ability to create new courses (as far as I recall)

It should be a form accessible to admin users 
https://github.com/Cal-CS-61A-Staff/ok/blob/v3/master/server/controllers/admin.py

Similiar Forms in Admin (for implementation guidance)
Create Assignments (https://github.com/Cal-CS-61A-Staff/ok/tree/master/server/controllers/admin.py#L283) 
Create Version (https://github.com/Cal-CS-61A-Staff/ok/tree/master/server/controllers/admin.py#L75)

Sumukh:
Resolves #735 

- Create courses
- Edit some fields for courses (like Display Name, URL, Active) 
- Minor UI improvements to sidebar 

Questions:
- Do we need to edit a timezone of a course? (I've hidden it for now, but I'm not sure if its useful)


server/utils.py
```diff
@@ -163,3 +172,13 @@ def generate_csv(query, items, selector_fn):
             data.update(dict)
         csv_writer.writerow(data)
         yield csv_file.getvalue()
+
+def is_valid_endpoint(endpoint, valid_format):
+    """Validates an endpoint name against a regex pattern VALID_FORMAT. """
+    r = re.compile(valid_format)
+    is_forbidden = any([endpoint.startswith(name) for name
```
Sumukh: Some tests for this function is a good idea.

server/utils.py
```diff
@@ -77,6 +79,13 @@ def is_safe_redirect_url(request, target):
     return redirect_url.scheme in ('http', 'https') and \
         host_url.netloc == redirect_url.netloc
 
+def str_to_tz(tz_str):
+    if tz_str not in pytz.common_timezones_set:
+        logger.warning("Invalid TZ choice {}".format(tz_str))
+        return None
```
Sumukh: This isn't a case we'll hit since form validation ensures timezone is valid. 

The other approach is to type check in the SQLAlchemy converter (and support String (Python) -> String (DB) storage instead of having to do String (HTML) -> Timezone (Python) -> String (DB)

server/constants.py
```diff
@@ -10,6 +10,9 @@
               'revision']
 API_PREFIX = '/api'
 
+COURSE_ENDPOINT_FORMAT = '^\w+/\w+/\w+$'
+ASSIGNMENT_ENDPOINT_FORMAT = COURSE_ENDPOINT_FORMAT[:-1] + '/\w+$'
```
knrafto: We should move the regexes in converters.py here instead so there's not two different ones

server/controllers/admin.py
```diff
@@ -255,6 +256,31 @@ def client_version(name):
...
+# Course #
+##########
+@admin.route("/course/new", methods=['GET', 'POST'])
+@is_staff()
+def create_course():
+    courses, current_course = get_courses()
+    form = forms.NewCourseForm()
+    if form.validate_on_submit():
+        new_course = Course()
+        form.timezone.data = utils.str_to_tz(form.timezone.data)
```
knrafto: Seems a little hacky...what about a custom TimeZoneField? I'm not sure where else it would be used though

server/templates/staff/sidebar.html
```diff
@@ -20,7 +20,7 @@
             {% for course in courses %}
               <li class="treeview  {% if (request.endpoint == "admin.index" and loop.length == 1) %}active default{% endif %} {% if request.view_args['cid'] == course.id %} active{% endif %}" >
                 <a href="#">
-                  <i class="fa fa-university"></i> <span>{{ course.display_name_with_semester }}</span>
+                  <i class="fa fa-university"></i> <span>{{ course.display_name_with_semester | truncate(25, True) }}</span>
```
knrafto: Does this line wrap if it's too long? Another option would be a text-overflow css property

server/constants.py
knrafto: The one in converters.py also takes care of forbidden assignment names, etc, so is_valid_endpoint would just be a reflex match.

server/controllers/admin.py
knrafto: Saw your note below of making the sqlalchemy time zone column be a string. I like that too and that's probably less work

server/constants.py
Sumukh: I agree that this is a good idea - I've left a TODO in constants.py for now.

server/templates/staff/sidebar.html
Sumukh: image
I tried text-overflow for a bit and it didn't do what I wanted - so I'm going to stick with this.

PR #805, opened 2016-08-23
fixed unicode error in generate.py

kyletliang:


server/generate.py
```diff
@@ -15,7 +15,7 @@
 from server.extensions import cache
 
 original_file = open('tests/files/fizzbuzz_before.py').read()
```
knrafto: You should probably change this one too...

knrafto:
Thanks! Just tested it on my machine too, it works.

Issue #801, opened 2016-08-23
Backup JSON in API responses should have IDs


knrafto:
See https://okpy.github.io/documentation/ok-api.html#assignments-export-backups
The backup json object should have `"id": "38DK3J"` or similar

PR #806, opened 2016-08-24
IDs for Backup JSON

kyletliang:
Addresses Issue #801


Sumukh:
LGTM. 

A side effect of this is that research exports will include the backup ID. I think that's fine since that alone doesn't lead to any confidential information leakage.

Issue #718, opened 2016-03-23
Automatically generate assignment names


knrafto:
Instead of making staff members enter assignment names in the form, it should always be `course.offering + '/' + display_name.lower().replace(' ', '')`.

Sumukh:
The only reason this taking a while is because I'd like the UI to indicate this as the assignment is being created. 

If displaying the endpoint in the success message after creation is sufficient, this should just be a 4-5 line change.

ericmnelson:
Hey I just joined in here, but to clarify are we looking to autofill the current form element using the naming formula above and them have them be able to change it, or to take that element out and auto-populate it on creation and then tell the user where the assignment was created?

Sumukh:
Either of those is acceptable. As long as the endpoint is easily copy/pasted - I think users won't mind. 

The former is ideal (where the form indicates what the endpoint is going to be) - but it's more difficult and probably not worth it.

Sumukh:
Resolved in #890

PR #814, opened 2016-08-25
Automatically generate assignment names #718

ericmnelson:
Tried just a simple javascript to update the name based on the display name. Not entirely sure if this is what you guys were going for but helps it autofill. Lemme know thoughts.

### DIFF
--- a/server/static/js/staff.js
+++ b/server/static/js/staff.js
@@ -1,4 +1,11 @@
 // initialize datepicker
 jQuery(document).ready(function($){
   $('.datepicker').datetimepicker({format:'YYYY-MM-DD hh:mm:ss'});
+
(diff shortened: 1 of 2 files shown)


knrafto:
Resolves #718 (so GitHub auto-closes the issue)

I played around with it - this is pretty much exactly what we need :) LGTM

Sumukh:
Neat. LGTM 

For the future we might want to consider this styling - but I think it'd be too much work to get that to work with WTForms + we'd have to change how the server handles the endpoint. 

image

```
<div class="input-group">
                <span class="input-group-addon">cal/cs61a/sp16/</span>
                <input type="text" class="form-control" placeholder="hog">
</div>
````

PR #816, opened 2016-08-26
Create lab assistant role

### LINKED ISSUE #799: Create lab assistant role
To be used by the lab assistant manager and the OH queue. We should be able to enroll people as lab assistants from the admin interface, too.

knrafto:
Resolves #799.

Opening to track progress.


knrafto:
Should we display lab assistants in the same table as the other staff members? It would simplify some of the logic (since some courses won't have lab assistants). It's a little weird that lab assistants are displayed in the staff table, but permissions-wise are not treated as staff members, but I think that's okay.

Sumukh:
Upon a quick glance: Looks good. 

@knrafto and I will take a closer look later.

tests/test_enrollment.py
```diff
@@ -13,21 +13,45 @@ def setUp(self):
...
+            'role': STUDENT_ROLE
+        }
+        self.lab_assistantA = {
+            'name': 'Ned Stark',
+            'email': 'eddard.stark@winterfell.com',
+            'sid': '152342343',
+            'section': '101',
+            'role': LAB_ASSISTANT_ROLE
+        }
+        self.lab_assistantA = {
```
Sumukh: This looks duplicated/should be removed?

raychan3: Yes cleaning up and merging master over to resolve conflicts.

raychan3:
Separating staff and lab assistants doesn't really have too complicated logic. I thought about having it in staff initially, but it made more sense to have it separate, especially if there are use cases that deal with only staff.

server/constants.py
```diff
@@ -4,7 +4,8 @@
 GRADER_ROLE = 'grader'
 STAFF_ROLE = 'staff'
 INSTRUCTOR_ROLE = 'instructor'
-VALID_ROLES = [STUDENT_ROLE, GRADER_ROLE, STAFF_ROLE, INSTRUCTOR_ROLE]
+LAB_ASSISTANT_ROLE = 'lab_assistant'
+VALID_ROLES = [STUDENT_ROLE, GRADER_ROLE, STAFF_ROLE, INSTRUCTOR_ROLE, LAB_ASSISTANT_ROLE]
```
knrafto: These seem to be in order of increasing permissions, so `LAB_ASSISTANT_ROLE` should be second

knrafto:
Can you generate some lab assistants too to make it easier to test manually? https://github.com/Cal-CS-61A-Staff/ok/blob/master/server/generate.py#L144

knrafto:
The underscore in `lab_assistant` shows up in the UI and it's slightly annoying, but I think a good solution is out of scope of this PR. I'll make another issue for it.

Sumukh:
re: `lab_assistant` Is it the worst if we have it be two words? I wanted it to be one word so we would be consistent - but I'm willing to have it be two words if it saves future effort.

knrafto:
Yeah, two words would be easier, and I can't really think of any drawbacks.

Sumukh:
Going to merge & deploy in a bit.

PR #817, opened 2016-08-26
[WiP] Automatically generate assignment names

### LINKED ISSUE #718: Automatically generate assignment names

knrafto:
Opening to track progress.

Resolves #718.


knrafto:
Currently this breaks when assignments are edited. Assignments names should _not_ be automatically generated when editing an assignment.

Possible solutions:
* Play with HTML IDs and jQuery selectors so the JS only gets activated on the assignment creation form
* Use {% block page_js %} to only include the JS on the assignment creation form

Sumukh:
Closing via #890

PR #818, opened 2016-08-26
[WiP] Datepicker for time fields

### LINKED ISSUE #785: Better Date/Time Picker
Typing dates in a specific format is not really the best way to select dates/times. 
Using one of the extension date/time picker libraries would make this easier.

knrafto:
Resolves #785.

Currently there's no way to specify PM times.

### DIFF
--- a/server/static/js/staff.js
+++ b/server/static/js/staff.js
@@ -1,6 +1,6 @@
 // initialize datepicker
 jQuery(document).ready(function($){
-  // $('.datepicker').datetimepicker({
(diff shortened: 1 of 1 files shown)


Sumukh:
Do staff members really want a date-picker? When we had it in prod - it sounded like it just made the TA experience harder. Maybe we just need a format validator?

Sumukh:
I'm just going to close this PR. The issue is still open so we can have it as a reminder.

PR #819, opened 2016-08-27
Use 'roster' variable instead of hardcoding 'enrollment'

### LINKED ISSUE #809: Total number of staff members is incorrect
On the enrollment page, the "Total: " at the bottom should be the number of staff members, but it looks like it's the number of students.

kevinlin1:
Closes #809.


Sumukh:
LGTM. Merging. 

Side note: We should probably rename this to `enrollment.list.html` now that it's used for more than students.

PR #824, opened 2016-08-28
rnl - Added the ability to unenroll students / staff from courses.

### LINKED ISSUE #813: Add ability to un-enroll a student
I accidentally enrolled someone in Summer 2016. Not a big deal, but it would be nice to unenroll them.

It could also be useful if we accidentally enroll someone as a staff member and need to revoke access. This could be done by re-enrolling them as a student, but I haven't tried that.

Rohit-L:
In this PR:

- Added "Unenroll" button to student / staff list row
- Endpoint (POST) to unenroll a student
- Flash message indicating student / staff was unenrolled
- Modal confirm dialog for unenrollment

Resolves #813

### DIFF
--- a/server/controllers/admin.py
+++ b/server/controllers/admin.py
@@ -651,11 +651,11 @@ def enrollment(cid):
 @is_staff(course_arg='cid')
 def unenrollment(cid, user_id):
     user = User.query.filter_by(id=user_id).one_or_none()
-    enrollment = user.is_enrolled(cid);
-    if user and enrollment:
-        enrollment.unenroll()
-        cache.delete_memoized(User.is_enrolled, user, cid)
-        flash("{email} has been unenrolled".format(email=user.email), "success")
+    if user:
+        enrollment = user.is_enrolled(cid);
+        if enrollment:
+            enrollment.unenroll()
+            flash("{email} has been unenrolled".format(email=user.email), "success")
     
     return redirect(url_for(".enrollment", cid=cid))
 
--- a/server/controllers/admin.py
+++ b/server/controllers/admin.py
@@ -656,6 +656,8 @@ def unenrollment(cid, user_id):
         if enrollment:
             enrollment.unenroll()
             flash("{email} has been unenrolled".format(email=user.email), "success")
+        else:
+            flash("{email} is not enrolled".format(email=user.email), "warning")
     
     return redirect(url_for(".enrollment", cid=cid))
 
--- a/server/controllers/admin.py
+++ b/server/controllers/admin.py
@@ -658,7 +658,9 @@ def unenrollment(cid, user_id):
             flash("{email} has been unenrolled".format(email=user.email), "success")
         else:
             flash("{email} is not enrolled".format(email=user.email), "warning")
-    
+    else:
+        flash("Unknown user", "warning")
+  
     return redirect(url_for(".enrollment", cid=cid))
 
 @admin.route("/course/<int:cid>/enrollment/batch",
--- a/server/controllers/admin.py
+++ b/server/controllers/admin.py
@@ -653,7 +653,8 @@ def unenrollment(cid, user_id):
     user = User.query.filter_by(id=user_id).one_or_none()
     enrollment = user.is_enrolled(cid);
     if user and enrollment:
-        Enrollment.unenroll(enrollment)
+        enrollment.unenroll()
+        cache.delete_memoized(User.is_enrolled, user, cid)
         flash("{email} has been unenrolled".format(email=user.email), "success")
     
     return redirect(url_for(".enrollment", cid=cid))
--- a/server/controllers/admin.py
+++ b/server/controllers/admin.py
@@ -643,8 +643,20 @@ def enrollment(cid):
                            enrollments=students, staff=staff,
                            lab_assistants=lab_assistants,
                            form=form,
+                           unenroll_form=forms.CSRFForm(),
                            courses=courses,
                            current_course=current_course)
+
+@admin.route("/course/<int:cid>/<int:user_id>/unenroll", methods=['POST'])
+@is_staff(course_arg='cid')
+def unenrollment(cid, user_id):
+    user = User.query.filter_by(id=user_id).one_or_none()
+    enrollment = user.is_enrolled(cid);
+    if user and enrollment:
+        Enrollment.unenroll(enrollment)
+        flash("{email} has been unenrolled".format(email=user.email), "success")
+    
+    return redirect(url_for(".enrollment", cid=cid))
 
 @admin.route("/course/<int:cid>/enrollment/batch",
              methods=['GET', 'POST'])
--- a/server/models.py
+++ b/server/models.py
@@ -524,12 +524,9 @@ class Enrollment(Model):
         }
         Enrollment.create(cid, [info], role)
 
-    @staticmethod
     @transaction
-    def unenroll(enrollment):
-        db.session.delete(enrollment)
-        db.session.commit()
-        cache.delete_memoized(User.is_enrolled)
+    def unenroll(self):
+        db.session.delete(self)
 
     @staticmethod
     @transaction
--- a/server/models.py
+++ b/server/models.py
@@ -526,6 +526,7 @@ class Enrollment(Model):
 
     @transaction
     def unenroll(self):
+        cache.delete_memoized(User.is_enrolled, self.user, self.course.id)
         db.session.delete(self)
 
     @staticmethod
--- a/server/models.py
+++ b/server/models.py
@@ -526,6 +526,13 @@ class Enrollment(Model):
 
     @staticmethod
     @transaction
+    def unenroll(enrollment):
+        db.session.delete(enrollment)
+        db.session.commit()
+        cache.delete_memoized(User.is_enrolled)
+
+    @staticmethod
+    @transaction
     def enroll_from_csv(cid, form):
         enrollment_info = []
         rows = form.csv.data.splitlines()
--- a/server/static/js/main.js
+++ b/server/static/js/main.js
@@ -19,3 +19,18 @@ if (typeof HTMLElement.prototype.removeClass !== "function") {
     }
 }
 
+$('body').on('click', 'button[data-confirm]', function(e) {
+  e.preventDefault();
+
+  var form = $(this).parents('form');
+  var confirmText = $(this).attr('data-confirm');
+
+  swal({
+    title: confirmText,
+    showCancelButton: true,
+    confirmButtonText: "Yes, I'm sure!",
+    closeOnConfirm: true
+  }, function(isConfirm){
+      if (isConfirm) form.submit();
+  });
+})
--- a/server/static/js/main.js
+++ b/server/static/js/main.js
@@ -23,7 +23,7 @@ $('body').on('click', 'button[data-confirm]', function(e) {
   e.preventDefault();
 
   var form = $(this).parents('form');
-  var confirmText = $(this).attr('data-confirm');
+  var confirmText = $(this).data('confirm');
 
   swal({
     title: confirmText,
--- a/server/templates/staff/course/students.list.html
+++ b/server/templates/staff/course/students.list.html
@@ -21,6 +21,7 @@
                 <th class="sort" data-sort="role">Role</th>
                 <th class="sort" data-sort="timestamp">Enrolled At</th>
                 <th class="sort" data-sort="section">Section</th>
+                <th data-sort="unenroll">Action</th>
             </thead>
 
             <tbody class="list">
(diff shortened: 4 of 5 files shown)


server/controllers/admin.py
```diff
@@ -637,9 +637,19 @@ def enrollment(cid):
...
                            single_form=single_form,
+                           unenroll_form=forms.CSRFForm(),
                            courses=courses,
                            current_course=current_course)
 
+@admin.route("/course/<int:cid>/<string:user_id>/unenroll", methods=['POST'])
+@is_staff(course_arg='cid')
+def unenrollment(cid, user_id):
+    user = User.query.filter_by(id=user_id).one_or_none()
+    if user and user.is_enrolled(cid):
```
knrafto: What happens if the user is not enrolled?

server/templates/staff/course/students.list.html
```diff
@@ -21,6 +21,7 @@ <h3 class="box-title"><span> {{ title }} </span></h3>
                 <th class="sort" data-sort="role">Role</th>
                 <th class="sort" data-sort="timestamp">Enrolled At</th>
                 <th class="sort" data-sort="section">Section</th>
+                <th class="sort" data-sort="unenroll">Action</th>
```
knrafto: This column shouldn't be sortable. It does some weird stuff if you try to sort by this column

server/controllers/admin.py
knrafto: I managed to get a 500 by opening the enrollment page in two different tabs and trying to delete the same user from both of them

tests/test_enrollment.py
```diff
@@ -51,6 +51,26 @@ def test_enroll_from_form(self):
...
+        user = User(name=self.studentA['name'], email=self.studentA['email'])
+        db.session.add(user)
+        db.session.commit()
+        self.studentA['id'] = user.id
+        
+        Enrollment.create(self.course.id, [self.studentA])
+        enrollment = Enrollment.query.filter_by(
+            course_id=self.course.id, 
+            user_id=user.id).one_or_none()
+        assert enrollment
```
Sumukh: It's better to explicit about what we expect. In this case it looks like you are asserting that `enrollment` is not None and the later one you are asserting that it should be None. 

If so, `self.assertIsNotNone` and `self.assertIsNone` would be better to use

https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertIsNotNone

server/controllers/admin.py
```diff
@@ -637,9 +637,19 @@ def enrollment(cid):
     return render_template('staff/course/enrollment.html',
                            enrollments=students, staff=staff,
                            single_form=single_form,
+                           unenroll_form=forms.CSRFForm(),
                            courses=courses,
                            current_course=current_course)
 
+@admin.route("/course/<int:cid>/<string:user_id>/unenroll", methods=['POST'])
```
Sumukh: Why is the `user_id` a string? 

Instead of having a specific endpoint for this route, it'd probably be better have that ID be part of the form data.

server/templates/staff/course/students.list.html
```diff
@@ -37,6 +38,11 @@ <h3 class="box-title"><span> {{ title }} </span></h3>
                     <td class="role"><span class="label label-success">{{item.role}}</span></td>
                     <td class="timestamp" data-timestamp="{{ item.created }}">{{ utils.local_time(item.created, current_course) }}</td>
                     <td class="section">{{item.section}}</td>
+                    <td class="unenroll">
+                        {% call forms.render_form(unenroll_form, action_url=url_for('.unenrollment', cid=current_course.id, user_id=item.user.id), action_text='Unenroll',
```
Sumukh: The traditional action button takes a lot of space. I'd prefer a smaller button with an icon. 

If you use `render_form_bare` you can get rid of the action_text and specify your own submit button inside of the call block. 

Here's what I had disabled before - I think it worked pretty well. 
` <button type="submit" class="btn btn-sm btn-danger"> <i class="fa fa-trash-o"></i></button>`
image

server/models.py
```diff
@@ -526,6 +526,13 @@ def enroll_from_form(cid, form):
 
     @staticmethod
     @transaction
+    def unenroll(cid, user_id):
+        enrollment = Enrollment.query.filter_by(course_id=cid, user_id=user_id).one_or_none()
+        db.session.delete(enrollment)
```
Sumukh: What happens when the enrollment doesn't exist? 

```
>>> db.session.delete(None)
 db.session.delete(None)
*** sqlalchemy.orm.exc.UnmappedInstanceError: Class 'builtins.NoneType' is not mapped
```

Maybe return None if the user isn't enrolled, otherwise return True

tests/test_enrollment.py
```diff
@@ -51,6 +51,26 @@ def test_enroll_from_form(self):
...
+
+        Enrollment.unenroll(self.course.id, user.id)
+        enrollment = Enrollment.query.filter_by(
+            course_id=self.course.id, 
+            user_id=user.id).one_or_none()
+        assert not enrollment
```
Sumukh: Also worth testing: Unenrolling a user.id that isn't enrolled and making sure that doesn't crash.

server/controllers/admin.py
Sumukh: `user.is_enrolled` actually returns the enrollment object if they are enrolled (though it's hard to tell from the name). 

Given that it returns the enrollment object - we don't need to query for it again in Enroll.unenroll. Instead maybe we can make unenroll a method for enrollment objects. That way we can do something like 
```
enrollment = user.is_enrolled(cid)
if enrollment:
   enrollment.unenroll() #
else:
   ... 
```

There's something else to consider in that all enroll does is a delete - so we don't really need a method for it. We could just call `enrollment.delete()` and be done with it. For now, I'm okay with still making an `unenroll` method since we may want to do other things in the future

server/templates/staff/course/students.list.html
knrafto: Should we have a confirm dialog too? Seems like it's easy to accidentally delete a student

Sumukh: a confirm dialog would be nice. It'd be nice to standardize on a way to get a form to have a confirm dialog - since we could use it multiple places. We might want to leave it out of the scope of this PR - (@Rohit-L - if you want to add this - feel free to) 

On the student side - we do this with a jQuery function. Something similar for the admin panel but using data attributes on the form would be neat. 

https://github.com/Cal-CS-61A-Staff/ok/blob/eafb805028c9ccb5812fdd9c6ce89e784cf6d5ba/server/static/js/comments.js#L13

`<button type="submit" class="btn form-confirm" data-confirm="Are you sure you want to delete student@example.com?">`

server/controllers/admin.py
Rohit-L: I though it made sense to have the user_id be a part of the route from an API standpoint. However, if you think it really should be a part of the form data, I'd be happy to change it. Let me know 👍 

Also changed user_id to an integer.

server/controllers/admin.py
Rohit-L: Thanks for the comments. I didn't realize that there was a layer of caching that was involved! That is what created the 500 error. I have modified it so that when the user is unenrolled, the cache for `is_enrolled` is also cleared. Additionally, I modified the code to take advantage of the fact that `user.is_enrolled` returns the enrollment object.

server/models.py
Rohit-L: Fixed in another comment. This only occurred because the cache for `is_enrolled

server/templates/staff/course/students.list.html
Rohit-L: Fixed. Thank you! 👍

server/templates/staff/course/students.list.html
Rohit-L: Added confirm dialog based on data-confirm attribute.

Rohit-L:
@Sumukh @knrafto I have addressed the comments. Let me know if there is anything else! 😃

Sumukh:
I'll take a look in a bit. For the future, having multiple commits in a PR is fine - we will squash the commits when merging if we need to. We try to avoid force pushing as much as possible.

server/controllers/admin.py
```diff
@@ -637,9 +637,21 @@ def enrollment(cid):
...
                            courses=courses,
                            current_course=current_course)
 
+@admin.route("/course/<int:cid>/<int:user_id>/unenroll", methods=['POST'])
+@is_staff(course_arg='cid')
+def unenrollment(cid, user_id):
+    user = User.query.filter_by(id=user_id).one_or_none()
+    enrollment = user.is_enrolled(cid);
+    if user and enrollment:
+        Enrollment.unenroll(enrollment)
```
Sumukh: This should just be a method instead of a class method. 
`enrollment.unenroll()`

server/models.py
```diff
@@ -526,6 +526,13 @@ def enroll_from_form(cid, form):
 
     @staticmethod
     @transaction
+    def unenroll(enrollment):
+        db.session.delete(enrollment)
+        db.session.commit()
+        cache.delete_memoized(User.is_enrolled)
```
Sumukh: We can specify the arguments to invalidate just one entry instead of clearing the entire cache.

server/static/js/main.js
```diff
@@ -19,3 +19,18 @@ if (typeof HTMLElement.prototype.removeClass !== "function") {
     }
 }
 
+$('body').on('click', 'button[data-confirm]', function(e) {
+  e.preventDefault();
+
+  var form = $(this).parents('form');
+  var confirmText = $(this).attr('data-confirm');
```
Sumukh: `$(this).data('confirm')` might also work. I'm not actually sure which one is better. I think this is sufficient for our use case.

server/static/js/main.js
```diff
@@ -19,3 +19,18 @@ if (typeof HTMLElement.prototype.removeClass !== "function") {
...
+
+  swal({
+    title: confirmText,
```
Sumukh: Might be better as the message instead of the title. I'll play around with it live and see which one is better.

server/models.py
```diff
@@ -526,6 +526,13 @@ def enroll_from_form(cid, form):
 
     @staticmethod
     @transaction
+    def unenroll(enrollment):
+        db.session.delete(enrollment)
+        db.session.commit()
```
Sumukh: It's not necessary to commit if we are using the `@transaction` decorator.

Sumukh:
Another test worth adding is posting to the unenrollment route, see test_version.py for an example of how to test submitting a form.

server/static/js/main.js
Rohit-L: Sure thing!

server/models.py
Rohit-L: Thank you! Noted.

server/static/js/main.js
Rohit-L: Fixed.

server/controllers/admin.py
Rohit-L: I see. Thank you. Fixed.

server/models.py
```diff
@@ -524,12 +524,9 @@ def enroll_from_form(cid, form):
...
         Enrollment.create(cid, [info], role)
 
-    @staticmethod
     @transaction
-    def unenroll(enrollment):
-        db.session.delete(enrollment)
-        db.session.commit()
-        cache.delete_memoized(User.is_enrolled)
+    def unenroll(self):
+        db.session.delete(self)
```
Sumukh: I think it's better if keep the cache clearing cache here (and not in the route) - so that others can use this method without forgetting to clear the cache.

Sumukh:
This is pretty close to merging. 

To merge: 
- [ ] Move cache clearing into the method 
- [ ] Resolve the merge conflict.  
- [ ] Test the route with a form submissions (like in `test_version.py`) (Optional)

server/models.py
Rohit-L: Moved in the next commit.

Rohit-L:
@Sumukh Made final changes.

server/controllers/admin.py
```diff
@@ -643,9 +643,22 @@ def enrollment(cid):
...
+@admin.route("/course/<int:cid>/<int:user_id>/unenroll", methods=['POST'])
+@is_staff(course_arg='cid')
+def unenrollment(cid, user_id):
+    user = User.query.filter_by(id=user_id).one_or_none()
+    if user:
+        enrollment = user.is_enrolled(cid);
+        if enrollment:
+            enrollment.unenroll()
+            flash("{email} has been unenrolled".format(email=user.email), "success")
+
```
Sumukh: Sorry one more thing
else:

Sumukh:
Looks good overall. I'll try this locally and then deploy & merge.

server/controllers/admin.py
Rohit-L: Added.

Rohit-L:
Ready to Merge.

PR #825, opened 2016-08-28
New assignment form datepicker and placeholder text

### LINKED ISSUE #810: New assignment form should default to 11:59:59 pm.
The new assignment form didn't default to the current time, but rather to a week from now at 11:59:59. 

It's pretty unlikely that a user wants assignments to be due exactly at the time they hit new assignment 
image

Approach suggestions: 
- Set the default value in the template when we render the field : `templates/staff/assignment/assignment.new.html` 
- Assume that we are using PST and set the default in NewAssignmentForm in `forms.py` 

Due date default to a week from now at 11:59:59pm
Lock dates should default to be a day after the default due date at 11:59:59pm

kevinlin1:
Resolves #810. Can someone check my code composition?

screen shot 2016-08-28 at 10 35 00

### DIFF
--- a/server/templates/staff/course/assignment.new.html
+++ b/server/templates/staff/course/assignment.new.html
@@ -35,7 +35,7 @@
                   {{ forms.render_field(form.max_group_size, label_visible=true, type='number', min='1', value='1') }}
                   {{ forms.render_field(form.due_date, label_visible=true, type='text', class='form-control datepicker') }}
                   {{ forms.render_field(form.lock_date, label_visible=true, type='text', class='form-control datepicker') }}
-                  {{ forms.render_field(form.url, label_visible=true, placeholder='http://cs61a.org/proj/hog/', type='text') }}
+                  {{ forms.render_field(form.url, label_visible=true, placeholder='http://cs61a.org/proj/hog/ (Optional)', type='text') }}
                   {{ forms.render_checkbox_field(form.visible, label_visible=true) }}
                   {{ forms.render_field(form.autograding_key, label_visible=true, placeholder='5abc... (Optional)', type='text') }}
                   {{ forms.render_checkbox_field(form.revisions_allowed, label_visible=true) }}
--- a/server/templates/staff/course/assignment.new.html
+++ b/server/templates/staff/course/assignment.new.html
@@ -33,8 +33,8 @@
                   {{ forms.render_field(form.display_name, label_visible=true, placeholder='Hog', type='text') }}
                   {{ forms.render_field(form.name, label_visible=true, placeholder='cal/cs61a/fa16/proj01', type='text') }}
                   {{ forms.render_field(form.max_group_size, label_visible=true, type='number', min='1', value='1') }}
-                  {{ forms.render_field(form.due_date, label_visible=true, type='text', class='form-control datepicker') }}
-                  {{ forms.render_field(form.lock_date, label_visible=true, type='text', class='form-control datepicker') }}
+                  {{ forms.render_field(form.due_date, label_visible=true, placeholder=utils.new_due_date(current_course), type='text', class='form-control datepicker') }}
+                  {{ forms.render_field(form.lock_date, label_visible=true, placeholder=utils.new_lock_date(current_course), type='text', class='form-control datepicker') }}
                   {{ forms.render_field(form.url, label_visible=true, placeholder='http://cs61a.org/proj/hog/ (Optional)', type='text') }}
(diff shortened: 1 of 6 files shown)


Sumukh:
Cool! - there is a small hitch though. 

This will work just fine locally - but the servers run on UTC time - so the form will be midnight UTC (5pm pacific)

I think the fix for this is to override/set the default value in the template instead of the form. If you're in the template you have access to the current course object.

You might make your function accept the current course object (which has a timeline attribute) and call that function in the template

server/utils.py
```diff
@@ -46,11 +46,7 @@ def local_time(time, course, fmt='%a %m/%d %I:%M %p'):
     """Format a time string in a course's locale.
     Note that %-I does not perform as expected on Alpine Linux
     """
-    if not time.tzinfo:
-        # Assume UTC
-        time = pytz.utc.localize(time)
-    local = time.astimezone(pytz.timezone('America/Los_Angeles'))
```
kevinlin1: @Sumukh Not sure if this is a bug. Shouldn't we use the course.timezone?

kevinlin1:
Now using placeholder text:

screen shot 2016-08-28 at 12 56 49

server/utils.py
```diff
@@ -67,6 +63,22 @@ def server_time_obj(time, course):
     # Store using UTC on the server side.
     return time.astimezone(pytz.utc)
 
+def next_week_obj(course):
+    """Get a datetime object representing 1 week from now at 11:59:59 pm."""
+    next_week_date = dt.datetime.now(course.timezone) + dt.timedelta(weeks=1)
+    midnight_time = dt.time(hour=23, minute=59, second=59, microsecond=0,
```
Sumukh: I think there's some danger in using pytz time zones with date time. I'll look into it into when I get back on my computer

server/utils.py
kevinlin1: See also: https://github.com/Cal-CS-61A-Staff/ok/commit/f93c3c25d8ec34844c5baae6ec8ce8feec9e5232

server/utils.py
kevinlin1: I just made a new commit that's hopefully a little closer: https://github.com/Cal-CS-61A-Staff/ok/pull/825/commits/9db1c09107ea26be7ded42fa29b79b53a1892d34

Sumukh:
Sorry for the delay. I haven't been able to look at it. I'll take a look at our next Ok meeting and then merge & deploy.

knrafto:
@kevinlin1 the tests are failing, can you look into why?

kevinlin1:
@knrafto It's this change: https://github.com/Cal-CS-61A-Staff/ok/pull/825/commits/42079b2310488022958ad0e69c2e2fd71a9d4a47

I'm not sure how that method is meant to be used.

server/templates/staff/course/assignment.new.html
```diff
@@ -31,11 +31,11 @@
             <div class="box-body">
               {% call forms.render_form(form, action_url="", action_text='Create Assignment', class_='form') %}
                   {{ forms.render_field(form.display_name, label_visible=true, placeholder='Hog', type='text') }}
-                  {{ forms.render_field(form.name, label_visible=true, placeholder='cs61a/sp16/proj01', type='text') }}
+                  {{ forms.render_field(form.name, label_visible=true, placeholder='cal/cs61a/fa16/proj01', type='text') }}
                   {{ forms.render_field(form.max_group_size, label_visible=true, type='number', min='1', value='1') }}
-                  {{ forms.render_field(form.due_date, label_visible=true, type='text', class='form-control datepicker') }}
-                  {{ forms.render_field(form.lock_date, label_visible=true, type='text', class='form-control datepicker') }}
-                  {{ forms.render_field(form.url, label_visible=true, placeholder='http://cs61a.org/proj/hog (Optional)', type='text') }}
+                  {{ forms.render_field(form.due_date, label_visible=true, placeholder=utils.new_due_date(current_course), type='text', class='form-control datepicker') }}
```
Sumukh: Let's just make these the values to save users time.

server/utils.py
```diff
@@ -42,15 +42,13 @@ def decode_id(value):
 # "tzinfo argument of the standard datetime constructors 'does not work'
 # with pytz for many timezones."
 
+iso_datetime_fmt = '%Y-%m-%d %H:%M:%S'
```
Sumukh: Should be in `constants.py`

PR #834, opened 2016-09-01
Use optimized SQL Query 

### LINKED ISSUE #829: Retrieving all final submissions makes requests time out
This is pretty critical - we can't do any grading until this is resolved. The culprit is [`Assignment.course_submissions`](https://github.com/Cal-CS-61A-Staff/ok/blob/master/server/models.py#L356).

@Sumukh and I have come up with a MySQL query that runs in a reasonable time. Still TODO:
- [ ] Sufficiently abstract the query so we can use it in various places and never have to touch it again
- [ ] Create a naive fallback for sqlite (local development)
- [ ] Test by comparing query results to fallback

Sumukh:
@knrafto and I made a better query for the course submission logic. It's a _lot_ faster (two orders of magnitude for large courses). 

Resolves #829


tests/test_grading.py
```diff
@@ -43,28 +44,98 @@ def setUp(self):
                 print("User {} | Submission {} | Time {}".format(user_id, num, time))
         db.session.commit()
 
-    def test_course_submissions(self):
-        students, submissions, no_submission = self.assignment.course_submissions()
+    def _course_submissions_ids(self, assignment):
+        """ Return course submissions with objects. """
```
Sumukh: Deceptive docstring. These aren't objects.

server/models.py
```diff
@@ -353,20 +359,130 @@ def user_status(self, user):
...
+                    if not fs and not include_empty:
+                        continue
+                    data = {
+                        'user': {
+                            'id': member.id,
+                            'email': member.email,
+                        },
+                        'group': {
+                            'group_id': group_obj.id,
+                            'group_member': ','.join([str(m_id) for m_id in group_ids]),
```
Sumukh: Why is this one joined here but `group_emails` isn't?

server/models.py
```diff
@@ -353,20 +359,130 @@ def user_status(self, user):
...
 
-    def course_submissions(self):
+    def course_submissions(self, include_empty=True):
+        """ Return data on all course submissions for all enrolled users
+        List of dictionaries with user, group, backup dictionaries.
+        Sample [ {'user': {}, 'group': {}, 'backup': {} }]
+        """
+        current_db = db.engine.name
+        if current_db != 'mysql':
+            print("RUnning slow query")
```
Sumukh: Don't need this print in production. Remove

server/models.py
```diff
@@ -145,6 +146,11 @@ def export(self):
     def as_dict(self):
         return {c.name: getattr(self, c.name) for c in self.__table__.columns}
 
+    def from_dict(self, dict):
```
Sumukh: This is used in the API to take the dict from course_submission into objects.

server/controllers/api.py
```diff
@@ -606,25 +606,34 @@ def get(self, user, name):
...
+        joined = models.db.joinedload
+        base_query = (models.Backup.query.options(joined('assignment'),
+                                                  joined('submitter'),
+                                                  joined('messages'))
+                            .filter(models.Backup.id.in_(subm_keys))
+                            .order_by(models.Backup.created.desc()))
+
+        output = base_query.all()
+        num_subms = len(subm_keys)
+        # output = base_query.limit(limit).offset(offset)
```
Sumukh: If we need to paginate in prod, this will help us do that.

server/controllers/admin.py
```diff
@@ -376,14 +376,16 @@ def assignment_stats(cid, aid):
...
 
     pie_chart = pygal.Pie(half_pie=True, disable_xml_declaration=True,
                           style=CleanStyle,
                           inner_radius=.5, legend_at_bottom=True)
     pie_chart.title = 'Students submission status'
-    pie_chart.add('Students with Submissions', stats['students_submitted'])
-    pie_chart.add('Not Submitted', stats['students_nosubmit'])
+    pie_chart.add('Students with Submissions', stats['students_with_subm'])
+    pie_chart.add('Students with Backups', stats['students_with_backup'])
+    pie_chart.add('Not Started', stats['students_no_backup'])
```
Sumukh: image

server/controllers/api.py
Sumukh: It looks like the query completes pretty quickly - but sending 1409 backups results in a long page - so I'm going to paginate it. I think it'll require a few changes to the API docs/

PR #836, opened 2016-09-02
Better stats

Sumukh:
Assignment Level Stats & some assorted deployment improvements. 

image

Potential future improvements:
Be able to view scores from this view.


server/static/js/staff.js
```diff
@@ -1,6 +1,14 @@
+var listIncreasePagination = function(list, amount) {
```
Sumukh: This probably isn't ideal javascript style.

knrafto: function listIncreasePagination(list, amount) {

tests/test_grading.py
```diff
@@ -92,9 +92,9 @@ def test_course_submissions_optimized(self):
             print("Running with {}".format(db.engine.name))
             query = self.assignment.mysql_course_submissions_query()
             data = [d for d in query]
-            backups_from_query = [d[6] for d in data if d[6]]
+            backups_from_query = [d[7] for d in data if d[7]]
```
knrafto: I really really don't like the magic number 7 appearing here :( can course_submissions_query return a sequence of dicts somehow? It could be a generator so it doesn't destroy the pagination that sqlalchemy might do

Sumukh: `course_submissions_query` does return a sequence of dict - but the mysql_query fn doesn't. I'd be okay with moving the creation of the sequence of dict into the `mysql_course_submissions_query` fn. 

Ideally we'd be using namedtuple objects and then we wouldn't have to worry about this.

knrafto: Why not just use course_submissions_query then?

Sumukh: 👍 

I think I was hoping to test that the mysql query behaves exactly like the slow query in terms of what backups it obtains - but that's done elsewhere and the code above didn't really test that anyway.

PR #841, opened 2016-09-07
Redirection for New Assigments

eurieoh:
Redirect to template upload page if the new assignment is visible.


Sumukh:
LGTM. Deploying soon!

PR #848, opened 2016-09-13
Interface for background jobs

### LINKED ISSUE #820: Run some tasks in background
Some requests time out. These include:
* Autograding all final submissions
* Course statistics

We should unload some of the work into a background task. Probably using [Flask-RQ](http://flask-rq.readthedocs.io/en/latest/).

knrafto:
Resolves #820. Putting this up now for feedback.
 
To test manually:
* Start Redis
* Start the server (`./manage.py server`)
* Start a worker (`./manage.py worker`)
* Go to `/admin/course/1/jobs/test/`
* Set how long the test job will run for, and whether it will fail.
* Start the job. A new window will appear. Refresh until the output appears.

TODO:
- [x] Set up kubernetes to spin up workers
- [x] Document how to set up a job (and logging) better
- [x] Fix Travis config
- [x] Make jobs course-specific (with course-level permissions)
- [x] Page to list all jobs
- [x] Write integration tests for job interface
- [x] Sentry
- [x] Manually test a job that times out


server/models.py
```diff
@@ -1373,3 +1373,21 @@ def delete(self):
...
+
+    id = db.Column(db.Integer, primary_key=True)
+    updated = db.Column(db.DateTime(timezone=True), onupdate=db.func.now())
+    status = db.Column(db.Enum(*statuses, name='status'), nullable=False)
+
+    # The user who started the job.
+    user_id = db.Column(
+        db.Integer, db.ForeignKey('user.id')
+    )
+    user = db.relationship('User')
```
Sumukh: Would it be useful to have a row for what course it belongs to as well? That way all staff members can see what jobs for a course have been run?

knrafto: That's a good idea

Sumukh:
Haven't really looked at this PR but I'm just going to note this for future reference: 
- One thing that would be nice is to not require that redis is running locally when we are not developing job related tasks.

server/jobs.py
```diff
@@ -0,0 +1,70 @@
...
+    get_queue().enqueue_call(
+        func=func,
+        args=args,
+        kwargs=kwargs,
+        job_id=str(job.id),
+    )
+
+    return job
+
+@background_job
```
Sumukh: Is the default way to create jobs to add to this file? Would it make sense to have this be a module (with all of the helper functions/setup done in `__init__.py`) and have individual job files that can be imported? 

I know this isn't as relevant for this PR since it only has one job - so if you want to push that discussion back that's fine too.

knrafto:
That's already the case. You don't get a nasty exception until you try to start a job. You can still view past jobs though.

kubernetes/ok-staging-worker.yaml
```diff
@@ -0,0 +1,65 @@
+apiVersion: v1
+kind: ReplicationController
```
Sumukh: Ideally this would be a deployment - but I'm okay with it being an RC

kubernetes/ok-staging-worker.yaml
```diff
@@ -0,0 +1,65 @@
...
+metadata:
+  name: ok-worker
+spec:
+  replicas: 5
```
Sumukh: 5 is a bit high considering we won't have a lot of demand - but it should be fine.

server/jobs/__init__.py
```diff
@@ -0,0 +1,76 @@
...
+        get_queue().enqueue_call(
+            func=func,
+            args=args,
+            kwargs=kwargs,
+            job_id=str(job.id),
+        )
+    except redis.exceptions.ConnectionError as e:
+        job.failed = True
+        job.status = 'finished'
+        job.log = 'Could not connect to Redis: ' + str(e)
```
Sumukh: In the future we might to consider running instantly here.

server/jobs/example.py
```diff
@@ -0,0 +1,13 @@
...
+from server import jobs
+
+@jobs.background_job
+def test_job(duration=0, should_fail=False):
+    logger = jobs.get_job_logger()
+
+    logger.info('Starting...')
+    time.sleep(duration)
+    if should_fail:
+        1/0
```
Sumukh: Quantified code would not be happy :/

Issue #852, opened 2016-09-14
Sorting by role has no effect


knrafto:
To reproduce:
* Go to course enrollment page
* Scroll down to the "Staff" section
* Try to sort by role. Nothing will happen.

Issue #850, opened 2016-09-14
Way of viewing full score tag. 


Sumukh:
It'd be useful to see the full contents of a score tag.
![image](https://cloud.githubusercontent.com/assets/882381/18497290/d0cc06b4-79e1-11e6-9241-5df6561a0dd4.png)

Simple implementation that'd be fine for now: These link to a page that outputs the full details. 

Route: `/admin/grading/<backup_id>/score/<score_id>` 

Files to look at: 
`admin.py`
`templates/staff/grading/code.html`

knrafto:
Closed in #908

Issue #849, opened 2016-09-13
Require User Name Field


colinschoen:
Make the student/lab assistant/staff name field required instead of optional. 
![pasted image at 2016_09_13 01_01 pm](https://cloud.githubusercontent.com/assets/2927722/18490834/90a8fa02-79b8-11e6-83d6-316347ae8986.png)

PR #854, opened 2016-09-14
[WiP] Fix miscellaneous issues and features

kevinlin1:
- [ ] #849: require user name for enrollment (upon clicking Enroll, the page will refresh and show the required field message)
- [x] #852: fix sorting columns in admin view
- [ ] #850: view the full score tag
- [x] #863: default regex for MOSS
- [x] #864: normalize user names from roster
- [ ] #866: fix page titles in templates and tests


server/utils.py
```diff
@@ -64,28 +63,41 @@ def server_time_obj(time, course):
...
+    """ Format a human-readable time difference (e.g. "6 days ago")"""
     if date.tzinfo:
         date = date.astimezone(pytz.utc).replace(tzinfo=None)
     now = dt.datetime.utcnow()
     return humanize.naturaltime(now - date)
 
+
+def humanize_name(name):
+    """ Return a canonical representation of a name in First Last format."""
+    parts = [part.strip() for part in name.split(",")]
```
Sumukh: If `name` is None - we are going to get an exception. It's also worth testing for having the name be None

server/utils.py
```diff
@@ -64,28 +63,41 @@ def server_time_obj(time, course):
...
+    if name == name.upper():
+        split = [["-".join(subword.capitalize() for subword in word.split("-"))
+                    for word in part.split()]
+                for part in parts]
+        parts = [" ".join(part) for part in split]
+    return " ".join(parts[::-1])
```
kevinlin1: This list comprehension is a little long: I wonder if there's a better library for this.

server/utils.py
kevinlin1: Okay, thanks! Was just about to ask whether the db string format would allow for NoneType names.

server/jobs/moss.py
```diff
@@ -10,7 +10,7 @@
 from server import jobs
 
 @jobs.background_job
-def submit_to_moss(moss_id=None, file_regex="*", assignment_id=None, language=None):
+def submit_to_moss(moss_id=None, file_regex=".*", assignment_id=None, language=None):
```
kevinlin1: Is #863 as simple as this?

kevinlin1:
screenshot from 2016-09-17 22 29 54
screenshot from 2016-09-17 22 29 19

server/jobs/moss.py
Sumukh: Kind of, unfortunately it needs to be changed in three places. forms.py, templates/staff/jobs/moss.html, and here I guess.

server/utils.py
Sumukh: http://nameparser.readthedocs.io/en/latest/ looks interesting but might be overkill. Also it seems like it needs a bit of tweaking to make it work with our name formatting 
```
nameparser.HumanName('HACKER, P ALYSAA')
<HumanName : [
	title: ''
	first: 'P'
	middle: 'ALYSAA'
	last: 'HACKER'
	suffix: ''
	nickname: ''
]>
```
This one would definitely be overkill: https://github.com/datamade/probablepeople

Sumukh:
This isn't the scope of this PR: What do you think about removing the `text-transform: uppercase;`  for the name part of the header? 

I don't feel strongly about it - your call about what looks better.

Sumukh:
BTW - I addressed the student side of #850 in 896b02c1faf8413d203f4ab9410a33ee3097ad17

server/utils.py
kevinlin1: Yeah, I found that too. I tested `humanize_name` against the entire fa16 roster and they looked good to my eye. But the bCourses dataset is a little limited in that regard since there's only up two parts (1 comma separating first name from the last name) and no rare suffixes like Jr. in this semester's class.

server/jobs/moss.py
kevinlin1: grepped! 434abfd

kevinlin1:
I still need to work on improving the behavior of the required name field #849: right now, there's an awkward refresh that seems to imply that the user has been added since validation (4e7398e) is just on the server side. Will probably need to add something to the client side template so that it will also reject form submission on an empty field.

PR #856, opened 2016-09-15
Add nginx in front of gunicorn

Sumukh:
Works in staging.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,22 +1,22 @@
 FROM python:3.5-alpine
 
+RUN apk add --update patch ca-certificates nginx
+
 RUN mkdir /code/
 WORKDIR /code/
 
 ADD requirements.txt .
-RUN apk add --update patch ca-certificates nginx && rm -rf /var/cache/apk/*
-
-RUN mkdir -p /tmp/nginx/client-body
-
-COPY docker/nginx/nginx.conf /etc/nginx/nginx.conf
-COPY docker/nginx/default.conf /etc/nginx/conf.d/default.conf
-
 RUN pip3 install -r requirements.txt
 
 ADD . .
 
+RUN mv docker/nginx/nginx.conf /etc/nginx/nginx.conf
+RUN mv docker/nginx/default.conf /etc/nginx/conf.d/default.conf
+
 RUN ./manage.py assets build
 
 CMD nginx; gunicorn --bind unix:/tmp/server.sock wsgi:app --workers 3
+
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -1,15 +1,22 @@
...
 
-CMD gunicorn -b 0.0.0.0:5000 wsgi:app --workers 3
+RUN mv docker/nginx/nginx.conf /etc/nginx/nginx.conf
+RUN mv docker/nginx/default.conf /etc/nginx/conf.d/default.conf
+
+RUN ./manage.py assets build
+
+CMD nginx; gunicorn --bind unix:/tmp/server.sock wsgi:app --workers 3
+
+RUN rm -rf /var/cache/apk/*
```
Sumukh: Low impact on our docker image sizes

image

docker/nginx/default.conf
```diff
@@ -0,0 +1,33 @@
...
+        root /code/server/;
+    }
+
+    location / {
+      proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+      proxy_set_header X-Nginx-IP $remote_addr;
+
+      # enable this if and only if you use HTTPS
+
+      # proxy_set_header X-Forwarded-Proto https;
```
Sumukh: TBD about how we enforce SSL.

docker/nginx/nginx.conf
```diff
@@ -0,0 +1,66 @@
...
+    sendfile        off;
+
+    keepalive_timeout  65;
+
+    gzip  on;
+
+    # Compression level (1-9).
+    # 5 is a perfect compromise between size and cpu usage, offering about
+    # 75% reduction for most ascii files (almost identical to level 9).
+    gzip_comp_level    5;
```
Sumukh: 1 is sufficient - but 5 leads to a 30% improvement on the assignment stats page.

server/templates/staff/sidebar.html
```diff
@@ -61,11 +61,10 @@
         <li><a href="https://console.cloud.google.com/home/dashboard?project=ok-server">
             <i class="fa fa-server"></i> <span>GCE Console</span></a>
         </li>
-        <li><a href="https://console.cloud.google.com/home/dashboard?project=ok-server">
-            <i class="fa fa-database"></i> <span>Logs</span></a>
+        <li><a href="/rq">
```
Sumukh: Not sure if there's a better way to do this.

PR #857, opened 2016-09-15
Use flask for GZIP and revert nginx addition.

Sumukh:
A weird nginx error that occurs on oauth login is making we not want to deal with nginx.


Dockerfile
```diff
@@ -1,22 +1,15 @@
...
 ADD . .
 
-RUN mv docker/nginx/nginx.conf /etc/nginx/nginx.conf
-RUN mv docker/nginx/default.conf /etc/nginx/conf.d/default.conf
-
-RUN ./manage.py assets build
-
-CMD nginx; gunicorn --bind unix:/tmp/server.sock wsgi:app --workers 3
-
-RUN rm -rf /var/cache/apk/*
```
Sumukh: This change would be good to keep.

manage.py
```diff
@@ -12,7 +11,7 @@
 
 from server import create_app, generate
 from server.models import db, User, Course, Version
-from server.extensions import assets_env, cache
```
Sumukh: We should also keep this command (and later migrate this to use flask/click)

server/settings/dev.py
```diff
@@ -13,6 +13,7 @@
 SQLALCHEMY_TRACK_MODIFICATIONS = False
 SQLALCHEMY_DATABASE_URI = 'sqlite:///../oksqlite.db'
 # SQLALCHEMY_ECHO = True
+COMPRESS_MIMETYPES = ['text/html', 'text/css', 'text/csv', 'text/xml', 'application/json', 'application/javascript']
```
Sumukh: Compressing text/csv is not faster than just sending results directly (have to run level 6 gzip first - which makes it slower)

server/templates/staff/sidebar.html
```diff
@@ -61,10 +61,11 @@
         <li><a href="https://console.cloud.google.com/home/dashboard?project=ok-server">
             <i class="fa fa-server"></i> <span>GCE Console</span></a>
         </li>
-        <li><a href="/rq">
-            <i class="fa fa-database"></i> <span>Redis Queue</span></a>
+        <li><a href="https://console.cloud.google.com/home/dashboard?project=ok-server">
```
Sumukh: This change also seems good to keep

server/settings/prod.py
```diff
@@ -34,6 +34,8 @@
 SQLALCHEMY_TRACK_MODIFICATIONS = False
 SENTRY_USER_ATTRS = ['email', 'name']
 
+COMPRESS_MIMETYPES = ['text/html', 'text/css', 'text/csv', 'text/xml', 'application/json', 'application/javascript']
```
Sumukh: See above

Sumukh:
Closing because nginx now works.

PR #858, opened 2016-09-15
Multiple Google User Info Attempts

Sumukh:
~~This still needs to be refactored to be a bit cleaner~~

This allows for a fallback for cases when our first attempt to talk to Google for userinfo fails.  Google seems to recommend the Google Plus endpoint over the oauth2 endpoint. 

TODO
- [x] The oauth2/v3/userinfo endpoint not responding in 30 seconds.


Sumukh:
^ I guess adding a timeout depends on whether the oauth library supports adding a timeout. I haven't looked into it - but if it doesn't we can always just do the request ourselves and just use `requests.get(..., timeout=15)`

Sumukh:
Tested in staging - seems to work.

PR #861, opened 2016-09-16
Ok -> MOSS Submission

Sumukh:
ok-moss-demo


server/jobs/moss.py
```diff
@@ -0,0 +1,108 @@
...
+    backup_query = (Backup.query.options(db.joinedload('messages'))
+                          .filter(Backup.id.in_(subm_keys))
+                          .order_by(Backup.created.desc())
+                          .all())
+
+    logger.info("Retreived {} final submissions".format(len(subm_keys)))
+
+    # TODO: Customize the location of the tmp writing (especially useful during dev)
+
+    tmp_dir = '/tmp/moss-{}-{}'.format(assignment_id, random.randint(100, 999))
```
knrafto: with tempfile.TemporaryDirectory: so that the files don't stick around

server/jobs/moss.py
```diff
@@ -0,0 +1,108 @@
...
+    command = ("perl moss.pl -l {lang} {templates} -d {folder}"
+               .format(lang=language, templates=templates, folder=' '.join(files)))
+
+    logger.info("Running {}".format(command[:50] + ' ...'))
+
+    try:
+        process = subprocess.check_output(shlex.split(command))
+        logger.info(process.decode("utf-8"))
+    except subprocess.CalledProcessError as e:
+        logger.warning("There was an error running your process")
```
knrafto: We should log the stdout and stderr of the process here

knrafto:
It looks like we can't use a Text column to store the log (see Sentry). What can we do instead? A blob?

Sumukh:
See #862. I was suggesting MEDIUMTEXT




On Fri, Sep 16, 2016 at 11:06 AM -0700, "Kyle Raftogianis" <notifications@github.com> wrote:












It looks like we can't use a Text column to store the log (see Sentry). What can we do instead? A blob?

knrafto:
Ah sorry. Yeah I think that'll work better me than a blob

server/jobs/moss.py
Sumukh: Yeah. I wasn't sure how to get it - but it turns out the exception contains that info (`e.output`) 

To include stderr, we'll have to add a keyword arg to `check_output` `stderr=subprocess.STDOUT)`

Sumukh:
Reviewed your comments. I'll make these changes later today (and include migration for #862)

server/jobs/moss.py
```diff
@@ -0,0 +1,123 @@
...
+        all_student_files = glob.glob("*/*")
+        chosen_files, available_file_names = [], []
+        match_pattern = re.compile(file_regex)
+
+        logger.info("Running with regex match {}".format(file_regex))
+
+        for file in all_student_files:
+            fname = ntpath.basename(file)
+            available_file_names.append(fname)
+            if match_pattern.match(fname):
```
knrafto: This match should be done while copying the backup files over, so that we don't need to copy files we aren't going to use. I think it'll make the logic easier to follow too (no need to differentiate between `all_student_files` and `chosen_files`)

server/jobs/moss.py
```diff
@@ -0,0 +1,123 @@
...
+    if not assign:
+        logger.info("Could not find assignment")
+        return
+
+    subms = assign.course_submissions(include_empty=False)
+    subm_keys = set(s['backup']['id'] for s in subms)
+
+    seen = set()
+    for subm in subms:
+        if subm['backup']['id'] in seen:
```
knrafto: Should the for loop be over `subm_keys`? then you don't need this check.

server/jobs/moss.py
```diff
@@ -0,0 +1,123 @@
...
+
+    logger.info("Retreived {} final submissions".format(len(subm_keys)))
+
+    # TODO: Customize the location of the tmp writing (especially useful during dev)
+
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        # Copy in the moss script
+        with open('server/jobs/moss-submission.pl', 'r') as f:
+            moss_script = f.read()
+            moss_script = moss_script.replace('YOUR_USER_ID_HERE', str(moss_id))
```
knrafto: Nit: you don't need the `.pl` file open for the next three lines, so you can dedent them

server/jobs/moss.py
```diff
@@ -0,0 +1,123 @@
...
+                f.write(assign.files[template])
+
+            template_files.append(template)
+
+        logger.info("Using template files: {}".format(' '.join(template_files)))
+
+        templates = ' '.join(["-b {file}".format(file=f) for f in template_files])
+
+        os.chdir(tmp_dir)
+        all_student_files = glob.glob("*/*")
```
knrafto: Might want to add a comment why this does not match template files (it confused me for a bit). Or to be more explicit, move this line to right after you finish writing student files

server/jobs/moss.py
Sumukh: I don't think it can be - there's no easy way to go from something in `subm_key` to it's corresponding group/backup dictionary

A refactor would be to have a dictionary to goes `{subm_id: data for subm_id in subm_keys}` - but we'd kind of implicitly overriding some keys - I'd rather be explicit about what is happening.

knrafto: Hmm. How about renaming `seen` to `subm_keys`?

Sumukh: Good idea. 👍

server/jobs/moss.py
Sumukh: Ok - I can move it earlier - but I'm still planning on keeping a list of names that we considered but rejected for debugging purposes.

Sumukh:
Migration has been run.

PR #872, opened 2016-09-22
Publish Grades

raychan3:
Open PR to track changes and comments.


colinschoen:
Maybe a test for the authorization check added in `admin.py`.

server/models.py
```diff
@@ -607,6 +616,13 @@ def _unflag_all(self, member_ids):
         if backup:
             backup.flagged = False
 
+    @transaction
+    def publish_grades(self, tag):
+        """Publish student grades for the assignment."""
+        self.grade_visibility = tag
+        db.session.add(self)
```
Sumukh: I don't think this method wants to do an add - it is not inserting an row here. You also don't need `db.session.commit()` because the decorator will do it for you.

server/models.py
```diff
@@ -285,6 +285,13 @@ class Assignment(Model):
     files = db.Column(JsonBlob)  # JSON object mapping filenames to contents
     course = db.relationship("Course", backref="assignments")
 
+    publish_tags = [tag.title() if tag != 'private' else 'Hidden' for tag in GRADE_TAGS]
+    grade_visibility = db.Column(db.String(255), nullable=False, default="Hidden")
```
Sumukh: How does this handle multiple tags? Will the default value pass the check constraint?

I think this should be nullable (for ease of migration) - but also because it makes sense to have no tags visible. 

An example of how we support a list of elements is available in the OAuth models (see the `Client` class)

knrafto: If we use `StringList` here, it shouldn't be nullable (the we have to deal with both null and the empty list). If we set the default value to the empty list the migration should go smoothly:

`db.Column(StringList, nullable=False, default=[])`

knrafto: Small bikeshed to rename this to `published_score_kinds` or similar

raychan3:
Student side hasn't been implemented yet. Pushed for comments on admin side.

raychan3:
Added student view changes.

TODO: Add tests.
TODO: also need to fix some of the current tests

raychan3:
Sorry it took so long to add the test case; been busy lately. @knrafto @Sumukh could you guys take another look? Thanks

server/models.py
```diff
@@ -283,13 +283,16 @@ class Assignment(Model):
     autograding_key = db.Column(db.String(255))
     uploads_enabled = db.Column(db.Boolean(), nullable=False, default=False)
     upload_info = db.Column(db.Text)
+    published_scores = db.Column(StringList, nullable=False, default=[])
```
Sumukh: Why does this have to be not nullable? Is it because StringList will complain  when the value from SQL is null? 

If so we might want to consider changing StringList to handle None as input?

server/models.py
```diff
@@ -306,6 +309,8 @@ def can(cls, obj, user, action):
         is_staff = user.is_enrolled(obj.course.id, STAFF_ROLES)
         if action == "view":
             return is_staff or obj.visible
+        if action == "publish":
+            return is_staff and obj.visible
```
Sumukh: Does it make sense to allow staff to be able to pre publish scores (even if it's marked as not visible to students)?

server/models.py
```diff
@@ -600,6 +605,18 @@ def _unflag_all(self, member_ids):
         if backup:
             backup.flagged = False
 
+    @transaction
+    def publish_grades(self, tag, hide=False):
+        """Publish student grades for the assignment."""
```
Sumukh: Seems more like a toggle to me. 

I don't have a strong opinion on this - but it's worth thinking if this might be better as two methods?

server/forms.py
```diff
@@ -327,6 +327,12 @@ class CourseUpdateForm(BaseForm):
     active = BooleanField('Activate Course', default=True)
     timezone = SelectField('Course Timezone', choices=[(t, t) for t in pytz.common_timezones])
 
+class GradeVisibilityUpdateForm(BaseForm):
+    publish_tags = [tag.title() for tag in GRADE_TAGS if tag != 'private']
+    grades = SelectField('Publish Grade', choices=[(t, t) for t in publish_tags], default='Total')
```
Sumukh: `grades` seemed somewhat confusing later on when it gets used. `tag` could be better. 

A better description might be along the lines
`Publish Scores With Tag:`

server/models.py
```diff
@@ -600,6 +605,18 @@ def _unflag_all(self, member_ids):
...
+        published = tag in self.published_scores
+        if hide and published:
+            self.published_scores = [t for t in self.published_scores if t != tag]
+        elif not hide and not published:
+            # No current use case for correctness tag
+            if tag == 'Revision' and 'Composition' not in self.published_scores:
+                self.published_scores = self.published_scores + [tag, 'Composition']
```
Sumukh: Is the purpose of this to automatically add `Composition` if revision is selected? 
It might be surprising to a user if they ever toggle revision scores. The flash will also say something like 
> Toggled Revision Scores 

but it will also have toggled composition scores. 

I think I understand the use case that this if statement is going for - but I don't think it's worth the extra communication cost to be clear to the user about what's happening. 

Maybe something that would accomplish this is simply flashing (in the route) a warning about how composition score is not visible if the user is toggling Revision but composition is not shown.

server/models.py
```diff
@@ -782,8 +799,11 @@ def visible_grades(self):
         """ Return public grades. "Autograder" kind are errors from the
         autograder and should not be shown.
         """
+        published_scores = self.assignment.published_scores
+
         return [s for s in self.scores
-            if s.public and s.kind not in HIDDEN_GRADE_TAGS]
+            if s.public and s.kind not in HIDDEN_GRADE_TAGS and s.kind.title() in published_scores]
```
Sumukh: This makes we wonder why `published_scores` are stored in title case. It's worth keeping it in lower case for consistency. 

It would also help make this list comprehension a bit shorted if it weren't in title case.

server/templates/staff/course/assignment/assignment.html
```diff
@@ -191,13 +191,15 @@ <h3 class="box-title">Actions</h3>
...
                   </a></li>
-
                   <li> <a href="{{ url_for('.assign_grading', cid=current_course.id, aid=assignment.id) }}" type="button">
                         <i class="fa fa-user-plus"></i> Assign Grading
                   </a></li>
                   <li> <a href="{{ url_for('.assignment_queues', cid=current_course.id, aid=assignment.id) }}" type="button">
                         <i class="fa fa-inbox"></i> Grading Queues
                   </a></li>
+                  <li> <a href="{{ url_for('.publish_grades', cid=current_course.id, aid=assignment.id) }}" type="button">
+                        <i class="fa fa-wpforms"></i> Publish Grades
```
Sumukh: I think we use the term "Scores" instead of Grades - so we should be consistent. (I might be wrong though)

server/templates/staff/course/assignment/assignment_publish.html
```diff
@@ -0,0 +1,48 @@
+{% extends "staff/base.html" %}
+{% import 'staff/_formhelpers.html' as forms %}
+
+{% block title %} Publish Grades | OK {% endblock %}
```
Sumukh: Same thing about scores

tests/test_grading.py
```diff
@@ -180,3 +183,75 @@ def test_scores_with_generate(self, generate=False):
             backup_creators.extend(s.backup.owners())
 
         self.assertEquals(len(backup_creators), len(csv_rows) - 1)
+
+    def test_publish_grades(self):
+        points = [20.0, 2.0, 15.0, 1.0, 17.0, 0.0, 20.0, 1.0]
```
Sumukh: Just curious: Any particular reason for these values?

tests/test_grading.py
```diff
@@ -180,3 +183,75 @@ def test_scores_with_generate(self, generate=False):
...
+        scores, i, users = {}, 0, [self.user1, self.user3]
+        for score_kind in ['Total', 'Composition']:
+            for user in users:
+                for assign in self.active_assignments:
+                    backup = assign.final_submission(assign.active_user_ids(user.id))
+                    scores = Score(backup_id=backup.id, kind=score_kind, score=points[i],
```
Sumukh: It doesn't seem like `i` is used at all (Only set for 0). Should it be? 

Can everything just have the same score?

tests/test_grading.py
```diff
@@ -180,3 +183,75 @@ def test_scores_with_generate(self, generate=False):
...
+                                   message="Good work", assignment_id=assign.id,
+                                   grader=self.staff1)
```
Sumukh: This might insert multiple scores with the same tag (but have both not be archived). One quick fix is to ensure that the backup doesn't already have a score with that kind.

tests/test_grading.py
```diff
@@ -180,3 +183,75 @@ def test_scores_with_generate(self, generate=False):
...
+
+        self.login(self.staff1.email)        
+        response = self.client.post(endpoint, data={})
+        self.assert_200(response)
+        for user in users:
+            check_visible_scores(user, self.assignment, hidden=['Composition'], visible=['Total'])
+            check_visible_scores(user, self.assignment2, hidden=['Total', 'Composition'])
+
+        self.login('okadmin@okpy.org')
+        endpoint ='/admin/course/{}/assignments/{}/publish'.format(self.course.id, self.assignment2.id)
```
Sumukh: Some comments about what these tests are doing would be nice. 

> Toggling and check for visibility from the student side.

tests/test_grading.py
```diff
@@ -180,3 +183,75 @@ def test_scores_with_generate(self, generate=False):
...
+            check_visible_scores(user, self.assignment2, hidden=['Total'], visible=['Composition'])
+
+        self.login(self.staff1.email)
+        endpoint = '/admin/course/{}/assignments/{}/publish'.format(self.course.id, self.assignment2.id)
+        response = self.client.post(endpoint, data={})
+        self.assert_200(response)
+        for user in users:
+            check_visible_scores(user, self.assignment, hidden=['Total', 'Composition'])
+            check_visible_scores(user, self.assignment2, visible=['Composition', 'Total'])
+
```
Sumukh: Nice to have but not required for this PR: A test to make sure the Admin grading view still has a way to view these hidden

server/models.py
Sumukh: I'm also open to keeping this as non nullable - not sure what the best thing to do is. I kind of like the idea of having it be null which tells us that the staff haven't touched anything. 

@knrafto thoughts?

Sumukh:
Great! This is really close - I left some stylistic comments and had a few questions and suggestions.

tests/test_grading.py
raychan3: They were test values that I thought I was going to use. I'l randomize it.

tests/test_grading.py
raychan3: It was suppose to be used with the points list. Forgot to delete it.

server/models.py
raychan3: I used title so that i showed up in the form UI dropdown as "Total" rather than "total". Not sure how picky we are on that.

Sumukh: In that case, we should have the form do `.title()` when generated the option choices instead of having the DB store value in title form - I believe there are examples of this in `forms.py`

tests/test_grading.py
raychan3: Wouldn't we need to pass in the user to visible_grades? Not sure how to do that from the front end as visible_grades is a hybrid property and only used in the html.

Sumukh: It's not on the student side - It would just for the Admin side grading view (which doesn't use visible_grade right now) 

That test isn't actually specific to this PR - so we'll just add that test later.

server/forms.py
```diff
@@ -327,6 +327,12 @@ class CourseUpdateForm(BaseForm):
     active = BooleanField('Activate Course', default=True)
     timezone = SelectField('Course Timezone', choices=[(t, t) for t in pytz.common_timezones])
 
+class PublishScoresWithTags(BaseForm):
+    publish_tags = [tag.title() for tag in GRADE_TAGS if tag != 'private']
+    grades = SelectField('Publish Grade', choices=[(t, t) for t in publish_tags], default='Total')
```
Sumukh: Instead of `tag.title` in the list comprehension above - you want it in the tuple below. 

See line 233 (https://github.com/Cal-CS-61A-Staff/ok/pull/872/files#diff-de3e8007a66bcd69728c8ddecc1f6f32L223)

server/models.py
Sumukh: Nevermind - I saw your earlier comment.

Sumukh:
Is this still a WiP? 

@knrafto: Do you want to review? If not, I can dismiss your request for changes.

Sumukh:
https://www.quantifiedcode.com/app/project/e6ae811f0c244655b5026e0883ffb8f9

server/models.py
```diff
@@ -266,8 +265,8 @@ class Assignment(Model):
         lock_date - DEADLINE+1 (Hard Deadline for submissions)
         url - cs61a.org/proj/hog/hog.zip
         flagged - User has indicated this one should be graded and not others
+        grade_visibility - grades are either hidden from or certain grades are published to students
```
knrafto: I don't think this comment is accurate?

server/controllers/admin.py
```diff
@@ -434,6 +434,45 @@ def templates(cid, aid):
...
+                flash("{visibility} scores for {assignment} already published".format(
+                    visibility=tag.title(), assignment=assign.display_name), "success")
+            elif tag == 'revision' and 'composition' not in assign.published_scores:
+                flash("{visibility} scores for {assignment} cannot be published with hidden composition scores".format(
+                    visibility=tag.title(), assignment=assign.display_name), "warning")
+            else:
+                assign.publish_score(tag)
+                flash("Published {assignment} {visibility} scores".format(
+                    assignment=assign.display_name, visibility=tag), "success")
+
```
knrafto: This should redirect to `url_for('.publish_scores', ...)`, so that if you refresh the page after you submit you won't resubmit the form. https://en.wikipedia.org/wiki/Post/Redirect/Get

server/controllers/admin.py
```diff
@@ -434,6 +434,45 @@ def templates(cid, aid):
...
+        tag = form.grades.data
+        hide = form.hide.data
+        if hide:
+            if tag not in assign.published_scores:
+                flash("{visibility} scores for {assignment} already hidden".format(
+                    visibility=tag.title(), assignment=assign.display_name), "success")
+            else:
+                assign.hide_score(tag)
+                flash("Hid {assignment} {visibility} scores".format(
+                    assignment=assign.display_name, visibility=tag), "success")
```
knrafto: `tag.title()`

raychan3:
With the newly added `tag.title()`, the flash messages for successfully publish and hid scores become "Published/Hid Hog Total scores" rather than "Published/Hid Hog total scores"

knrafto:
I thought it would be more consistent with what's shown in the drop down, but really it's not a big deal

tests/test_grading.py
```diff
@@ -180,3 +184,120 @@ def test_scores_with_generate(self, generate=False):
...
+        self.assertTrue("{} scores for {} already hidden".format('Total', self.assignment.display_name) in source)
+        for user in users:
+            check_visible_scores(user, self.assignment, hidden=['Total', 'Composition'])
+            check_visible_scores(user, self.assignment2, visible=['Composition', 'Total'])
+
+        # If assignment is not visible, cannot publish
+        self.assignment.visible = False
+        self.login(self.staff1.email)
+        endpoint = '/admin/course/{}/assignments/{}/publish'.format(self.course.id, self.assignment.id)
+        response = self.client.post(endpoint, data={})
```
knrafto: For future reference: follow_redirects will follow redirects if you want to test the final page instead

server/models.py
```diff
@@ -306,6 +308,8 @@ def can(cls, obj, user, action):
         is_staff = user.is_enrolled(obj.course.id, STAFF_ROLES)
         if action == "view":
             return is_staff or obj.visible
+        if action == "publish":
```
Sumukh: Nit: `publish` is kind of confusing since there is also a publish action on assignments. Maybe we could add a comment or change the name to publish-scores 

Also: Not being able to publish scores for a hidden assignment seems a bit off? (For example - a unpublished assignment might want to default to having scores visible) 

Not a big deal - just wondering.

Sumukh:
Thanks! This will be very useful. 

Reminder to run the schema change before merging.

knrafto:
I ran into some issues generating the migration, so I'm going to rebase this branch off the current master

knrafto:
One more thing - is there a way to view which scores have been published?

migrations/versions/7f15415c7216_publish_scores_for_assignments.py
```diff
@@ -0,0 +1,33 @@
...
+import sqlalchemy as sa
+import server
+from sqlalchemy.dialects import mysql
+
+def upgrade():
+    ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('assignment', sa.Column('published_scores', server.models.StringList(), nullable=False))
+    op.alter_column('enrollment', 'role',
+               existing_type=mysql.ENUM('student', 'lab assistant', 'grader', 'staff', 'instructor'),
+               nullable=False)
```
knrafto: was making this column non-nullable part of this PR?

Sumukh: I discussed it here - but it seems like there was some earlier discussion and I think you had a comment on it. 

https://github.com/Cal-CS-61A-Staff/ok/pull/872/files#r83545928

knrafto: I'm referring to `enrollment.role` (not `assignment.published_scores`)

PR #875, opened 2016-09-22
Require name field when enrolling participant

kevinlin1:
#849 

- [x] Prevent submission in the template instead of allowing form submission


Sumukh:
Note that this doesn't require names when doing a CSV upload - but that's fine for now since CSV uploads usually include the name if available.

Issue #864, opened 2016-09-18
Normalize user names from rosters


Sumukh:
We should do a better job in presenting the names of roster. 
Right now we display that name directly - for example: 
![image](https://cloud.githubusercontent.com/assets/882381/18612180/36d07620-7d07-11e6-8331-8d34f1eb5979.png)

A user's name from the roster might be HACKER, P ALLYSA 

We should detect the comma and normalize the above name to: 
Allysa P Hacker

Function to change:
1. server/models.py: User.identifier 

Test Suite: 
This PR should also include a few tests to ensure the function behaves as expected for a variety of names.

PR #876, opened 2016-09-22
Normalize names

kevinlin1:
#864

### DIFF
--- a/server/utils.py
+++ b/server/utils.py
@@ -28,7 +28,6 @@ logger = logging.getLogger(__name__)
 # link with an ID in it.
 hashids = Hashids(min_length=6)
 
-
 def encode_id(id_number):
     return hashids.encode(id_number)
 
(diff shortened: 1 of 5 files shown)


tests/test_utils.py
```diff
@@ -53,3 +53,34 @@ def test_generate_number_table(self):
         self.assertEquals(results[1], "SELECT 1 as pos")
         self.assertEquals(results[2], "SELECT 1 as pos UNION SELECT 2 as pos")
         self.assertEquals(results[3], "SELECT 1 as pos UNION SELECT 2 as pos UNION SELECT 3 as pos")
+
+    def test_humanize_name(self):
+        test_corpus = (
```
Sumukh: Nice tests! 👍

server/utils.py
```diff
@@ -64,28 +63,40 @@ def server_time_obj(time, course):
...
     if date.tzinfo:
         date = date.astimezone(pytz.utc).replace(tzinfo=None)
     now = dt.datetime.utcnow()
     return humanize.naturaltime(now - date)
 
+
+def humanize_name(name):
+    """ Return a canonical representation of a name in First Last format."""
+    if not isinstance(name, str):
+        return name
```
Sumukh: I think the only case we'll hit this is if `name` is None. If that wasn't the case - I would have preferred just returning `None` or the empty string. 

For now - this should be fine.

PR #879, opened 2016-09-23
[WiP] Search Github for Posted Solutions

Sumukh:
This searches github for files similar to the template that have been posted recently (in the last X weeks). 

This is purposefully slow/delayed so as to avoid the Github Rate Limiter/Abuse Monitor (which is more aggressive for search queries) 

A future enhancement is to set an issue template - to automatically file an issue (if one hasn't been filed yet)


Sumukh:
Thoughts on excluding jobs from test coverage? It's pretty hard to test these without mocks. 

My thinking is if there was a distinction between jobs that only tested the DB and jobs that hit some other service - I would want to test the former. I don't really think that's the for the external services unless it's critical & there are mocks. Right now since we don't have that distinction - I'm tempted to exclude the server/jobs folder from tests.

server/templates/staff/jobs/github_search.html
```diff
@@ -0,0 +1,47 @@
...
+    <div class="col-md-12">
+      <div class="box">
+        <div class="box-body">
+          {% call forms.render_form(form, action_text='Search Github') %}
+            {{ forms.render_field(form.access_token) }}
+            {{ forms.render_field(form.template_name, placeholder="hog.py") }}
+            {{ forms.render_field(form.weeks_past, default=12) }}
+            {{ forms.render_field(form.keyword, default="def ") }}
+            {{ forms.render_field(form.language) }}
+          {% endcall %}
```
Sumukh: It seems like all of these jobs are just the same thing - render the form. Is there an easy way to just render the form with one call? If so we could generalize this template.

server/forms.py
```diff
@@ -335,9 +336,19 @@ class TestJobForm(BaseForm):
...
-    language = SelectField('Language', choices=[(pl, pl) for pl in valid_languages])
+    language = SelectField('Language', choices=[(pl, pl) for pl in COMMON_LANGUAGES])
+
+class GithubSearchRecentForm(BaseForm):
+    access_token = StringField('Github Access Token',
+                               description="Get a token at https://github.com/settings/tokens",
+                               validators=[validators.required()])
+    template_name = StringField('Template File Name',
+                                validators=[validators.required()])
+    keyword = StringField('Search for lines starting with', default="def ",
```
knrafto: I don't think it's very clear what this field is for. Does it extracts lines from the template before searching? What about leading white space (so we could e.g. match Java method names)

server/forms.py
```diff
@@ -335,9 +336,19 @@ class TestJobForm(BaseForm):
...
+                          validators=[validators.required()])
+    weeks_past = IntegerField('Weeks since course started', default=12)
```
knrafto: What is this field for? And why is 12 a good default value for it?

server/jobs/github_search.py
```diff
@@ -0,0 +1,196 @@
...
+import arrow
+
+from server.models import Assignment, Backup, db
+from server.utils import encode_id
+from server.extensions import cache
+from server import jobs
+
+def get_file_from_github(html_url, logger):
+    """ Github HTML url to raw url with github """
+    # return html_url.replace('http://github.com/', 'https://raw.githubusercontent.com/')
```
knrafto: Remove commented line

server/jobs/github_search.py
```diff
@@ -0,0 +1,196 @@
...
+    html_url = html_url.replace('/blob', '')
+    return html_url.replace('https://github.com/', 'http://rawgit.com/')
+
+    try:
```
knrafto: Unreachable code?

server/jobs/github_search.py
```diff
@@ -0,0 +1,196 @@
...
+def make_github_req(url, access_token):
+    r = requests.get(url, headers={"Authorization": "Bearer {}".format(access_token)})
+    r.raise_for_status()
+    return r.json()
+
+def search_request(search_line, language, logger, access_token, page=1):
+    """ Query https://api.github.com/search/code """
+    search_query = ('"{line}" in:file language:{lang}'
+                    .format(line=search_line, lang=language))
+    if page > 6:
```
knrafto: This logic should be in get_all_results instead

server/jobs/github_search.py
```diff
@@ -0,0 +1,196 @@
...
+        results = search_request(search_line, language, logger, access_token,
+                                 page=page_num)
+        if not results or 'items' not in results:
+            logging.warning("Request {} failed {}".format(page_num, results))
+            return data_set
+        else:
+            data_set.extend(results['items'])
+    return data_set
+
+def download_repos(repos):
```
knrafto: I know this is for future use, but this should probably take a directory path and the caller can make a tmpdir if they want (and they might not want to)

server/jobs/github_search.py
```diff
@@ -0,0 +1,196 @@
...
+    if template_name not in possible_file_names:
+        logger.info("{} is not in {}".format(template_name, possible_file_names))
+
+    source_file = assign.files[template_name]
+    repos = get_online_repos(source_file, logger, language, access_token,
+                             keyword=keyword)
+    if not repos:
+        logger.warning("No repos found. Try a different keyword?")
+        return
+    recent_repos = list_recent_repos(repos, logger, access_token, weeks_past)
```
knrafto: This return value gets thrown away? The function also doesn't seem to have side effects so I think line does nothing

server/templates/staff/jobs/github_search.html
knrafto: That's a good idea. I noticed all of the job routes in admin.py all do the same thing too.

server/jobs/github_search.py
Sumukh: Good point. I left it in because I plan on using it for the future improvement (filing issues). There are "side-effects" in that things get logged (what I care about for now)

Sumukh:
Thanks for the review. I'll get back to this in a few days and add issue filing as well.

PR #881, opened 2016-09-28
Show logs while running for long running jobs

### LINKED ISSUE #880: Output status of job as its running
For long running jobs - it'd be nice to have some idea of what it's doing.

The reason we can't right now is because the text is only set when the job is complete. 

Maybe we could provide a callback that updates the log and commits to the functions (and then display that? The job would then decide when to commit. 

It might be too hard to periodically do updates.

Sumukh:
Resolves #880

### DIFF
--- a/server/jobs/__init__.py
+++ b/server/jobs/__init__.py
@@ -28,6 +28,12 @@ def background_job(f):
         logger.setLevel(logging.INFO)
         logger.addHandler(logging.StreamHandler(stream))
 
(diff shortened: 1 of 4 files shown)


knrafto:
I don't like this approach - you have to remember to take in the `show_log` function as an argument and call it when you need to.

Maybe we can spawn a background thread for each job that periodically updates the log, and terminates when the job is complete.

PR #886, opened 2016-09-28
Backup Diffs & Time Graphs

stanfieldchu:
WIP


server/controllers/admin.py
```diff
@@ -851,6 +851,74 @@ def student_assignment_detail(cid, email, aid):
...
+    scores = [backup.scores for backup in backups]
+    print(scores)
+
+    files_list = []
+    for i, backup in enumerate(backups):
+        prev = backup.files()
+        if i:
+            prev = backups[i - 1].files()
+        curr = backup.files()
+        files = highlight.diff_files(prev, curr, "full") # short for now
```
Sumukh: Should be `short`.

server/controllers/admin.py
```diff
@@ -851,6 +851,74 @@ def student_assignment_detail(cid, email, aid):
...
+
+    backups = (Backup.query.options(db.joinedload('scores'),
+                                    db.joinedload('submitter'))
+                     .filter(Backup.submitter_id.in_(user_ids),
+                             Backup.assignment_id == assign.id)
+                     .order_by(Backup.flagged.desc(), Backup.submit.desc(),
+                               Backup.created.desc())).all()
+
+    scores = [backup.scores for backup in backups]
+    print(scores)
```
Sumukh: Remove print

server/controllers/admin.py
```diff
@@ -879,11 +879,17 @@ def student_commit_overview(cid, email, aid, commit_id):
...
     # only keep "near" backups
+    found = False
     for i, backup in enumerate(backups):
         backup_commit_id = backup.hashid
         if backup_commit_id == commit_id:
+            found = True
             break
 
+    if not found:
+        flash('Cannot access commit_id: {0}'.format(commit_id), 'error')
```
Sumukh: This won't flash until later AFAIK. 404 pages don't display flashed messages.

knrafto:
When will this be ready for review?

server/templates/staff/student/assignment.overview.html
```diff
@@ -0,0 +1,320 @@
...
+            </div>
+        </div>
+    </div>
+{% endmacro %}
+
+
+{% block page_js %}
+<script type="text/javascript">
+    // show only diff 1
+    $('.timeline').hide()
```
Sumukh: Could we have this all be hidden by default and then show this so we don't see a ton of content on the screen and a flash where it disappears

knrafto:
Thanks for the PR! @Sumukh and I took a look and here’s our suggestions:

## Everywhere

* Rename “commit ID” to “backup ID” everywhere. The names in the source could should be `backup_id` instead of `commit_id` too. Ok has no concept of “commits” and it will be confusing to future collaborators. 

## Diff page

* Diff information is a little hard to read. Check out the "submitter information"
box in the grading view. In particular, don't center, and bold the data messages
* Link to the backup grading page wherever possible (whenever you print a backup ID)
* The breadcrumb (Home > cal/cs61a/fa16 > Assignments > Hog) should link back the the student’s assignment. It should be similar to the breadcrumb for `/admin/course/4/<student>/49/`
* Lots of junk shows while the page is loading and is hidden when the JS loads, which could be several seconds later. Instead hide the junk by default (e.g. with `class=”hidden”`), then show the relevant data with JS when it loads.
* The score box is unnecessary most of the time - we could just get rid of it.
* Instead of a “toggle timeline” and rendering the timeline, a link to existing timeline page would be better. That was we can integrate the "view diff" button with the existing timeline as well. To support the “current backup” in that view, we can pass in a URL param (`.../timeline?backup=sV3DtG`)

## Graph page

* Make graphs a little smaller/responsive
* Nice to have: Make the box that shows up when you hover on a dot clickable?
* Remove the “scores” box
* Put the current partner’s graph first
* If there are say 4 members in the group, we should show 4 graphs. Right now all the other partners will get lumped together in one graph.
* The links in the “submitter information” box should link back to the student’s assignment
* The breadcrumb (Home > cal/cs61a/fa16 > Assignments > Hog) should link back the the student’s assignment. It should be similar to the breadcrumb for `/admin/course/4/<student>/49/`

## Performance

The pages are pretty slow to load. Here’s some things that may or may not help:
* Backup Query: do joined load with messages (and don't join with scores)
* Change `backup.analytics()` to use a list comprehension over `backup.messages` instead of a new query - when there's a joined load it'll save queries and should make this far more performant. 
* When doing the split for partners - we should just do one query and then filter it out in Python. Your code should be able to handle 3 person groups.

server/controllers/admin.py
stanfieldchu: Resolved

server/controllers/admin.py
stanfieldchu: Resolved

server/controllers/analyze.py
```diff
@@ -0,0 +1,341 @@
...
+    time1_string, time2_string = analytics1.get("time"), analytics2.get("time")
+    if not (time1_string and time2_string):
+        return None
+    time_format = "%Y-%m-%d %X"
+    time1 = datetime.datetime.strptime(time1_string.split(".")[0], time_format)
+    time2 = datetime.datetime.strptime(time2_string.split(".")[0], time_format)
+    time_difference = time2 - time1
+    return time_difference.total_seconds()
+
+def sort_backups(backups):
```
stanfieldchu: Resolved

server/templates/diff.html
```diff
@@ -97,3 +97,36 @@
     </table>
   </div>
 {% endmacro %}
+
+{% macro diff_lines(file) %}
+  <div class="source-file table-responsive"
+    data-filename={{ filename }}>
+    <table class="highlight">
+      <tbody>
+        {% for line in file.lines %}
```
knrafto: This looks copy-pasted from above, but without the `<thead>`. The part starting from `{% for line in file.lines %}` should be factored into a jinja function

server/contrib/analyze.py
```diff
@@ -0,0 +1,345 @@
...
+        Returns backup b_p or None if nothing is found
+        """
+        nonlocal partner_index
+        while partner_index < backup_count:
+            backup = backups[partner_index]
+
+            analytics = backup and backup.analytics()
+            if analytics:
+                time_diff = _get_time_diff_seconds(analytics, current_analytics)
+                if time_diff < 0: # curr_analytics - analyics < 0 => too far ahead
```
Sumukh: Consecutive If statements should be joined

server/contrib/analyze.py
```diff
@@ -0,0 +1,345 @@
...
+
+        partner_backup_files = None
+
+        if analytics:
+            backup_stats['time'] = analytics.get('time')
+            partner_backup = get_recent_backup(analytics)
+            if partner_backup:
+                backup_stats["partner_backup_id"] = partner_backup.hashid
+                backup_stats["partner_bid"] = partner_backup.id
+            if partner_backup:
```
Sumukh: Can this just go in the previous case?

server/models.py
```diff
@@ -995,26 +1004,37 @@ def enrollment_info(self):
...
+        for message in self.messages:
+            if message.kind == 'grading':
+                return dict(message.contents)
+        return {}
+
+    def unlocking(self):
+        """ Return a string for which question the student is unlocking."""
+        for message in self.messages:
+            if message.kind == 'unlock':
+                if len(message.contents):
```
Sumukh: Consecutive If statements should be joined

server/contrib/analyze.py
stanfieldchu: Yes!

server/contrib/analyze.py
stanfieldchu: I dont think this consecutive `if` can be joined without making the code more complicated.
```
if time_diff < 0 and partner_index > 0: # curr_analytics - analyics < 0 => too far ahead
    prev_backup = backups[partner_index-1]
    return prev_backup
elif time_diff < 0:
    return None
```

Sumukh: Whoops - I agree. The line I left it on doesn't even match the definition of consecutive if statements

Sumukh:
This might be ready to merge once the conflicts are resolved.

knrafto:
I'm gonna force-push this back and try the merge again locally so I can test it :(

Sumukh:
Tests are passing again - is this good to deploy?

knrafto:
I think it's good to merge

PR #887, opened 2016-09-29
Adds CourseEnrollment resource to find students

### LINKED ISSUE #851: Enrollment API for a Course
View all students enrolled in the course. 

Makes it easier to write export scripts. 

Files to look at: 
- `controllers/api.py`
- `models.py` 
- `documentation/ok-api.md`

tanaythan:
Returns all students enrolled in a specific course

Resolves #851

### DIFF
--- a/server/controllers/api.py
+++ b/server/controllers/api.py
@@ -28,7 +28,7 @@ from server.extensions import cache
 from server.utils import encode_id, decode_id
 import server.models as models
 from server.autograder import submit_continous
-from server.constants import STAFF_ROLES
+from server.constants import STAFF_ROLES, VALID_ROLES
 
 endpoints = Blueprint('api', __name__)
 endpoints.config = {}
@@ -246,6 +246,7 @@ class EnrollmentSchema(APISchema):
         'courses': fields.List(fields.Nested(ParticipationSchema.get_fields))
     }
 
+
 class UserSchema(APISchema):
     get_fields = {
         'id': HashIDField,
@@ -260,6 +261,9 @@ class UserSchema(APISchema):
         'email': fields.String,
     }
 
+class CourseEnrollmentSchema(APISchema):
+    get_fields = { role : fields.List(fields.Nested(UserSchema.simple_fields))
+                    for role in VALID_ROLES }
 
 class AssignmentSchema(APISchema):
     get_fields = {
@@ -700,6 +704,29 @@ class Enrollment(Resource):
             return True
         return resource == requester
 
+class CourseEnrollment(Resource):
+    """ Information about all students in a course.
+    Authenticated. Permissions: >= User or admins
+    Used by: Export scripts.
+    """
+    model = models.Course
+    schema = CourseEnrollmentSchema()
+
+
+    @marshal_with(schema.get_fields)
+    def get(self, offering, user):
+        course = self.model.by_name(offering)
+        if course == None:
+            restful.abort(404)
+        if not self.model.can(course, user, 'staff'):
+            restful.abort(403)
+        data = {}
+        for role in VALID_ROLES:
+            data[role] = []
+        for p in course.participations:
+            data[p.role].append(p.user)
+        return data
+
 class Score(Resource):
     """ Score creation.
         Authenticated. Permissions: >= Staff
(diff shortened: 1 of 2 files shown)


server/controllers/api.py
```diff
@@ -811,6 +824,7 @@ def get(self, user, email=None):
 
 # Other
 api.add_resource(Enrollment, '/v3/enrollment/<string:email>/')
+api.add_resource(CourseEnrollment, '/v3/course_enrollment/<string:cid>')
```
Sumukh: - cid should be an int right? Regardless - We're trying to do away with using course IDs (since no-one knows what the ID is off the top of their heads - and go by the course offering 

> '/v3/course_enrollment/<offering:offering>'

Sumukh: Also I think the URL route should be a bit more modular: 
I'd suggest the route: `/v3/course/<offering:offering>/enrollment` 

I think `/v3/course/cal/cs61a/fa16/enrollment`  is better `/v3/course_enrollment/4`

server/models.py
```diff
@@ -715,6 +715,10 @@ def create(cid, enrollment_info=None, role=STUDENT_ROLE):
 
         cache.delete_memoized(User.is_enrolled)
 
+    @staticmethod
+    def get_all_students(cid):
+        return Enrollment.query.filter_by(course_id=cid, role=STUDENT_ROLE).all()
```
Sumukh: Because of this line 
```
class Enrollment 
    ... 
    course = db.relationship("Course", backref="participations")
``` 

We can access all enrollments by simply doing `course.participations`. It doesn't filter for only students - but you can do that in the API endpoint.

server/controllers/api.py
```diff
@@ -672,6 +672,18 @@ def can(action, resource, requester):
...
+    """ Information about all students in a course.
+    Authenticated. Permissions: >= User or admins
+    Used by: Export scripts.
+    """
+    model = models.Enrollment
+    schema = EnrollmentSchema()
+
+    @marshal_with(schema.get_fields)
+    def get(self, cid, user):
+        return {'courses': self.model.get_all_students(cid)}
```
Sumukh: `EnrollmentSchema.get_fields` doesn't seem like a good fit for this. It's definitely not a list of courses. 

I'd suggest a schema that looks like 
```
{ 
  'students': fields.List,  
  'staff': fields.List, 
  'instructor': fields.List
  ... 
}
```
An easy way to generate that schema is to use a dictionary comprehension 
`all_enrollments_schema = { role: fields.List(...) for role in constants.VALID_ROLES } `

Sumukh: This doesn't provide any kind of authentication/permissions right now. That means that any user can get the contact information for students for any course. 

```
if not self.model.can(course, user, 'staff'): 
    restful.abort(403)
```

Sumukh:
Also before merging: 
- Add a test to test api. It should be as simple as copying one of the existing tests and changing the route inside of the test.

Sumukh:
Thanks for the PR @tanaythan  - do you think you can address the comments above? If not let us know and we can find someone else to make the changes.

tanaythan:
I am starting to work on the comments now.

tests/test_api.py
```diff
@@ -452,3 +452,17 @@ def test_both_endpoints(user):
         current, specific = test_both_endpoints(student)
         self.assert_200(current)
         self.assert_403(specific)
+
+    def test_course_enrollment(self):
+        self._test_backup(True)
+        student = User.lookup(self.user1.email)
+        courses = student.enrollments()
+        course = courses[0]
+        student_endpoint = '/api/v3/course/cal/cs61a/sp16/enrollment'
```
Sumukh: Nit:`'/api/v3/course/{}/enrollment'.format(course.offering)`

server/controllers/api.py
```diff
@@ -700,6 +704,29 @@ def can(action, resource, requester):
...
+
+    @marshal_with(schema.get_fields)
+    def get(self, offering, user):
+        course = self.model.by_name(offering)
+        if course == None:
+            restful.abort(404)
+        if not self.model.can(course, user, 'staff'):
+            restful.abort(403)
+        data = {}
+        for role in VALID_ROLES:
```
Sumukh: Nit/Doesn't need to be addressed - seems like a useful place for a defaultdict but I'm not 100% sure flask-restful will handle it (I think it will - but if this is only place we will use it then it's not worth importing)

tests/test_api.py
```diff
@@ -452,3 +452,17 @@ def test_both_endpoints(user):
...
+        self.login(self.staff1.email)
+        response = self.client.get(student_endpoint)
```
Sumukh: We should check that some field is in the dictionary. Maybe something like
```
staff_emails = [s['email'] for s for s in data['staff']]
assert self.staff1.email in staff_emails 
```

tests/test_api.py
tanaythan: I tried that, but `course.offering` didn't exist, so I hardcoded it.

PR #888, opened 2016-10-03
Create comment API

Sumukh:
Tested manually and through the test suite. Seems to work. 

I didn't add the ability to view comments from the API because no one has asked for it (also [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it))

### DIFF
--- a/documentation/api.md
+++ b/documentation/api.md
@@ -521,9 +521,7 @@ response = r.json()
 >><h4> Response </h4>
 ```
 {
-    "data": {
-        'html': "<div class='comment'> ... </div>"
-    },
+    "data": {},
     "code": 200,
     "message": "success"
 }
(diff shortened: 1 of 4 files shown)


documentation/api.md
```diff
@@ -504,6 +504,57 @@ access_token | None | (Required) Access Token of submitter
...
+url = "https://okpy.org/api/v3/backups/a24x23/comment?access_token={}"
+access_token = 'test'
+r = requests.post(url.format(access_token), data=data)
+response = r.json()
+```
+>><h4> Response </h4>
+```
+{
+    "data": {
+        'html': "<div class='comment'> ... </div>"
```
knrafto: The HTML is pretty specific to our frontend (e.g. it includes "edit" and "close" buttons), so it shouldn't be included in our API response. The client can just use their own markdown renderer if they want a preview.

Sumukh: Yeah. I was thinking of our own internal use case but it's probably better off not hitting the API. I'll make the data field empty in that case.

Sumukh:
@knrafto: I've reverted the HTML commit - feel free to merge if you don't have any other comments.

PR #894, opened 2016-10-05
Allow access to Group API endpoint from partner 

Sumukh:
This is needed by the collab editor.

### DIFF
--- a/server/controllers/api.py
+++ b/server/controllers/api.py
@@ -779,13 +779,15 @@ class Group(Resource):
             restful.abort(403)
 
         group = self.model.lookup(target, assign)
-        member_emails = [m.user.email.lower() for m in group.members]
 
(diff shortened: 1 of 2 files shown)


server/controllers/api.py
```diff
@@ -779,9 +779,11 @@ def get(self, user, name, email):
             restful.abort(403)
 
         group = self.model.lookup(target, assign)
+        member_emails = [m.user.email.lower() for m in group.members]
+
         is_admin = user.is_admin
         is_staff = user.is_enrolled(assign.course.id, STAFF_ROLES)
-        is_self = user.email.lower() == email.lower()
+        is_self = user.email.lower() in member_emails
```
knrafto: Nit: should be `is_group_member` or similar, rather than `is_self`

PR #895, opened 2016-10-05
[WiP] Speedup Travis

Sumukh:
Lets see if this helps Travis get a bit faster.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,13 +4,13 @@ language: generic
 services:
   - docker
 python:
-    - "3.4"
-install: "pip3 install -r requirements.txt"
+    - "3.5"
+install: "python3 -m pip install -r requirements.txt"
 script:
   - make test
   - make docker-build
 after_success:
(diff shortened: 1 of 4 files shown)


Sumukh:
I think this is going to have to turn into a fix travis branch that makes travis run without docker and use mysql natively.

Sumukh:
I think this is the ideal config: 
- Travis (use the native build matrix to make this happen)
    - Build the docker image (and ideally test in the docker env - though now that seems to require redis) 
    - Run with normal config (python 3.5 and sqlite) & coveralls 
- CircleCI: Run with mysql5.7 and parallelize.

Sumukh:
I'm going to merge this so we at least have passing CI. The things that are missing: 
- Accurate coverage information. 

Travis CI runs this without test_web because it seems to fail there. When using CircleCI's parallelization, coveralls seems to report inaccurate coverage numbers.

PR #896, opened 2016-10-05
Prevent Cross-Site Scripting

kvakil:
The Markup class assumes that the string is HTML safe. Therefore no
filtering is performed on comments. This fix uses jinja2.escape()
before the data is passed to the markdown class, ensuring that all
HTML tags are properly filtered.

### DIFF
--- a/server/__init__.py
+++ b/server/__init__.py
@@ -8,6 +8,7 @@ from flask_rq import RQ
 from flask_wtf.csrf import CsrfProtect
 from webassets.loaders import PythonLoader as PythonAssetsLoader
 from werkzeug.contrib.fixers import ProxyFix
+from jinja2 import escape
 
 from server import assets, converters, utils
 from server.forms import CSRFForm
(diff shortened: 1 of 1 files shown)


Sumukh:
Thanks. I think we had assumed that the markdown function would take care of XSS but it doesn't.

The tests have passed (I think CircleCI just can't submit the job to coveralls)

server/__init__.py
```diff
@@ -108,7 +109,7 @@ def not_found_error(error):
     })
 
     app.jinja_env.filters.update({
-        'markdown': lambda data: Markup(markdown(data)),
+        'markdown': lambda data: Markup(markdown(escape(data))),
```
Sumukh: This is good - but I also think we should also escape when we accept comments.

Sumukh:
I've added a test to ensure that stored XSS are not presented into our integration tests. I'll merge this soon.

Sumukh:
@kvakil - You are eligible for a security bug bounty. Come into lecture on Friday and find me for your reward!

Thanks for the PR.

kvakil:
@Sumukh, ooh exciting, I'll be there!

Issue #883, opened 2016-09-28
Roles should be displayed differently


knrafto:
In drop down menus, we should show more humanized roles:
* staff -> Teaching Assistant
* grader -> Reader
* lab assistant -> Lab Assistant

Sumukh:
Files/Lines to edit: https://github.com/Cal-CS-61A-Staff/ok/blob/master/server/forms.py#L161

shivasheeshyadav:
@knrafto @Sumukh I think i've found the piece of code which needs to edited. I would like to take up this issue and work on it.

shivasheeshyadav:
@knrafto @Sumukh I just made a pull request regarding this issue. But all checks have failed. As I'm new to Open Source, I don't know about these checks. Please help.

PR #899, opened 2016-10-10
#883 patch,Roles displayed

shivasheeshyadav:
#883 Issue fixed.

### DIFF
--- a/server/constants.py
+++ b/server/constants.py
@@ -5,21 +5,12 @@ GRADER_ROLE = 'grader'
 STAFF_ROLE = 'staff'
 INSTRUCTOR_ROLE = 'instructor'
 LAB_ASSISTANT_ROLE = 'lab assistant'
-<<<<<<< HEAD
-ROLES_NEW={
-'student':'Student',
-'grader':'Reader',
-'staff':'Teaching Assistant',
-'instructor':'Instructor',
-'lab assistant':'Lab Assistant'
-=======
 ROLE_DISPLAY_NAMES = {
     'student' : 'Student',
     'grader' : 'Reader',
     STAFF_ROLE : 'Teaching Assistant',
     'instructor' : 'Instructor',
     'lab assistant' : 'Lab Assistant',
->>>>>>> master
 }
 VALID_ROLES = [STUDENT_ROLE, LAB_ASSISTANT_ROLE, GRADER_ROLE, STAFF_ROLE,
                INSTRUCTOR_ROLE]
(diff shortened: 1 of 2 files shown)


Sumukh:
You should not change the actual roles themselves but rather how they are displayed in the form. The only file you should be changing is forms.py - https://github.com/Cal-CS-61A-Staff/ok/blob/master/server/forms.py#L161

The documentation on the form library is here: http://wtforms.simplecodes.com/docs/1.0.1/crash_course.html

shivasheeshyadav:
@Sumukh So you mean to say that i should not change the text of the already given choices, instead add more choices ?

Sumukh:
Not quite. The strings in the constants.py file should not be changed. If you look at line 170 of forms.py (https://github.com/Cal-CS-61A-Staff/ok/blob/master/server/forms.py#L170) you can see how the fields are rendered. Current it's` [r, r.title() for r in roles ]`  which results in `[('student', 'Student'), ('grader', 'Grader')...]` That second value is what's rendered in the form. 

We should keep the raw values of the roles (`['student', 'grader', ...]`) the same. 

Here's one approach that could work. Define a dictionary of role_names in constants.py that maps each of the role to the desired name
```
role_names = { 
   'grader': 'Tutor', 
   'student': 'Student', 
    ...
}
```
and then in the form you can do `[(r, role_names[r]) for r in VALID_ROLES)`

server/constants.py
```diff
@@ -5,6 +5,13 @@
 STAFF_ROLE = 'staff'
 INSTRUCTOR_ROLE = 'instructor'
 LAB_ASSISTANT_ROLE = 'lab assistant'
+ROLES_NEW={
```
knrafto: `ROLES_NEW` is not a very descriptive name. `ROLE_DISPLAY_NAMES` would be better

server/constants.py
```diff
@@ -5,6 +5,13 @@
...
+'student':'Student',
+'grader':'Reader',
+'staff':'Teaching Assistant',
```
knrafto: We should reuse `STAFF_ROLE` from above instead of repeating the string literal `'staff'`

server/forms.py
```diff
@@ -168,7 +168,7 @@ class EnrollmentForm(BaseForm):
     section = StringField('Section',
                           validators=[validators.optional()])
     role = SelectField('Role', default=STUDENT_ROLE,
-                       choices=[(r, r.title()) for r in VALID_ROLES])
+                       choices=[(r, ROLES_NEW[r]) for r in VALID_ROLES])
```
knrafto: I think simply `choices=ROLES_NEW.items()` would work, but I'm not 100% sure

shivasheeshyadav:
@knrafto I've made all the changes as requested.

PR #911, opened 2016-10-14
Timeline View of Student Progress

Sumukh:
Needs a small bit of cleanup. I mainly want to make the main timeline generation loop cleaner. 

Example from local view: 
image


server/controllers/admin.py
```diff
@@ -835,6 +836,95 @@ def student_view(cid, email):
...
+                       .outerjoin(Message)
+                       .filter(Backup.submitter_id.in_(user_ids),
+                               Backup.assignment_id == assign.id,
+                               Message.kind == "analytics")
+                       .order_by(Backup.created.asc())
+                       .all())
+
+    unlock_started_q = {}
+    started_questions = {}
+    solved_questions = {}
```
Sumukh: At some point these needed to be dictionaries - but these can just be sets.

server/controllers/admin.py
```diff
@@ -835,6 +836,95 @@ def student_view(cid, email):
...
+    submitter_counts = {}
+    history, timeline = [], []
+    last_q = (None, False, 0)
```
Sumukh: I think this `last_q` makes it harder to understand what exactly is happening when we modify/read it. Being a dictionary or a namedtuple would make the usage clearer. 

`x[2]` vs `x['count']` 

I think namedtuple is the way to go

server/controllers/admin.py
```diff
@@ -835,6 +836,95 @@ def student_view(cid, email):
...
+        contents = message.contents
+        current_question = contents.get('question', [None])[0]
+        if not current_question:
+            continue
+        if ('history' not in contents or 'questions' not in contents['history'] or
+            type(contents['history']['questions']) != dict):
+            continue
+
+        if backup.submitter.email not in submitter_counts:
+            submitter_counts[backup.submitter.email] = 1
```
Sumukh: Should be a defaultdict

server/controllers/admin.py
```diff
@@ -835,6 +836,95 @@ def student_view(cid, email):
...
+        else:
+            submitter_counts[backup.submitter.email] += 1
+
+        curr_q_stats = message.contents['history']['questions'].get(current_question)
+        total_attempt_count = message.contents['history'].get('all_attempts')
+        is_solved = curr_q_stats.get('solved', None)
```
Sumukh: `curr_q_stats` could be None - causing this to crash

server/controllers/admin.py
```diff
@@ -835,6 +836,95 @@ def student_view(cid, email):
...
+                last_q = (current_question, is_solved, last_q[2]+1)
+
+        if current_question not in started_questions:
+            started_questions[current_question] = backup.hashid
+            timeline.append({"event": "Started", "title": "Started {}".format(current_question),
+                             "backup": backup})
+        elif last_q[0] != current_question and last_q[0] is not None:
+            # Didn't just start it but did switch questions.
+            timeline.append({"event": "Switched", "title": "Switched to {}".format(current_question),
+                             "body": "{} Backups Later. Attempt: {}".format(last_q[2], total_attempt_count),
```
Sumukh: Attempt is a bit unclear. It's the student specific attempt number

server/generate.py
```diff
@@ -161,20 +161,60 @@ def gen_enrollment(user, course):
...
+    fizzbuzz_lines = modified_file.split("\n")
+    random.shuffle(fizzbuzz_lines)
+    shuffled_file = "\n".join(fizzbuzz_lines)
+
+    if seconds_offset < 0:
+        progress = abs(seconds_offset)/100100
+    else:
+        progress = (seconds_offset+100000)/100100
+
+    attempts = int(progress * 50)
```
Sumukh: arbitrarily decided that 50 attempts will be the upper limit

server/generate.py
```diff
@@ -161,20 +161,60 @@ def gen_enrollment(user, course):
...
+                    'fizzbuzz': {
+                        'attempts': attempts,
+                        'solved': is_passed
+                    }
+                },
+                'all_attempts': attempts
+            },
+            'unlock': gen_bool(),
+            'question': ['fizzbuzz'],
+            'started': {},
```
Sumukh: Most of this is cherrypicked from enhancements/schu/diff-view

the change is that I refactored this a bit to separate gen_message from gen_backup

server/static/css/staff.css
```diff
@@ -24,3 +24,9 @@
 .email-reminder {
     margin-right: 20px;
 }
+
+.timeline-inverted {
```
Sumukh: Not used anymore - can be removed.

server/templates/staff/student/assignment.timeline.html
```diff
@@ -0,0 +1,94 @@
...
+                      <!-- timeline time label -->
+                      <li class="time-label">
+                          <span class="bg-red">
+                              {{ submitter }} ( {{ submitters[submitter] }} backups)
+                          </span>
+                      </li>
+                      <!-- /.timeline-label -->
+                      {% for event in timeline %}
+                      <!-- timeline item -->
+                      <li style="{% if event['backup'].submitter.email == submitter %} visibility: hidden;{% endif %}" >
```
Sumukh: Should just have a CSS util class that does this instead of specifying a style inline

server/generate.py
```diff
@@ -118,8 +118,8 @@ def gen_assignment(course):
     name = course.offering + '/' + display_name.lower().replace(' ', '')
 
     last_night = (datetime.datetime.utcnow()
-                          .replace(hour=0, minute=0, second=0, microsecond=0)
-                          - datetime.timedelta(seconds=1))
```
knrafto: It's whatever but https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator

server/generate.py
```diff
@@ -293,11 +333,11 @@ def seed_scores():
     admin = User.query.filter_by(is_admin=True).first()
     for backup in Backup.query.filter_by(submit=True).all():
         if random.choice([True, False]):
```
knrafto: `if gen_bool():`

Sumukh:
Did you run `./manage.py resetdb`? (Otherwise the old db won't have messages with analytics)

server/generate.py
Sumukh: Makes sense. Interestingly Flake8 threw a W503 at me when the line break was before the operator. Seems like that's flake8/pycodestyle hasn't been updated to the latest pep8 change 

https://gitlab.com/pycqa/flake8/issues/139

Sumukh:
^ Looks like a bad merge for tests.

server/models.py
```diff
@@ -375,6 +376,85 @@ def by_name(name):
         """ Return assignment object when given a name."""
         return Assignment.query.filter_by(name=name).one_or_none()
 
+    def user_timeline(self, user_id):
+        """ Timeline of user submissions. """
```
knrafto: Can you document the type of the return value?

server/models.py
```diff
@@ -375,6 +376,85 @@ def by_name(name):
...
+        analytics = (db.session.query(Backup, Message)
+                       .outerjoin(Message)
+                       .filter(Backup.submitter_id.in_(user_ids),
+                               Backup.assignment_id == self.id,
+                               Message.kind == "analytics")
+                       .order_by(Backup.created.asc())
+                       .all())
+
+        unlock_started_q, started_questions, solved_questions = {}, {}, {}
+        submitter_counts = defaultdict(lambda: 1)
```
knrafto: `collections.Counter` would be more appropriate

server/models.py
```diff
@@ -375,6 +376,85 @@ def by_name(name):
...
+        history, timeline = [], []
+        last_q = (None, False, 0)  # current_question, is_solved, count
+
+        for backup, message in analytics:
+            contents = message.contents
+            curr_q = contents.get('question', [None])[0]
```
knrafto: what if `contents['question']` is an empty list?

server/models.py
```diff
@@ -375,6 +376,85 @@ def by_name(name):
...
+            if not curr_q:
+                continue
+            if ('history' not in contents or 'questions' not in contents['history'] or
+                    not isinstance(contents['history']['questions'], dict)):
```
knrafto: In what case would this not be a dict?

server/models.py
```diff
@@ -375,6 +376,85 @@ def by_name(name):
...
+
+            curr_q_stats = message.contents['history']['questions'].get(curr_q)
+            total_attempt_count = message.contents['history'].get('all_attempts')
+            is_solved = curr_q_stats.get('solved')
+
+            if contents.get('unlock'):
+                # Is unlocking.
+                if curr_q not in unlock_started_q:
+                    unlock_started_q[curr_q] = backup.hashid
+                    timeline.append({"event": "Unlock",
```
knrafto: namedtuple?

knrafto:
Nothing blocking though, feel free to merge

Sumukh:
There's no rush to merge so I'll try to address them. I'll at least fix the tests.

server/generate.py
```diff
@@ -333,12 +333,12 @@ def seed_scores():
...
-            db.session.add(score)
-        if gen_bool():
-            score = gen_score(backup, admin, kind='total')
-            db.session.add(score)
+        if gen_bool(0.6):
+             score = gen_score(backup, admin, kind='composition')
+             db.session.add(score)
+        if gen_bool(0.8):
+             score = gen_score(backup, admin, kind='total')
+             db.session.add(score)
```
knrafto: 5 spaces??

Sumukh:
Going to merge & deploy after tests run.

PR #913, opened 2016-10-16
Generate tokens for the autograder and eliminate all fields

knrafto:
Now autograding could not be easier. The form is now a single button.


server/autograder.py
```diff
@@ -30,34 +33,40 @@ def send_autograder(endpoint, data):
...
-    @token: OK Access Token (from auth)
+    @assignment: Assignment object
     """
+    if not assignment.autograding_key:
+        raise ValueError('Assignment has no autograder key')
+
+    # Create an access token for this run
+    autograder_client = Client.query.get('autograder')
+    if not autograder_client:
+        raise ValueError('Autograder OAuth client does not exist')
```
Sumukh: It seems to easy for us to just add the "autograder" client here. Why not make it easier for people on the first install?

Sumukh:
What's the point of a new page & a template for this? If we remove this page, we could make it even easier to autograde in that you don't have to load an extra page that just has a button

Is it to avoid accidental presses? (We do have an easy to use javascript form confirmation)

We might as well make the form appear in the assignment overview. (We should keep the route though)

knrafto:
I tried putting the form in the "Actions" dropdown but I couldn't get the CSS to work. How else can I put the autograder button on that page?

Sumukh:
Alternative: 
image


The text is a bit too small for readability though. 

```
                        <button type="submit" class="ag-submit-btn" data-confirm="Are you sure you want to send to the autograder?">
                        {% if assignment.autograder_key %}
                          <i class="fa fa-send"></i>
                        {% else %}
                            <i class="fa fa-ban"></i>
                        {% endif %}
                          Queue on Autograder
                        {% if not assignment.autograder_key %}
                          <span class="label label-warning pull-right">No Autograder Key</span>
                        {% endif %}
                        </button>
```

Sumukh:
It makes a lot of sense to hide the button - I'm just worried about it not being clear that people have to enter in an autograder key to get the autograder button to show up. 

With more thought - it's becoming clear that displaying the form is the easiest/clearest option for now

Sumukh:
@knrafto - Feel free to review my changes & merge if you want.

knrafto:
Thanks for your work on the CSS and the tests! Going to merge and deploy

PR #921, opened 2016-10-19
Change revision score to 2.0

knrafto:


knrafto:
Once this is merged, I'll write an `UPDATE` to change past scores for fall 2016 cs61a.

Issue #922, opened 2016-10-19
Multiline code blocks are rendered as a single line


knrafto:
![multiline-bug](https://cloud.githubusercontent.com/assets/2042019/19537160/d39dd9ba-9603-11e6-8849-1ae6410bdbb3.gif)

knrafto:
Closed in #925

PR #925, opened 2016-10-20
Fix multi-line code blocks

kevinlin1:
#922. Preview still renders incorrectly since markdown.js doesn't support Github-flavored markdown.

screenshot 2016-10-19 at 6 04 34 pm

### DIFF
--- a/server/templates/diff.html
+++ b/server/templates/diff.html
@@ -19,7 +19,7 @@
       </div>
       {% endif %}
     </div>
-    <div class="comment-body markdown-body">{{ comment.message | safe | markdown }}</div>
+    <div class="comment-body markdown-body">{{ comment.message | markdown }}</div>
   </div>
 {% endmacro %}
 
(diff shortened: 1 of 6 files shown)


Sumukh:
It look like this change introduces an XSS vuln (see circleci)

server/templates/diff.html
```diff
@@ -19,7 +19,7 @@
       </div>
       {% endif %}
     </div>
-    <div class="comment-body markdown-body">{{ comment.message | markdown }}</div>
+    <div class="comment-body markdown-body">{{ comment.message | safe | markdown }}</div>
```
kevinlin1: This seems to cause the problem but at the cost of breaking valid characters like `'`.

kevinlin1:
image

knrafto:
I think I fixed all the double-escaping issues. I made another issue for the JS library (#926)

knrafto:
LGTM, merging

Issue #877, opened 2016-09-22
Add emoji to composition text box.


Sumukh:
It'd be cool to type in emoji's kind of like how github supports it. `:smile:` -> 😄 

one promising library: 
https://github.com/diy/jquery-emojiarea

shivasheeshyadav:
Hey,
I'm completely new to Open Source and would like to work on this issue.

Sumukh:
@shivasheeshyadav - awesome. This is all yours.  Let us know if you have any questions!

shivasheeshyadav:
Thanks.
So which is the particular file in the codebase which i'll be working on ?

Sumukh:
This issue would touch a lot of files - you'll be adding a javascript library to the page and it would interact with the editor here. 

https://github.com/Cal-CS-61A-Staff/ok/blob/master/server/templates/diff.html

As a good first commit - I'm might suggest tackling #883 first.

shivasheeshyadav:
@Sumukh Ok. I'll tackle #883 first, and then 'll get back to this issue. Thanks for the suggestion.

kevinlin1:
Should be pretty simple with [`markdown-it-emoji`](https://www.npmjs.com/package/markdown-it-emoji).

colinschoen:
Is this WIP? If not I'll take this.

colinschoen:
Not necessary. Closed.

PR #927, opened 2016-10-20
Real-time comment previews

### LINKED ISSUE #926: Use a better JS markdown library, with live previews
The current library has problems with multiline code blocks. See #925

Also, we should be able to preview the comment in real time, instead of having to change tabs.

@Sumukh suggested Vue.js.

kevinlin1:
Resolves #926.

image

Real-time comment previews with debouncing. Some of the styles are from #925, so the code blocks don't look as nice as they are in the screenshot.

### DIFF
--- a/server/static/css/code.css
+++ b/server/static/css/code.css
@@ -166,15 +166,10 @@ table.highlight tr:last-child td.comment-container {
   color: #333;
   font-size: 14px;
   line-height: 1.6;
-  padding: 15px;
+  margin: 15px;
 }
 
 /* Editor */
-.comment-editor-header {
-  background: #f7f7f7;
-  border-radius: 3px 3px 0 0;
-}
-
 .comment-editor .nav {
   padding: 6px 10px 0;
 }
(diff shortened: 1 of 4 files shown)


knrafto:
Rebasing off master, excuse the force-push

Sumukh:
Is `markdown-it` a requirement? It's really 5x the size of `markdown.js`

Sumukh:
Also do you have an opinion on doing a side by side render? I kind of see the advantages of the current approach - you can see the comment at full width.

knrafto:
I think we can decrease the render delay by quite a bit. It feels very sluggish now. I did some profiling and rendering takes < 5ms on my machine, so don't be afraid to put a low number.

kevinlin1:
`markdown-js` is pretty out of date and doesn't support multi-line code blocks. Another option is to use [showdown](https://github.com/showdownjs/showdown) although it doesn't have a strong security policy.

Sumukh:
I'm fine with your choice of library (markdown-it) - I was mainly curious. It certainly seems like the best option




On Wed, Oct 19, 2016 at 10:31 PM -0700, "Kevin Lin" <notifications@github.com> wrote:












markdown-js is pretty out of date and doesn't support multi-line code blocks. Another option is to use showdown although it doesn't have a strong security policy.

kevinlin1:
Side-by-side gets a little cluttered since the comment width is constrained and I think it's nice to have a preview reflect exactly what you'll get.

knrafto:
We don't care about security for this use case, since we're not using it server-side. The worst the user could do is XSS themselves.

knrafto:
But I'm also fine with this library. I can't find any differences between this library's output and our server's

kevinlin1:
There are also some tools like Babelmark if you want to test for rendering differences between Markdown implementations.

image

Possible dual-pane demo. What do you think?

Sumukh:
I like it! But the vertical view has its merits. Either way I think there should be some indication of seperation between the two components (a `<hr>` or a vertical `<hr>`).  It seems like most markdown editors do a side by side render. 

I'll defer to you - not sure what's better here.

kevinlin1:
Side effect of switching to markdown-it is that it probably makes it [easier to integrate emoji](https://www.npmjs.com/package/markdown-it-emoji) too #877.

kevinlin1:
I'm personally very partial to the vertical layout (the way the code is now). There's subtle borders between the editable textarea and the preview region, but I do see how it might still look a little too similar. Not sure the best way to resolve this either.

knrafto:
I think it's good. Merging and deploying. Thanks Kevin!

Issue #929, opened 2016-10-21
Fix QuantifiedCode issues


Sumukh:
QuantifiedCode has a bunch of code quality suggestions.

It'd be a good idea to address those. 

![image](https://cloud.githubusercontent.com/assets/882381/19587413/0b98ba3c-9713-11e6-8ba8-1bb5133c451e.png)

https://www.quantifiedcode.com/app/project/e6ae811f0c244655b5026e0883ffb8f9

If you are a first time contributor and only want to tackle some of them - that's fine.

the-zebulan:
@Sumukh ,

I'll try to fix a few of these right now.

Sumukh:
Resolved by @the-zebulan in #930

Sumukh:
Nice!

![image](https://cloud.githubusercontent.com/assets/882381/19590369/5b61f566-9726-11e6-87cb-4cd977e59430.png)

![image](https://cloud.githubusercontent.com/assets/882381/19590371/5f9684da-9726-11e6-905a-0a115caf3f03.png)

PR #930, opened 2016-10-21
QuantifiedCode_issues

the-zebulan:
Issue: #929

* Fixed "non-Pythonic" method name
* Converted to set comprehension
* Used `is None` instead of `== None`
* Fixed "non-Pythonic" variable names
* Used an iterator instead of list comprehension inside of `any`
* Removed `None` as redundant second argument to `dict.get()`
* Merged consecutive if-statements (also used `elif`'s instead of multiple `if`'s)

### DIFF
--- a/server/models.py
+++ b/server/models.py
@@ -767,7 +767,9 @@ class Backup(Model):
             return user.is_authenticated
         elif not obj:
             return False
-        elif user.is_admin or (action == "view" and user.id in obj.owners()):
+        elif user.is_admin:
+            return True
+        elif action == "view" and user.id in obj.owners():
             # Only allow admin or group members to view
             return True
         return user.is_enrolled(obj.assignment.course.id, STAFF_ROLES)
--- a/server/models.py
+++ b/server/models.py
@@ -41,6 +41,7 @@ convention = {
 metadata = MetaData(naming_convention=convention)
 db = SQLAlchemy(metadata=metadata)
 
+
 def transaction(f):
     """ Decorator for database (session) transactions."""
     @functools.wraps(f)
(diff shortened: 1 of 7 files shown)


server/models.py
```diff
@@ -755,14 +765,11 @@ class Backup(Model):
     def can(cls, obj, user, action):
         if action == "create":
             return user.is_authenticated
-        if not obj:
+        elif not obj:
             return False
-        if user.is_admin:
+        elif user.is_admin or (action == "view" and user.id in obj.owners()):
```
Sumukh: Do you think it makes sense to break the second part of this conditional into it's own `elif` branch? If we think of each conditional representing a specific case - the case of something being an admin is very different from being an owner of the object. 

I think quantifiedcode will be just as happy with 
```
elif user.is_admin
   return True 
elif action == "view"  and user.id in obj.owners():
   return True
```

the-zebulan:
Uh oh, tests failed?

Sumukh:
I think the test failure is unrelated - it's because we haven't mocked the autograder. I'll look into it and try to rerun the test.

the-zebulan:
Ok, just let me know if I messed something up. If so, I will try and fix it ASAP.

Sumukh:
Github PR tool: In the future if you include the phrase `Resolves #929` in your first comment - the issue will automatically be closed when the PR is merged :) 

https://help.github.com/articles/closing-issues-via-commit-messages/

the-zebulan:
Thanks for the tip!

I didn't know typing that would close the issue when the PR was merged. I'll make sure to use that message in the future.

PR #934, opened 2016-10-23
Handle spaces in StringList

Sumukh:
Useful for things with spaces (such as `Partner A`)

### DIFF
--- a/server/models.py
+++ b/server/models.py
@@ -122,7 +122,12 @@ class StringList(types.TypeDecorator):
         return ' '.join(items)
 
     def process_result_value(self, value, dialect):
-        # SQL -> Python
+        """ SQL -> Python
+        Uses shlex.split to handle values with spaces.
+        It's a fragile solution since it will break in some cases.
+        For example if the last character is a backslash or otherwise meaningful
+        to a shell.
+        """
         values = []
         for val in shlex.split(value):
             if " " in val and '"' in val:
--- a/server/models.py
+++ b/server/models.py
@@ -115,7 +115,7 @@ class StringList(types.TypeDecorator):
         # Python -> SQL
         items = []
         for item in string_list:
-            if " " in item:
+            if " " in item or not item:
                 items.append('"{}"'.format(item))
             else:
                 items.append(item)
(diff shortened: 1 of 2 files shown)


Sumukh:
- Handle cases where the python value is None. -> set that to `''`

Do you think this makes sense to add?

server/templates/staff/course/assignment/assignment.publish.html
```diff
@@ -34,6 +34,7 @@ <h3 class="box-title">Publish Scores</h3>
 
           </div>
           <div class="box-body">
+            {{ assignment.published_scores }}
```
Sumukh: Whoops - didn't mean to commit this. Might as well improve it a bit.

knrafto: We can make this slightly nicer? Like `Published scores: ', '.join(kind.title() for kin in assignment.published_scores)` or something

server/models.py
```diff
@@ -112,12 +113,23 @@ class StringList(types.TypeDecorator):
...
+            else:
+                items.append(item)
+        return ' '.join(items)
 
     def process_result_value(self, value, dialect):
         # SQL -> Python
-        return value.split()
-
+        values = []
+        for val in shlex.split(value):
```
knrafto: This will break if the value e.g. ends in a backslash (`foo \`) or if it has other characters that have meaning to a shell. I don't have a better way though so can you add a comment that this is fragile?

knrafto:
We should have to handle `None`, since this is a StringList, not a StringOrNoneList.

That's a good point though - this will break with empty strings too.

server/models.py
Sumukh: Yeah - I couldn't find a less fragile way to do it (most solutions involved some kind of regex)

server/models.py
```diff
@@ -112,12 +113,23 @@ class StringList(types.TypeDecorator):
 
     def process_bind_param(self, string_list, dialect):
         # Python -> SQL
-        return ' '.join(string_list)
+        items = []
+        for item in string_list:
+            if " " in item:
```
knrafto: `if " " in item or not item:`

PR #937, opened 2016-10-23
Use a list of checkboxes to publish grades

knrafto:
publish-scores

### DIFF
--- a/tests/test_grading.py
+++ b/tests/test_grading.py
@@ -5,6 +5,7 @@ from io import StringIO
 
 from werkzeug.exceptions import BadRequest
 
+from server.constants import SCORE_KINDS
 from server.models import db, Backup, Group, Message, GradingTask, Score
 import server.utils as utils
 from server import generate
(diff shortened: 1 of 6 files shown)


knrafto:
I just discovered the `MultiCheckboxField`. I'll probably use that instead.

Sumukh:
Thanks. I will look at this later tonight.

Sumukh:
Would it be useful to have a count of the number of scores that exist for each one?

knrafto:
Probably, but I'd consider that an out-of-scope enhancement

tests/test_grading.py
```diff
@@ -224,85 +237,50 @@ def check_visible_scores(user, assignment, hidden=(), visible=()):
...
         # If assignment is not visible, still can publish
         self.assignment.visible = False
         self.login(self.staff1.email)
-        endpoint = '/admin/course/{}/assignments/{}/publish'.format(self.course.id, self.assignment.id)
-        response = self.client.post(endpoint, data={}, follow_redirects=True)
-        self.assert_200(response)
+        response = publish_scores(self.assignment, ['Total'])
         source = response.get_data().decode('utf-8')
-        self.assertTrue("Published {} {} scores".format(self.assignment.display_name, 'Total') in source)
+        self.assertTrue("Saved published scores for {}".format(self.assignment.display_name) in source)
```
Sumukh: Sidenote: I wish we could have used the `assertMessageFlashed` flask-testing gives use

PR #940, opened 2016-10-24
Edit submission time on backups

### LINKED ISSUE #842: Staff submissions for students should not be marked late by autograder
Additionally, there should be an option to get the 1-day early EC point.

~~The simple solution is to have an option to set the `created` time (with nice defaults) when staff members submit.~~

knrafto:
Resolves #842.

When staff submits, there's now an option to set the submission time. It defaults to the submission creation time. If moved the staff submission to a separate page to make form errors easier to handle (both for the programmer and the user).

The grading page shows the custom submission time (if any), and allows you to edit the submission time after-the-fact.


server/controllers/admin.py
```diff
@@ -440,7 +467,7 @@ def templates(cid, aid):
         if files:
             templates = {}
             for template in files:
-                templates[template.filename] = str(template.read(), 'latin1')
+                templates[template.filename] = str(template.read(), 'utf-8')
```
Sumukh: Any reason this was changed?  AFAIK `latin1` was used specifically instead of `utf-8`

knrafto: I found the commit: 22b7f32

Seems like we should store files as binary data rather than as a string, but since we store files as json we have to encode it somehow. I'll look how ok-client does it

knrafto: ok-client assumes files are utf-8 text. I think it's reasonable that web uploads and templates have the same restriction

Sumukh: Does it prevent uploads of zip files or images for example? (Scheme Art is a use case for submitting pictures)

Sumukh: ^ that's specifically for web uploads. Keeping templates to something that's just utf-8 text should be fine.

knrafto: Made #941. I think we'll still 500 on an invalid utf-8 text file though

server/models.py
```diff
@@ -776,13 +776,20 @@ class Backup(Model):
...
     assignment_id = db.Column(db.ForeignKey("assignment.id"), nullable=False)
     submit = db.Column(db.Boolean(), nullable=False, default=False)
     flagged = db.Column(db.Boolean(), nullable=False, default=False)
 
     extension = db.Column(db.Boolean(), default=False)
 
-    submitter = db.relationship("User")
+    # The time we should treat this backup as being submitten. If NULL, use
+    # the `created` timestamp instead.
+    submission_time = db.Column(db.DateTime(timezone=True), nullable=True)
```
Sumukh: Do you think it makes sense to have a hybrid_property called `time` and it'll choose which to report? 

I can't see where we would use it right now - but it might be useful if we just want to display the time we treated it as.

server/models.py
```diff
@@ -776,13 +776,20 @@ class Backup(Model):
     id = db.Column(db.Integer, primary_key=True)
 
     submitter_id = db.Column(db.ForeignKey("user.id"), nullable=False)
+    # NULL if same as submitter
+    creator_id = db.Column(db.ForeignKey("user.id"), nullable=True)
     assignment_id = db.Column(db.ForeignKey("assignment.id"), nullable=False)
     submit = db.Column(db.Boolean(), nullable=False, default=False)
     flagged = db.Column(db.Boolean(), nullable=False, default=False)
 
     extension = db.Column(db.Boolean(), default=False)
```
Sumukh: Should we remove this property? Once we get full extension support - maybe it'll be useful but for now - it seems like it won't be

server/forms.py
```diff
@@ -240,9 +292,57 @@ class CreateTaskForm(BaseForm):
                                    default=False)
 
 class UploadSubmissionForm(BaseForm):
-    upload_files = FileField('Submission Files', [FileRequired()])
-    flag_submission = BooleanField('Flag this submission for grading',
-                                   default=False)
+    upload_files = BackupUploadField('Submission Files', [FileRequired()])
```
Sumukh: Is the removal of `flag_submission` intentional? I'd be okay with it but just want to make sure. 

Students/Staff can still flag by clicking on the specific submission - but for staff specifically keeping this option might be useful if the submission they created is going to vanish in their submission history (For example staff submits something for yesterday and its now on page 2 of the history) 

^ This isn't an issue if we keep sorting by created time though.

server/forms.py
```diff
@@ -33,6 +53,38 @@ class MultiCheckboxField(SelectMultipleField):
...
+
+        files = {}
+        for upload in request.files.getlist(self.name):
+            data = upload.read()
+            if len(data) > 2 * 1024 * 1024:  # file is too large (over 2 MB)
+                self.errors.append(
+                    '{} is larger than the maximum file size '
+                    'of 2MB'.format(upload.filename))
+                return False
+            files[upload.filename] = str(data, 'utf-8')
```
Sumukh: To prevent 500s can we have a try catch? 
```
try: 
  files[upload.filename] = str(data, 'utf-8')
except UnicodeDecodeError
  flash("Could not upload {}".format(upload.filename), 'warning') 
```

server/forms.py
knrafto: Yeah, I think having flagging be a separate action is better, especially since it affects other backups than just the one being submitted

server/controllers/admin.py
```diff
@@ -438,7 +465,7 @@ def templates(cid, aid):
...
```
Sumukh: Try catch here as well?

PR #945, opened 2016-10-26
Add indicies to tables

### LINKED ISSUE #944: Add More Indices To Models
When I created the models I forgot that MySQL doesn't support partial indexes (but Postgres does). 

We should add some indices to speed up operations. 

```
Backup.submitter_id
Backup.submitter_id, Backup.assignment_id
Score.assignment_id, Score.kind, Score.archived
Enrollment.course_id, Enrollment.role
```

Sumukh:
Resolves #944 

Unrelated: I'm accepting suggestions for a commit spellchecker.


Sumukh:
Going to put OK in maintenance mode and then role out these indexes.

PR #935, opened 2016-10-23
Show max scores for assignments instead of final submission scores

knrafto:
Resolves #906 

Log in as nikkid@gmail.com and publish "total" and "correctness" scores.

You can see all scores for hog here: http://localhost:5000/admin/course/1/nikkid%40gmail.com/2

At these pages, you should see the max scores. Scores will link to the backup that the score is on. If there is a tie, the score from the most recent backup is chosen.
http://localhost:5000/cal/cs61a/sp16/
http://localhost:5000/cal/cs61a/sp16/hog/
http://localhost:5000/admin/course/1/nikkid%40gmail.com

PR #946, opened 2016-10-27
Show max scores for assignments instead of final submission scores

knrafto:
See #935 for original PR.

The scores query is very slow, and caused us downtime on Oct 26.


knrafto:
We decided to:

Keep a redundant user_id column on scores. This will never change, so it's safe.
Put the max score logic in Python, since the number of scores will likely be small.

Sumukh:
I'm not sure if it's good practice to backfill the data in the migration or not. I think it makes sense so I'll add that to the migration.

knrafto:
ready for review now

server/models.py
```diff
@@ -586,6 +595,31 @@ def revision(self, user_ids):
                       .order_by(Backup.created.desc())
                       .first())
 
+    def scores(self, user_ids, only_published=True):
```
Sumukh: Since active assignments don't display scores - is it worth it to add a flag that doesn't include scores? My guess is no - but it would reduce the number of queries we make when there are lots of active assignments.

server/templates/student/course/_assigntable.html
```diff
@@ -1,20 +1,20 @@
-{% macro render_assign(assignments, course, tname="Assignments", empty=False, scores=False) %}
+{% macro render_assign(assignments, course, tname="Assignments", empty=False, show_scores=False) %}
```
Sumukh: This is probably better off as a separate issue: 

There are some courses that keep all of their assignments active (to allow late submissions I would guess). 
Since we are computing the scores anyway should we just show them somewhere - maybe under status? 
It'd be a bit weird if scores are in two different locations (for active vs inactive assignments).  
I can imagine having the score column on the active assignments becoming useful as we have continual autograding.

Sumukh:
image

^ When running a profiler. I think it's due to not calling `.all()` on the scores.

server/models.py
```diff
@@ -586,6 +595,31 @@ def revision(self, user_ids):
...
+        """
+        scores = Score.query.filter(
+            Score.user_id.in_(user_ids),
+            Score.assignment_id == self.id,
+            Score.archived == False,
+        ).order_by(Score.score.desc(), Score.created.desc())
+
+        scores_by_kind = {}
+        # keep only first score for each kind
+        for score in scores:
```
Sumukh: The assignment detail page takes 4 seconds to load locally. The profiler is showing that it's spending a lot of time in the cursor. Calling `scores.all()` here doesn't seem to fix it - so it must be somewhere else.

Sumukh:
It's in this query for me: (revisions). It seems like we could improve that query anyway because we now have submitter id on scores. 

image

I'm not sure why that query would suddenly be so expensive

Sumukh:
Here's a quick implementation of revision that would perform better
```
    def revision(self, user_ids):
        """ Return the revision backup for a user, or None."""
        revision_score = Score.query.filter(
            Score.user_id.in_(user_ids),
            Score.assignment_id == self.id,
            Score.kind == "revision",
            Score.archived == False,
        ).order_by(Score.created.desc()).first()

        if revision_score:
            return revision_score.backup
``` 

I haven't tested it (other than it fixes the performance issue)

PR #948, opened 2016-10-27
Added Initial Code for Editing OAuth Client

### LINKED ISSUE #827: Edit/delete OAuth clients
Right now you can only create them.

dip-kush:
Fixes #827 

I have added the controller and edit_client page. 
I am little new to the Flask. 
I have tried to make necessary change. but the update is not happening in the Database.
@Sumukh Could you please look into my code and provide me some help. 
I will make the necessary changes and proceed with adding the delete functionality.

Thanks

### DIFF
--- a/server/controllers/admin.py
+++ b/server/controllers/admin.py
@@ -856,6 +856,20 @@ def clients():
 
     return render_template('staff/clients.html', clients=clients, form=form)
 
+@admin.route("/clients/<string:client_id>", methods=['GET', 'POST'])
+@is_staff()
+def client(client_id):
+    client = Client.query.get(client_id)
+    client.client_secret = utils.generate_secret_key()
+    form = forms.ClientForm(obj=client)
+    if form.validate_on_submit():
+        form.populate_obj(client)
+        db.session.commit()
+        flash('OAuth client "{}" updated'.format(client.name), "success")
+        return redirect(url_for(".clients"))
+
+    return render_template('staff/edit_client.html', client=client, form=form)
+
 ################
 # Student View #
 ################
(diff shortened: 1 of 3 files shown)


Sumukh:
Cool. Thanks for the PR - I'll take a look later today and leave some comments/suggestions

server/controllers/admin.py
```diff
@@ -859,6 +859,21 @@ def clients():
...
     return render_template('staff/clients.html', clients=clients, form=form)
 
+@admin.route("/clients/<string:client_id>", methods=['GET', 'POST'])
+@is_staff()
+def client(client_id):
+    client = Client.query.get(client_id)
+    client.client_secret = utils.generate_secret_key()
+    form = forms.ClientForm(obj=client)
+    if form.validate_on_submit():
+        client = Client(user=current_user)
```
knrafto: We shouldn't create a new client here, instead just use the one returned from the query above

server/templates/staff/clients.html
```diff
@@ -34,7 +34,7 @@ <h3 class="box-title">OAuth Clients</h3>
                   {% for client in clients %}
                   <tr>
                     <td>{{ client.name }}</td>
-                    <td>{{ client.client_id }}</td>
+                    <td> <a href="{{client.client_id}}">{{ client.client_id }}</a></td>
```
knrafto: This should be a url_for (in case we change the route)

Sumukh:
@dip-kush - There's some comments above that should be addressed before we can merge this

dip-kush:
I was out for a while. I will try to address the review comments soon.

dip-kush:
@Sumukh @knrafto  I have made changes mentioned in the review comments.

PR #949, opened 2016-10-27
Removes Create A Submission Button

### LINKED ISSUE #939: Don't show "Create a new submission" button when assignment is not active
If you enable web uploads for an assignment, an upload button still appears long after the deadline. Clicking it gives you a warning no matter what, so we should just not show the button.

tanaythan:
Removes the create a submission button on
inactive assignments.
Resolves #939


Sumukh:
Thanks!

Issue #866, opened 2016-09-18
Pages should have actual titles.


Sumukh:
Right now every page basically has the same title. 
- Since every page is titled the same, when I hold the back button on Chrome to figure out my path through OK, it's impossible to pick the previous page I want. So, I don't know whether I'm jumping back to a particular assignment or the assignments overview page, for example.

Tasks involved: 
- Look through all of the templates and change the title block. 
- Some tests will break, change the test to match the new page title.

Sumukh:
Also related: Standardize on using `OK` instead of `Ok` in titles.

PR #950, opened 2016-10-28
[WiP] New landing page

kevinlin1:
landing-demo


Sumukh:
Nice! What are the changes you'd like to make? 

There's some copy we need to add. I like your selection of screenshots.

Sumukh:
Current Status: 
2016-10-27-23-21-localhost-5000

Sumukh:
We can add the testimonials back in once we have something to put in there. 

This PR LGTM - feel free to merge it if you think it's ready. We can continue to ship improvements as we go.

knrafto:
I have some small editorial nitpicks. I'm leaving a comment instead of making changes so you have a change to reject my suggestions
  * "CS 61A" and "CS 168" are both two words
  * Is it "OK" or "Ok"? We should capitalize it consistently
  * "Used by many courses" to me seems a bit...off. Since there's a list right after it, "Used by courses with millions of submissions" might be fine
  * Don't capitalize "Computer Science" in the last block
  * "With a language agnostic autograder - you can..." -> "With a language agnostic autograder, you can..." -> 
  * "Give personalized feedback" -> "Personalized feedback" since the rest of the "feature" blocks are noun phrases
    * "composition" is pretty 61A specific. "style" would be a more familiar term

knrafto:
All of our page titles use "Ok"

Sumukh:
We're inconsistent on those titles - but I don't want to scope that into this PR.  I've commented on https://github.com/Cal-CS-61A-Staff/ok/issues/866

I think `OK` should be what we roll with for now since it matches the logotype.

Sumukh:
I'm going to merge this - future improvements can be another PR.

PR #953, opened 2016-10-28
Add submission_time to Backup API

### LINKED ISSUE #952: Add submission time to API
When serializing a Backup object, include the computed `submission_time` property. The autograder will use this instead of the created time.

Sumukh:
Resolves #952 

The autograder needs to be changed to use this property (will happen after this is merged & deployed)

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -203,6 +203,8 @@ class TestAuth(OkTestCase):
         self.assert_200(response)
         backups = response.json['data']['backups']
         self.assertEquals(len(backups), 1)
+        self.assertTrue('submission_time' in backups[0])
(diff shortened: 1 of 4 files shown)


tests/test_api.py
```diff
@@ -203,6 +203,8 @@ def test_export_user(self):
         self.assert_200(response)
         backups = response.json['data']['backups']
         self.assertEquals(len(backups), 1)
+        self.assertTrue('submission_time' in backups[0])
+        self.assertEquals(backups[0]['submission_time'], backups[0]['created'])
```
knrafto: This won't catch bugs where submission_time is always the created time. We should have a test where we change the custom_submission_time attribute

This can probably go in `test_export_final`, which tests submissions

PR #958, opened 2016-11-02
Add Assignments API

Sumukh:


server/controllers/api.py
```diff
@@ -894,9 +922,13 @@ def post(self, user, backup_id):
 api.add_resource(ExportBackup, ASSIGNMENT_BASE + '/export/<string:email>')
 api.add_resource(ExportFinal, ASSIGNMENT_BASE + '/submissions/')
 
+# Course Info
+COURSE_BASE = '/v3/course/<offering:offering>'
+api.add_resource(CourseEnrollment, COURSE_BASE + '/enrollment')
+api.add_resource(CourseAssignment, COURSE_BASE + '/assignments')
```
knrafto: Do we have to update `FORBIDDEN_ASSIGNMENT_NAMES`?

Sumukh: I don't think so - but I could be wrong. I haven't tried using any trick course/assignment names. 

`api` is already in `FORBIDDEN_ROUTE_NAMES`

knrafto:
This is ready to merge, right?

PR #966, opened 2016-11-03
Update title for queue

colinschoen:


colinschoen:
Small fix - self approving.

PR #969, opened 2016-11-06
Fix composition issues that arose during Ants

### LINKED ISSUE #968: Never assign duplicate grading tasks
Basically this means removing the "Ignore submissions that already have a grader" option and making it the default always.

knrafto:
Resolves #968. Also, even fairer queue generation.


Sumukh:
I don't think the test failure is the fault of this branch - (it relates to some stuff I pushed to master) - I'll fix that later. Feel free to merge

PR #980, opened 2016-11-15
bCourses configuration

knrafto:
CRUD for bCourses configuration. I'll leave the actual uploading of scores for another PR. I'd like to have this one merged (and the migration run) so that I can test the upload scores job in staging.


knrafto:
screen shot 2016-11-14 at 8 23 15 pm
screen shot 2016-11-14 at 8 25 13 pm

server/forms.py
```diff
@@ -458,3 +459,37 @@ class GithubSearchRecentForm(BaseForm):
                                 default="Academic Integrity - Please Delete This Repository")
     issue_body = TextAreaField('Issue Body (Optional)', validators=[validators.optional()],
                                description="The strings '{repo}' and '{author}' will be replace with the approriate value")
+
+##########
+# Canvas #
+##########
+
+# e.g. https://bcourses.berkeley.edu/courses/1234567
+CANVAS_COURSE_URL_REGEX = r'^https?://(([a-zA-Z0-9-]+\.)+[a-zA-Z0-9-]+)/courses/(\d+)'
```
Sumukh: I was going to suggest constants.py - but it looks like this is only used in this file - so this is probably the right place for this.

server/controllers/admin.py
```diff
@@ -1164,3 +1166,121 @@ def start_test_job(cid):
...
+        canvas_course=canvas_course,
+        external_names=external_names,
+    )
+
+@admin.route('/course/<int:cid>/canvas/edit/', methods=['GET', 'POST'])
+@is_staff(course_arg='cid')
+def edit_canvas_course(cid):
+    courses, current_course = get_courses(cid)
+    canvas_course = CanvasCourse.by_course_id(cid)
+    if canvas_course:
```
Sumukh: I think this should be `if not canvas_course`

server/templates/staff/sidebar.course.html
```diff
@@ -35,6 +35,11 @@
                     </a>
                   </li>
                   <li>
+                    <a href="{{ url_for('.canvas_course', cid=course.id) }}">
+                      <i class="fa fa-address-book {% if (request.endpoint == 'admin.canvas_course' and request.view_args['cid'] == course.id) %}text-blue{% endif %}"></i> bCourses
```
Sumukh: If you want a better place - a link under course settings might be good 

image

It is kind of hidden - but that should be ok

server/canvas/api.py
```diff
@@ -0,0 +1,60 @@
...
+from server.extensions import cache
+
+def canvas_api_request(canvas_course, method, endpoint, **kwargs):
+    url = 'https://' + canvas_course.api_domain + '/api/v1' + endpoint
+    headers = {
+        'Accept': 'application/json',
+        'Authorization': 'Bearer ' + canvas_course.access_token,
+    }
+    response = requests.request(method, url, headers=headers, **kwargs)
+    response.raise_for_status()
```
Sumukh: Does this mean we'll 500 if the canvas API ever does anything unexpected? It's similar to how we used to (or currently respond) the autograder. 

It's not a big deal - I'm okay only changing this behavior if see lots of errors from the canvas API

server/controllers/admin.py
knrafto: Good catch. The previous version would delete all assignments 😐

Sumukh:
Do you think it makes sense to check if a user's access token even works before letting them save their changes?

server/controllers/admin.py
```diff
@@ -1164,3 +1166,121 @@ def start_test_job(cid):
...
+@admin.route('/course/<int:cid>/canvas/edit/', methods=['GET', 'POST'])
+@is_staff(course_arg='cid')
+def edit_canvas_course(cid):
+    courses, current_course = get_courses(cid)
+    canvas_course = CanvasCourse.by_course_id(cid)
+    if not canvas_course:
+        canvas_course = CanvasCourse(course_id=cid)
+    form = forms.CanvasCourseForm()
+    if form.validate_on_submit():
+        form.populate_canvas_course(canvas_course)
```
Sumukh: I would like to avoid leaking a user's personal access token. If a professor puts their access token - any staff who has that info - has professor level access to _other_ courses. 

Instead of populating the object - could we just set the value of the URL and leave the access token blank

server/canvas/api.py
knrafto: I've caught the errors in the controller so we can flash a message (just in case)

knrafto:
I've addressed your comments. Do you want to take another look?

Sumukh:
I took a look - seems fine - feel free to merge/migrate

PR #981, opened 2016-11-15
Show latest backup on student overview page

Sumukh:
This adds a column for the latest backup to the staff view. 

Performance: also means that we are doing an extra query on the student side - but I didn't think it was worth it to only add this for staff (based on the `is_staff` argument) 

Usage: Easier to link from Office Hours app to this page


knrafto:
LGTM

PR #986, opened 2016-11-23
Write job to upload scores to bCourses

knrafto:
Example output: http://ok-prod.cs61a.org/admin/course/4/jobs/76/

### DIFF
--- a/server/canvas/api.py
+++ b/server/canvas/api.py
@@ -43,6 +43,7 @@ def get_assignments(canvas_course):
         '/courses/{}/assignments'.format(canvas_course.external_id),
     )
 
+@cache.memoize(300)
 def get_students(canvas_course):
     # https://bcourses.berkeley.edu/doc/api/courses.html#method.courses.users
     return canvas_api_get_list(
--- a/server/canvas/api.py
+++ b/server/canvas/api.py
@@ -12,14 +12,22 @@ def canvas_api_request(canvas_course, method, endpoint, **kwargs):
     response.raise_for_status()
     return response.json()
 
-def score_endpoint(canvas_assignment, enrollment):
-    # We can use the student ID directly here. See
-    # https://bcourses.berkeley.edu/doc/api/file.object_ids.html
-    return '/courses/{}/assignments/{}/submissions/sis_user_id:{}'.format(
-        canvas_assignment.canvas_course.external_id,
-        canvas_assignment.external_id,
-        enrollment.sid,
-    )
+def canvas_api_get_list(canvas_course, endpoint, **kwargs):
+    url = 'https://' + canvas_course.api_domain + '/api/v1' + endpoint
+    headers = {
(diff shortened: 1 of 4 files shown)


server/canvas/api.py
```diff
@@ -12,14 +12,22 @@ def canvas_api_request(canvas_course, method, endpoint, **kwargs):
...
+    }
+    params = kwargs.pop('params', {})
+    params['per_page'] = 100
+    response = requests.get(url, headers=headers, params=params, **kwargs)
+    response.raise_for_status()
+    results = response.json()
+    while 'next' in response.links:
+        response = requests.get(response.links['next']['url'], headers=headers)
+        response.raise_for_status()
+        results.extend(response.json())
```
Sumukh: Have you ever hit the API rate limit while doing this? 

We haven't hit the rate limit in the past - so possibly not worth worrying about - but if we do have to - we can use the response headers to figure out how many requests we have left.

server/canvas/api.py
```diff
@@ -30,15 +38,31 @@ def get_course(canvas_course):
...
 def get_assignments(canvas_course):
-    return canvas_api_request(
+    return canvas_api_get_list(
         canvas_course,
-        'GET',
-        '/courses/{}/assignments?per_page=100'.format(canvas_course.external_id),
+        '/courses/{}/assignments'.format(canvas_course.external_id),
+    )
+
+def get_students(canvas_course):
```
Sumukh: Sidenote: Could this enable automatic/recurring enrolling of students?

server/canvas/upload.py
```diff
@@ -1,16 +1,64 @@
...
+        else:
+            max_score = max(scores, key=lambda score: score.score)
+            new_score = max_score.score
+            backup_id = encode_id(max_score.backup_id)
+            if old_score != new_score:
+                new_scores[canvas_user_id] = new_score
+                stats['updated'] += 1
+            else:
+                stats['not_changed'] += 1
+        logger.info(row_format.format(sid, emails, backup_id, old_score, new_score))
```
Sumukh: If it's a group - is it ok to just print one SID? 

Also - why we need to print SIDs? It's somewhat sensitive information - so the less we display it the better. If it makes it much easier to debug in bCourses - then we can keep it

server/canvas/api.py
cycomachead: It heavily depends on the endpoint. If you're just iterating over students or assignments you should almost never the limit.

server/canvas/upload.py
knrafto: (Re: groups) We're iterating over students, not groups. So if we have a group of two students, both students will be printed separately and both students will get a score (it should be the same score, though).

server/canvas/api.py
knrafto: We can cache getting all students for a bit (since every upload has to do this).

PR #988, opened 2016-11-23
fixes #885: renders comments below composition score

### LINKED ISSUE #885: Display comments below score so that it's easier to see comments
List comments (render markdown) and the line number/author below the score (so students don't miss comments)

kvakil:
This commit displays all comments written by the grader below the
composition score. Clicking on a comment brings the student to the
line where the comment was written. The template logic is handled in
student/assignment/list_comments.html. Changes were also made to
diff.html in order to make the comments linkable with href anchors.

I don't particularly like adding the second argument to
render_comment in diff.html, but its necessary in order to make sure
that each ID occurs only once (because I used render_comment) to
render the comments twice.

Also a lot of the logic is handled in the list_comments template,
and some of that might be better rewritten in Python. I think that
its simple enough to not be a problem right now.

Here is an example of how it looks. I'm open to changing the exact
design.

example

### DIFF
--- a/server/templates/diff.html
+++ b/server/templates/diff.html
@@ -3,9 +3,9 @@
   <td class="line-number" data-line-number="{{ '' if i is none else i }}"></td>
 {% endmacro %}
 
-{% macro render_comment(comment, linkable_id=True) %}
+{% macro render_comment(comment) %}
   <div class="comment"
-    {% if linkable_id %} id="{{ utils.encode_id(comment.id) }}" {% endif %} 
+    id="{{ utils.encode_id(comment.id) }}" 
     data-id="{{ utils.encode_id(comment.id) }}"
     data-message="{{ comment.message }}"
     data-no-instant>
--- a/server/templates/diff.html
+++ b/server/templates/diff.html
@@ -3,8 +3,9 @@
   <td class="line-number" data-line-number="{{ '' if i is none else i }}"></td>
 {% endmacro %}
 
-{% macro render_comment(comment) %}
+{% macro render_comment(comment, linkable_id=True) %}
   <div class="comment"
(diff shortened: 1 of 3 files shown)


Sumukh:
Cool! 

Thanks I think we should change the design a bit - we don't need to render full comment blocks for each one - we really just want a summary right? 

If it's just a list of comments (or a table or something) - it should be fine. 
```
fizzbuzz.py: Line 30 - comment text  from ____@____ 
fizzbazz.py: Line 12 - comment from ____@_____
```

kvakil:
Okay, how does this look?

dr1

Right now I only show comments which were written by whoever grades the
composition section. Do you want me to change that or still include the
grader's email again?

Sumukh:
Limiting it to comments from the author seems reasonable for now  (I think in the future we'll want all the comments)

I haven't looked at the code - I'll take a look later today

cc: @kevinlin1

kevinlin1:
Nice! I can definitely see this being useful for students to treat as a checklist before submitting a composition revision.

Is it necessary to explicitly label this as "Grader's comments"? Perhaps there's a way to restyle it so that it's clear from context what the comments are referring to. Maybe if we play around with the style of the box and make

kvakil:
I changed it to use a table instead, which I think makes it look
way better. I also made it so that the current (hovered) row is
highlighted. Does it look better now?

table

knrafto:
Nice! Since it's no longer "grader's comments" I think we should show all the comments. That way they'll still show up if another staff member edits the composition score.

kvakil:
Alright I changed it to show all comments and also changed it to show
the commenter.

commenter

knrafto:
Looks good! Merging and deploying. Thanks a lot!

PR #990, opened 2016-11-24
Upload all the scores to bCourses

knrafto:
* Add a button to start a bunch of jobs at once. Right now there's only two workers, so we're still probably not going to hit bCourses API limits.
* bCourses jobs would previously fail if no scores changed. Now they don't fail.


knrafto:
This is pretty uncontroversial and I've already used it a couple times, so merging

PR #999, opened 2016-12-04
Subtract Templates for MOSS

Sumukh:
Since MOSS doesn't seem to be handling our templates well - we should try removing templates manually.

### DIFF
--- a/server/forms.py
+++ b/server/forms.py
@@ -431,7 +431,6 @@ class PublishScores(BaseForm):
         choices=[(kind, kind.title()) for kind in SCORE_KINDS],
     )
 
-
 ########
 # Jobs #
 ########
@@ -446,6 +445,8 @@ class MossSubmissionForm(BaseForm):
     file_regex = StringField('Regex for submitted files', default='.*',
                              validators=[validators.required()])
     language = SelectField('Language', choices=[(pl, pl) for pl in COMMON_LANGUAGES])
+    subtract_template = BooleanField('Subtract Template', default=False,
+                                     description="Only send the changes from the template to MOSS")
 
 class GithubSearchRecentForm(BaseForm):
     access_token = StringField('Github Access Token',
--- a/server/jobs/moss.py
+++ b/server/jobs/moss.py
@@ -4,13 +4,15 @@ import shlex
 import glob
 import tempfile
 import re
+import difflib
 
 from server.models import Assignment, Backup, db
 from server.utils import encode_id
 from server import jobs
 
 @jobs.background_job
-def submit_to_moss(moss_id=None, file_regex=".*", assignment_id=None, language=None):
+def submit_to_moss(moss_id=None, file_regex=".*", assignment_id=None, language=None,
+                   subtract_template=False):
     logger = jobs.get_job_logger()
     logger.info('Starting MOSS Export...')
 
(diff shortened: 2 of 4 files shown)


server/jobs/moss.py
```diff
@@ -72,6 +88,13 @@ def submit_to_moss(moss_id=None, file_regex=".*", assignment_id=None, language=N
             for file in contents:
                 if file == 'submit':  # ignore fake file from ok-client
                     continue
+                if subtract_template and file in assign.files:
+                    # Compare to template and only include lines that new
+                    template, source = assign.files[file], contents[file]
+                    d = difflib.Differ()
```
Sumukh: I think we should make difflib ignore spacing and new lines (Extra indentation, Tabs v Spaces ...)  - we can do that with the two args to `Differ` (Junk Char being one of them)

server/forms.py
```diff
@@ -446,6 +445,8 @@ class MossSubmissionForm(BaseForm):
     file_regex = StringField('Regex for submitted files', default='.*',
                              validators=[validators.required()])
     language = SelectField('Language', choices=[(pl, pl) for pl in COMMON_LANGUAGES])
+    subtract_template = BooleanField('Subtract Template', default=False,
```
knrafto: Should the default be True?

server/jobs/moss.py
```diff
@@ -57,6 +59,20 @@ def submit_to_moss(moss_id=None, file_regex=".*", assignment_id=None, language=N
         match_pattern = re.compile(file_regex)
         ignored_files = set()
 
+        template_files = []
+        for template in assign.files:
+            dest = "{}/{}".format(tmp_dir, template)
```
knrafto: Use `os.path.join`

server/jobs/moss.py
```diff
@@ -57,6 +59,20 @@ def submit_to_moss(moss_id=None, file_regex=".*", assignment_id=None, language=N
...
+            with open(dest, 'w') as f:
+                f.write(assign.files[template])
+            template_files.append(template)
```
knrafto: template_files is just list(assign.files). You could pass assign.files to the ''.join below directly

server/jobs/moss.py
```diff
@@ -72,6 +88,13 @@ def submit_to_moss(moss_id=None, file_regex=".*", assignment_id=None, language=N
             for file in contents:
                 if file == 'submit':  # ignore fake file from ok-client
                     continue
+                if subtract_template and file in assign.files:
+                    # Compare to template and only include lines that new
+                    template, source = assign.files[file], contents[file]
+                    d = difflib.Differ()
+                    diff = d.compare(template.splitlines(True), source.splitlines(True))
```
knrafto: splitlines(keepends=True) instead of splitlines(True) is clearer

knrafto:
I'm assuming you've tested it already, so besides a few style nits I think it's good to go

server/forms.py
Sumukh: I think we should leave it off by default (from my testing) - It just seems really good at catching last year's template. I have to evaluate it some more.

server/jobs/moss.py
```diff
@@ -57,21 +59,44 @@ def submit_to_moss(moss_id=None, file_regex=".*", assignment_id=None, language=N
...
+
         for backup in backup_query:
             # Write file into file
             file_contents = [m for m in backup.messages if m.kind == 'file_contents']
             if not file_contents:
                 logger.info("{} didn't have any file contents".format(backup.hashid))
                 continue
             contents = file_contents[0].contents
-            dest_dir = "{}/{}/".format(tmp_dir, backup.hashid)
+            dest_dir = "{}/".format(os.path.join(tmp_dir, backup.hashid))
```
knrafto: I think this `format` is unnecessary? Maybe you need `os.path.isdir`

PR #1000, opened 2016-12-04
CSV with highest score per Score Kind

Sumukh:
This needs work - this is just a hacky prototype for a course request. (also there are some commits I need to get out of this branch) 

I'd like to see this be the default download and the current one (which gets all scores) become the secondary one 

We might want to keep the same endpoint but we'll just use request args 

`..../scores` -> One with the highest score per person per score kind 
`..../scores?all=1` -> All Scores


Sumukh:
No other course has asked for this - so I guess it's not worth it.

PR #1011, opened 2016-12-21
Split models into a module

Sumukh:
I'm a bit unhappy with how this turned out for three reasons: 

- A giant `__init__.py` file with a bunch of imports so that the current app can continue to function. 
- Needed a ModelProxy to handle circular imports (or have to import inside of the function) 
- I had to create a model folder inside of models for some reason (It does make the separation nice but I'm don't like the current naming schema) 

Suggestions on how to fix those flaws (or a better way to do the split) would be welcomed. 

With the current structure we can also further modularize to have: `server/models/canvas/CanvasAssignment.py` - Is this worth it?


server/models/__init__.py
```diff
@@ -0,0 +1,24 @@
+import server.models.model
+from server.models.db import db
+from server.models.proxy import ModelProxy
+
+# The imports below are order sensitive
+from server.models.model.GradingTask import GradingTask
```
Sumukh: Not only is this somewhat order sensitive, it doesn't look nice and we'll have to remember to add a line in here when creating new modules. 

There is probably an automated to do this - it would be hacky but I think it'd be better than this ...

server/models/proxy.py
```diff
@@ -0,0 +1,6 @@
+class ModelProxy(object):
+    def __getattribute__(self, key):
+        import server.models.model
```
Sumukh: I don't think this is actually needed here and can go to the top of the files.

server/models/proxy.py
```diff
@@ -0,0 +1,6 @@
+class ModelProxy(object):
```
Sumukh: The return of the `ModelProxy`! 😨 

https://github.com/Cal-CS-61A-Staff/ok/blob/37b27d779129d2e1b63000dbaf4496f2e45fe1d5/server/app/utils.py#L36

knrafto:
Perhaps `__init__.py` can use [`__all__`](https://docs.python.org/3/tutorial/modules.html#importing-from-a-package)? I'm not too sure how it works, but my gut feeling is that it might be useful here.

Why exactly we need another `model` directory?

server/models/model/GradingTask.py
```diff
@@ -0,0 +1,108 @@
...
+        backups = [b for b in backups if not cls.query.filter_by(backup_id=b).count()]
+
+        paritions = chunks(list(backups), len(staff))
+        tasks = []
+        for assigned_backups, grader in zip(paritions, staff):
+            for backup_id in assigned_backups:
+                task = cls(kind=kind, backup_id=backup_id, course_id=course_id,
+                           assignment_id=assignment_id, grader=grader)
+                tasks.append(task)
+                cache.delete_memoized(ModelProxy.User.num_grading_tasks, grader)
```
knrafto: `num_grading_tasks` could be a static method on `GradingTask`

server/models/model/User.py
```diff
@@ -0,0 +1,62 @@
...
+    email = db.Column(db.String(255), unique=True, nullable=False, index=True)
+    is_admin = db.Column(db.Boolean(), default=False)
+
+    export_items = ('email', 'name')
+
+    def __repr__(self):
+        return '<User {0}>'.format(self.email)
+
+    def enrollments(self, roles=None):
+        Enrollment = ModelProxy.Enrollment
```
knrafto: Same: make `enrollments` (and maybe `is_enrolled`) static methods? Or do a `filter` on `self.participations` like `is_enrolled` currently does

Sumukh:
Don't think it's worth the work right now.

PR #1013, opened 2016-12-23
Score Dashboard & Audit (WIP)

### LINKED ISSUE #1007: Score Audit Log
Having a score audit log for all assignments (and on an assignment by assignment basis) can be useful. 

- Audit log to review changes that instructors are making
- Quick way to glance at the distribution of scores.

Sumukh:
Score Dashboard:
image

Example of how this dashboard is helpful: I was able to identify that some graders made manual adjustments to give students 2.0 when this homework should only be out of one point. 

Example of what the audit job outputs (with fake data):
image

Work to be done: 
- Change the graphs: They don't really capture what I want to see as a TA
- (Maybe not part of this PR) Once we have an efficient scores query: 
   - Scores: Just grab scores for final submissions - The existing graph just plots all of the non-archived scores
   - Display Submitter information in table. 

Resolves #1007


knrafto:
I'm not really sure how to interpret the graphs. Most useful would be a histogram

knrafto:
It would also be useful if the job printed out links to student pages (e.g. one per line, each like `https://okpy.org/cal/cs61a/fa16/agrothendieck@berkeley.edu`)

knrafto:
This looks great though, thanks!

Sumukh:
It seems useful to have this even if the graphs aren't quite right (a histogram would be much better). 

I'm leaning towards just deploying this as is.

Sumukh:
Closing to make a PR with a cleaner set of commits.

PR #1017, opened 2016-12-27
Store & Display Job results

### LINKED ISSUE #1014: Jobs should store return value
The return value can provide a summary or something more useful than the log. 

MOSS: Returns a one line string for the MOSS url (instead of having to scroll through the log to find it) 
(Future) Export Jobs: Returns a link to the file url 
bCourses Jobs: Returns a summary of changes ("15 grades changed") 

Bonus: It would be great if our jobs could output HTML so we could render tables etc. 

Proposed solution. 

Add a parameter to the job field (and the job schema) to store what kind of output we are expecting. 
Add a MEDIUMTEXT field (16 MiB) to the job schema store the output of the job. 

Render the return value above the log. 

```
enqueue_job(..., result_format='link') 
enqueue_job(..., result_format='string')
enqueue_job(..., result_format='html') 
```

Sumukh:
Job Table: 
image

Within the job: 
image

Resolves #1014


Sumukh:
It's worth noting that this requires us to trust the output of jobs that render HTML because we render it inline (otherwise there's a stored XSS vulnerability) - but since we have to review all of the jobs that are created it should be fine. 

We might be able to use javascript if we _need_ to have unsafe content in there

Sumukh:
I'm going to wait to merge this until the migration has been run (and might try to bundle a bunch of migrations if we have any coming)

PR #1018, opened 2016-12-27
Edit OAuth Client without changing secret

### LINKED ISSUE #973: Edit OAuth Client - Optionally update secret key
Use case: I just want to add a domain to an OAuth client - I don't want to have to change the secret key inside of my application. 

Right now - we will overwrite the secret key if an edit is made. 

Maybe we can use a child class of the original form with an additional checkbox field - change secret key.

Sumukh:
edit-oauth

Resolves #973

### DIFF
--- a/server/controllers/admin.py
+++ b/server/controllers/admin.py
@@ -863,14 +863,19 @@ def clients():
 @is_staff()
 def client(client_id):
     client = Client.query.get(client_id)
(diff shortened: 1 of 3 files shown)


knrafto:
Seems fine. I find having to copy the secret from the flashed message a little annoying (rather than a separate form), but I don't think it'll be used often.

Issue #1020, opened 2016-12-27
OAuth page takes up too much space


knrafto:
Here's how it looks on my 13-inch laptop:
<img width="1280" alt="screen shot 2016-12-27 at 12 04 22 am" src="https://cloud.githubusercontent.com/assets/2042019/21494692/127f844a-cbc8-11e6-969f-8b77fe8fa59b.png">

I have to scroll to see the "confirm" button. It should easily fit on one page so I don't have to scroll. In particular, there's a lot of empty space at the top (probably because some of the CSS is shared with the landing page CSS).

Sumukh:
I tried removing some of it with  https://github.com/Cal-CS-61A-Staff/ok/commit/2e8d415a2dc6e5b713aee41b898fdb20524faf4d - but I did not go far enough because I didn't make a new base template. 

It'll be much easier once there is a new base template for the OAuth view.

PR #1023, opened 2016-12-27
Out-of-band OAuth codes

### LINKED ISSUE #878: Browserless Auth for Ok Client
We should generate access tokens by generating a URL that students load in their browser. 

This will be useful for the Ok Client in iPython Notebooks.

knrafto:
oauth-oob

Resolves #878.

You can try it out by running `documentation/sample_out_of_band_oauth.py`. The code page still needs some CSS tweaks. This is a good opportunity to create a new OAuth base template and tackle #1020.

### DIFF
--- a/server/controllers/oauth.py
+++ b/server/controllers/oauth.py
@@ -96,9 +96,9 @@ def intercept_out_of_band_redirect(f):
     return wrapper
 
 @oauth.route('/oauth/authorize', methods=['GET', 'POST'])
+@login_required
 @intercept_out_of_band_redirect
 @oauth_provider.authorize_handler
-@login_required
 def authorize(*args, **kwargs):
     # Only CSRF protect this route.
     csrf_check()
--- a/server/controllers/oauth.py
+++ b/server/controllers/oauth.py
@@ -138,7 +138,10 @@ def revoke_token():
 
 @oauth.route('/oauth/errors')
 def oauth_errors():
-    error = request.args.get('error', 'Unknown Error')
-    description = request.args.get('error_description', 'No details available')
+    error = request.args.get('error')
+    if error:
+        # 'access_denied' -> 'Access Denied'
+        error = error.replace('_', ' ').title()
+    description = request.args.get('error_description')
     return render_template('errors/generic.html',
                            error=error, description=description), 400
--- a/server/controllers/oauth.py
+++ b/server/controllers/oauth.py
@@ -1,11 +1,14 @@
 import datetime as dt
+import functools
 import logging
+import urllib.parse
 
 from flask import (Blueprint, flash, redirect, render_template,
-                   request, session, url_for, jsonify)
+                   request, session, url_for, jsonify, make_response)
 
 from flask_login import current_user, login_required, logout_user
 
+from server.constants import OAUTH_OUT_OF_BAND_URI
 from server.models import db, Client, Token, Grant
 from server.extensions import csrf, oauth_provider
 from server.controllers.auth import csrf_check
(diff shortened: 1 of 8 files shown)


knrafto:
This is ready for review now.

ok-client will give instructions on how to copy-paste into the terminal, so we don't need to do that here.

I did some CSS work and decided to fix #1020 as well. Here's how it looks:

screen shot 2016-12-29 at 5 22 29 pm
screen shot 2016-12-29 at 5 22 35 pm
screen shot 2016-12-29 at 5 22 43 pm

Sumukh:
Thanks! Something about the desktop view is throwing me off a bit - I'll see if I can play around with it a bit. 

Re: Clipboard code - do you think it makes sense to use: https://github.com/zenorocha/clipboard.js

It'll get rid of the jQuery code we have right now and make it as easy as
```
<!-- Target -->
<input id="foo" value="https://github.com/zenorocha/clipboard.js.git">

<!-- Trigger -->
<button class="btn" data-clipboard-target="#foo">
    <img src="assets/clippy.svg" alt="Copy to clipboard">
</button>
```

knrafto:
Not enough space at the top maybe? I gutted it maybe a little too much.

I did see clipboard.js, but decided it was overkill. We don't really need another library for 10 lines of JS used on one page.

server/models.py
```diff
@@ -1503,6 +1503,8 @@ def default_redirect_uri(self):
         return self.redirect_uris[0]
 
     def validate_redirect_uri(self, redirect_uri):
+        if redirect_uri == OAUTH_OUT_OF_BAND_URI:
```
Sumukh: I get why we want loopback addresses to automatically be allowed - but why not make users add the URI manually if they want it? 

(I'm not really opposed to it - just want to see if there's some special reasoning for it)

server/controllers/oauth.py
```diff
@@ -75,7 +78,25 @@ def save_token(token, orequest, *args, **kwargs):
     db.session.commit()
     return tok
 
+def intercept_out_of_band_redirect(f):
+    @functools.wraps(f)
+    def wrapper(*args, **kwargs):
+        response = make_response(f(*args, **kwargs))
+        if response.status_code == 302:
```
knrafto: Yes, except it should never return a 400 (always a 200 or 302). If an error occurs, it'll 302 with an error instead of a code.

server/controllers/oauth.py
```diff
@@ -91,6 +112,13 @@ def authorize(*args, **kwargs):
     confirm = request.form.get('confirm', 'no')
     return confirm == 'yes'
 
+@oauth.route('/oauth/code')
+def oauth_code():
```
knrafto: This page doesn't access any user information (it just displays whatever is in the query params), so I don't see why logging in is required. You have to log in to get the code anyway.

The reason I made it a standard route is because of https://en.wikipedia.org/wiki/Post/Redirect/Get: If the user refreshes the code page nothing bad will happen, whereas if we return the response directly in the oob decorator they'll get a "confirm form resubmission" popup on refresh.

server/models.py
knrafto: No reason, other than convenience and I didn't think it was a security risk.

knrafto:
Okay, made a few changes. I still think that the page that displays the code should be a separate route, and that logging in should not be required

server/controllers/oauth.py
Sumukh: For what it's worth - Google's code view is a POST.  Refreshes trigger a resubmission warning and generate a new code. 

You can try it out with
> gcloud auth login --no-launch-browser

knrafto: I don't really see the benefit of that - the user should go through the entire flow again to get a new code

Issue #941, opened 2016-10-25
Binary file uploads


knrafto:
Right now our backend doesn't have a good way to store binary files. We have a JSON dict mapping filenames to file contents, where the contents is a JSON (Unicode) string.

One option is store binary file contents as base64-encoded into a string, but we don't have a way of knowing if the original file was binary or not. Perhaps we can determine the binaryness based on the filename.

knrafto:
Related: if we upload a binary file accidentally, we'll get an internal server error. We should catch the error instead.

Sumukh:
What if we keep the files in a cloud storage provider? It'd be an order of magnitude cheaper than storing it in our DB (17 a gig on the DB vs ~2c on DB) and I think it's bad practice to store largish binary files in DBs anyway. 

We should do #918 first.

Sumukh:
A new course has requested this feature for their assignments.

Sumukh:
Resolved in #1024

PR #1024, opened 2016-12-31
Support for creation & download of Binary Files

### LINKED ISSUE #918: Be able to export/store files
Some layer of persistent storage for large files would be useful. 

Use cases:
- Download MOSS urls and serve them locally. 
- Serve exports of student code (instead of using scripts)
- Jobs could store files and provide the user a download. 

I think the only viable way to do this is to use the cloud storage system (gcloud storage buckets or S3). 

Ideally we would not be too dependent on GCloud specific features for storage. 

Current approach: 
- Create a new table to store exported file metadata: 
  URL to access (or Gcloud ID), Gcloud bucket, Course (nullable), Assignment (nullable), Creator, Filename, Time Created

Another Idea:
- What if exported files were not persistent? Could we store them in a temporary KV store? 
  (not a fan of this - but I thought I'd put it out there)

Sumukh:
This PR resolves #918 and unblocks #941 

I found a few bugs in the libraries being used so there's now a [fork](https://github.com/okpy/flask-cloudy) on the okpy github. 

Sidenote: I'm not a huge fan of the library since I've had to fight it a few times - but it does provide the ability to easily switch providers (Local Storage, S3, Google Cloud Storage) and works. We had a version of this in v2 and using the official cloud storage library was also painful in its own way - but that might have been App Engine specific. 

I've tested it locally with Google Cloud Storage but haven't tested in staging because of the migration I'd need to run. 

Future: 
Once this is merged - I'll file a new issue on this repo to figure out how we can use service accounts instead of having to use the S3 interop keys that Google provides. Service Accounts allow for more granular control of access - and we might just be able to use the service account that Google Container Engine provides to instances.


server/jobs/example.py
```diff
@@ -1,13 +1,35 @@
...
 @jobs.background_job
-def test_job(duration=0, should_fail=False):
+def test_job(duration=0, should_fail=False, make_file=False):
     logger = jobs.get_job_logger()
 
     logger.info('Starting...')
     time.sleep(duration)
     if should_fail:
         1/0
+    if make_file:
```
Sumukh: I kind of stuck this in here so that I could test file creation in production. I can remove it once there is another place to put an example of how to use file upload.

server/models.py
```diff
@@ -1639,3 +1639,71 @@ class CanvasAssignment(Model):
...
+        db.Integer, db.ForeignKey('assignment.id'), index=True
+    )
+    assignment = db.relationship('Assignment')
+
+    @property
+    def object(self):
+        with storage.use(self.container) as container:
+            return container.get(self.object_name)
+
+    def delete(self):
```
Sumukh: It's possible that we'll have a file dashboard and a way to delete files so I wanted to provide support now (to avoid another migration)

server/utils.py
```diff
@@ -253,3 +254,23 @@ def generate_number_table(num):
     Used in models.Assignment.mysql_course_submissions_query
     """
     return ' UNION '.join('SELECT {} as pos'.format(i) for i in range(1, num + 1))
+
+def upload_file(file, container=None, name=None, prefix=None, overwrite=False):
+    """ A utility function to get around having to specify
+    what file extensions are allowed.
+    """
+    extension = get_file_extension(file)
```
Sumukh: I didn't want to hardcode all of the files we will accept and have to redeploy every time we change the filetype so I added this code in here. 

I could specify it as a giant list in the constants or settings - but I think this is better.

server/models.py
```diff
@@ -1643,3 +1643,71 @@ class CanvasAssignment(Model):
...
+#########
+# Files #
+#########
+
+class ExternalFile(Model):
+
+    id = db.Column(db.Integer, primary_key=True)
+    # Bucket/Folder that the file is stored in
+    container = db.Column(db.String(255), nullable=False)
+    filename = db.Column(db.String(1024), nullable=False)
```
Sumukh: Why `filename` over `file_name`? 

- http://english.stackexchange.com/a/5374/10058 
- The real reason: Python seems to go with `filename` when possible

server/controllers/files.py
```diff
@@ -0,0 +1,57 @@
...
+files = Blueprint('files', __name__)
+
+logger = logging.getLogger(__name__)
+
+@files.route("/files/<hashid:file_id>")
+@login_required
+def file_url(file_id):
+    ext_file = ExternalFile.query.filter_by(id=file_id, deleted=False).first()
+    if not ext_file or not ExternalFile.can(ext_file, current_user, 'download'):
+        return abort(404)
```
knrafto: `abort` raises an exception, so there's nothing to return (same with below)

server/controllers/files.py
```diff
@@ -0,0 +1,57 @@
...
+    try:
+        storage_obj = ext_file.object
```
knrafto: I find it extremely surprising that accessing a property will raise an exception. I think a method would be better

server/controllers/files.py
```diff
@@ -0,0 +1,57 @@
...
+    except libcloud.common.types.InvalidCredsError as e:
+        logger.warning("Could not get file {0} - {1}".format(file_id, ext_file.filename),
+                       exc_info=True)
+        storage_obj = None
```
knrafto: This line has no effect

server/controllers/files.py
```diff
@@ -0,0 +1,57 @@
...
+        raise e
```
knrafto: Could be simply `raise`

server/controllers/files.py
```diff
@@ -0,0 +1,57 @@
...
+
+    # Do not use .download_url for local storage.
+    if "local" in storage.driver.name.lower():
+        return redirect(url_for('.send_file', file_id=ext_file.id))
+    else:
+        return redirect(storage_obj.download_url())
+
+@files.route("/files/<hashid:file_id>/download")
+@login_required
+def send_file(file_id):
```
knrafto: I don't really see why this is necessary. We shouldn't be downloading files to the web server machine in production. And to serve local files, we can just send the file directly instead of copying it somewhere else.

server/utils.py
```diff
@@ -253,3 +254,23 @@ def generate_number_table(num):
     Used in models.Assignment.mysql_course_submissions_query
     """
     return ' UNION '.join('SELECT {} as pos'.format(i) for i in range(1, num + 1))
+
+def upload_file(file, container=None, name=None, prefix=None, overwrite=False):
```
knrafto: This method should probably create an `ExternalFile` object too to track the file, instead of having to remember to do that at each call site

server/utils.py
knrafto: So, essentially, we're accepting any file extension? That's fine with me, but why bother with `allowed_extensions` at all? Does flask-cloudy require it?

Sumukh: Yeah - it has a default list of extensions and if the file extension you are uploading isn't in that list it'll raise an exception.

server/controllers/files.py
knrafto: Actually, wouldn't this 500? 404ing seems better

Sumukh: Yeah - I should remove this - it was left in for debugging.

server/controllers/files.py
Sumukh: I don't think this is true anymore (the raise below it has been removed). (Though this line could be moved up to line 20)

server/models.py
```diff
@@ -1643,3 +1644,93 @@ class CanvasAssignment(Model):
...
+            return 'application/octet-stream'
+        return guess[0]
+
+    def delete(self):
+        self.object.delete()
+        self.deleted = True
+        db.session.commit()
+
+    @staticmethod
+    def upload(iterable, user_id, name, is_staff=True, course_id=None,
```
Sumukh: I'm going to move this over into Storage. I initially put it here to avoid circular dependencies - but that never happened - it makes more sense in `storage.py`

server/settings/prod.py
```diff
@@ -49,6 +49,12 @@
 REDIS_PORT = 6379
 RQ_POLL_INTERVAL = 2000
 
+STORAGE_PROVIDER = 'GOOGLE_STORAGE'
+STORAGE_SERVER = False
+STORAGE_CONTAINER = 'ok-v3-user-files'
```
Sumukh: Should be an env var.

server/controllers/files.py
```diff
@@ -0,0 +1,45 @@
...
+        storage_obj = None
+
+    if storage_obj is None:
+        abort(404, "File does not exist")
+
+    # Do not use .download_url for local storage.
+    if "local" in storage.driver.name.lower():
+        def streamer(stream):
+            for data in stream:
+                yield data
```
knrafto: Isn't this function a noop?

Sumukh: In the sense that it'll never be executed in production? I could see some one running their own deployment on heroku or their own server and then this would be useful.

knrafto: No, I mean it's the identity function. It takes an iterable and returns an iterable with the same elements

server/models.py
Sumukh: Nevermind - there is a dependency. 

forms.py import canvas_api which imports extensions which imports Storage which (after the change) imports models and models imports storage (to be able to get blobs)

server/models.py
```diff
@@ -1643,3 +1644,93 @@ class CanvasAssignment(Model):
...
+               assignment_id=None, **kwargs):
+        object = storage.upload(iterable, name=name, **kwargs)
+        external_file = ExternalFile(
+            container=storage.container_name,
+            filename=name,
+            object_name=object.name,
+            course_id=course_id,
+            assignment_id=assignment_id,
+            user_id=user_id,
+            is_staff=False)
```
knrafto: Should be `is_staff`

server/models.py
```diff
@@ -1643,3 +1644,93 @@ class CanvasAssignment(Model):
...
+class ExternalFile(Model):
+
+    id = db.Column(db.Integer, primary_key=True)
+    # Bucket/Folder that the file is stored in
+    container = db.Column(db.String(255), nullable=False)
+    filename = db.Column(db.String(1024), nullable=False)
+    object_name = db.Column(db.String(1024), nullable=False)
+
+    is_staff = db.Column(db.Boolean, nullable=False)
+    deleted = db.Column(db.Boolean, default=False)
```
knrafto: `nullable=False`

server/models.py
```diff
@@ -1643,3 +1644,93 @@ class CanvasAssignment(Model):
...
+
+class ExternalFile(Model):
+
+    id = db.Column(db.Integer, primary_key=True)
+    # Bucket/Folder that the file is stored in
+    container = db.Column(db.String(255), nullable=False)
+    filename = db.Column(db.String(1024), nullable=False)
+    object_name = db.Column(db.String(1024), nullable=False)
+
+    is_staff = db.Column(db.Boolean, nullable=False)
```
knrafto: I don't really like the name `is_staff`, it kinda makes it sound like the file is a staff member. Actually, is this column needed at all? Students can only see their partner's files, so the only time it would make a difference is if a student is partnered with a staff member.

server/jobs/example.py
```diff
@@ -1,14 +1,27 @@
...
+        data = io.BytesIO(b'print("Hello World")\n'*duration)
+        upload = ExternalFile.upload(data, user_id=1, course_id=1,
+                                     name='temp.py', prefix='jobs/example/')
+        logger.info("Saved as: {}".format(upload.object_name))
+        logger.info('File ID: {0}'.format(encode_id(upload.id)))
+        msg = ("Waited for <a href='/files/{0}'> {1} seconds </a>"
```
knrafto: `url_for`? I know it's a test job, but it'll be used as an example for other jobs so we should get in the habit.

server/jobs/example.py
```diff
@@ -1,14 +1,27 @@
...
-def test_job(duration=0, should_fail=False):
+def test_job(duration=0, should_fail=False, make_file=False):
     logger = jobs.get_job_logger()
 
     logger.info('Starting...')
     time.sleep(duration)
     if should_fail:
         1/0
+    if make_file:
+        data = io.BytesIO(b'print("Hello World")\n'*duration)
```
knrafto: Should we make this an iterable as an example on how to create a streaming iterator? I think using StringIO/BytesIO is a bit of an anti-pattern (and caused v2 to run out of memory). Something like

```
def data():
    for _ in range(duration):
        yield b'print("Hello World")\n'
```

Incidentally, what happens if we yield strings instead of bytes?

server/controllers/files.py
```diff
@@ -0,0 +1,41 @@
...
+
+@files.route("/files/<hashid:file_id>")
+@login_required
+def file_url(file_id):
+    ext_file = ExternalFile.query.filter_by(id=file_id, deleted=False).first()
+    if not ext_file or not ExternalFile.can(ext_file, current_user, 'download'):
+        abort(404)
+
+    try:
+        storage_obj = ext_file.object
```
knrafto: I find it extremely surprising that accessing a property can raise an exception. We could make it a plain method, or even could catch the exception in the @property and have it return None in that case

server/controllers/files.py
```diff
@@ -0,0 +1,41 @@
...
+    except libcloud.common.types.InvalidCredsError as e:
+        logger.warning("Could not get file {0} - {1}".format(file_id, ext_file.filename),
+                       exc_info=True)
+        storage_obj = None
+
+    if storage_obj is None:
+        abort(404, "File does not exist")
+
+    # Do not use .download_url for local storage.
+    if "local" in storage.driver.name.lower():
```
knrafto: There has to be a better way than string manipulation. I think `storage.provider == Provider.LOCAL` would work?

tests/test_files.py
```diff
@@ -0,0 +1,175 @@
...
+        self.assertEquals(self.blob1.name, self.file1.object_name)
+        self.assertEquals(prefix_obj.name, 'test_fizz.txt')
+        self.assertEquals(prefix.filename, 'fizz.txt')
+        prefix_obj.delete()
+
+    def test_malicious_directory_traversal(self):
+        with open(CWD + "/files/fizzbuzz_after.py", 'rb') as f:
+            prefix = ExternalFile.upload(f, name='fizz.txt',
+                                               user_id=self.staff1.id,
+                                               prefix='test/../../',
```
knrafto: Nice!

server/jobs/example.py
Sumukh: It's run outside of the app context so `url_for` won't work until we import current_app from flask.

server/controllers/files.py
Sumukh: Making it a method makes a lot more sense.

server/models.py
Sumukh: Even in the case where a student gets a staff member to be in their group I think we should still offer some level of protection. (or we should disallow staff members from joining a group with students) 

The other reason to have it is a slightly easier query for all student generated files for a particular assignment. Without it, you could just instead do a query where user.id in student_ids. 

Renaming sounds fine (`staff_file`? `for_staff`?) . Let me know if you want to just remove it now or remove it after we disallow staff members from joining student groups.

knrafto: Renaming is fine (I'm partial to `staff_file`). Good point about it making queries easier

Sumukh:
Should be ready for review/testing now. 

Before deployment: 
- I might just regenerate the migration and add an index for `staff_file` 
- Add secrets to kubernetes config.

knrafto:
A couple bugs I found:

* I uploaded a file with the prefix `../jobs/example` just for kicks and it failed with:
`libcloud.common.types.LibcloudError: <LibcloudError in <class 'libcloud.storage.drivers.s3.S3StorageDriver'> 'Unknown error. Status code: 302'>`
* Uploading a file with the name `🍪.okfile` gave the same error. In addition, the log wouldn't save in the db because of http://stackoverflow.com/a/10959780. Weird because my database URL includes `?charset=utf8mb4`. Can you reproduce?

requirements.txt
```diff
@@ -25,6 +25,11 @@ Flask-Assets>=0.12,<0.12.99
 Flask-Cache==0.13.1
 Flask-Login==0.3.2
 
+# File Storage
+# Use custom version until https://github.com/mardix/flask-cloudy/pull/17 is merged
+https://github.com/okpy/flask-cloudy/archive/master.zip
+# Flask-Cloudy==0.15
```
Sumukh: Yes

Sumukh:
Uploading `foo.txt` to `../jobs/example` makes google cloud issue a 302 to `https://storage.googleapis.com/jobs/examplefoo.txt` - which libcloud doesn't expect so it's going to toss up an exception - which I think is fair.

Considering we will have some degree of control over the prefix - it should be fine. I added some code to optionally prefix the filename with a random string - we can enable that if we need to down the line. 

As for `🍪.okfile` -> I reproduced the issue - the underlying collation in the table was `utf8` - so it didn't really matter if your client was using `utf8mb4`. 

This was my fix: save this config as ~/.my.cnf and then delete the database (via SQL) and create it again (instructions in documentation/migrations.md) 
```
[client]
default-character-set = utf8mb4

[mysql]
default-character-set = utf8mb4
```

It'll work fine in prod since our database is already set to have a default character set of ut8mb4

Sumukh:
Going to be deploying this soon - after configuring the secrets and running the migration.

PR #1027, opened 2017-01-03
Enroll from bCourses

### LINKED ISSUE #987: Import enrollment from bCourses
If we set up bCourses, we could use `server.canvas.api.get_students` to get a list of students that we can enroll straight from bCourses at the touch of a button (the button should probably go in `server/templates/canvas/index.html`). No more rearranging columns in Excel.

knrafto:
Resolves #987. Tested by enrolling last semester's CS 61A locally.

A couple of issues that I'm not sure need solutions right now:
* Enrolling takes almost 30 seconds. Should it be a background job?
* The button is kind of hidden away. You need to know to set up the bCourses configuration.

But, it's enough for 61A to use for next semester.

### DIFF
--- a/server/models.py
+++ b/server/models.py
@@ -817,17 +817,10 @@ class Enrollment(Model):
     @staticmethod
     @transaction
     def enroll_from_form(cid, form):
-        usr = User.lookup(form.email.data)
-        if usr:
-            form.populate_obj(usr)
-        else:
-            usr = User()
-            form.populate_obj(usr)
-            db.session.add(usr)
-        db.session.commit()
         role = form.role.data
         info = {
-            'id': usr.id,
+            'email': form.email.data,
+            'name': form.name.data,
             'sid': form.sid.data,
             'class_account': form.secondary.data,
             'section': form.section.data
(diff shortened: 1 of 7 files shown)


knrafto:
I think it will eventually need to be a background job in order to schedule it periodically. I'll make the change

knrafto:
Okay ready for review
screen shot 2017-01-03 at 12 07 26 am

server/models.py
```diff
@@ -845,57 +838,49 @@ def enroll_from_csv(cid, form):
...
+            else:
+                user = User(name=name, email=email)
+                db.session.add(user)
+            record = Enrollment.query.filter_by(user_id=user.id,
                                                 course_id=cid).one_or_none()
             if not record:
-                record = Enrollment(course_id=cid, user_id=usr_id)
-                new_records.append(record)
-
+                record = Enrollment(course_id=cid, user_id=user.id)
```
Sumukh: It seems like the `Model.from_dict` method might be useful here.

Sumukh: It seems like the `Model.from_dict` method might be useful in this function.

knrafto: Hmm, might be more trouble than it's worth

PR #1030, opened 2017-01-09
Logging to Google Stackdriver

### LINKED ISSUE #802: Log to Google Cloud Stackdriver
Currently only gunicorn logs are written. Request logs should be persistent somewhere, and ideally we should be able to view them with Google Cloud console.

knrafto:
Resolves #802.

We use a custom gunicorn logger class that logs to Google cloud using their client library. This requires setting `PYTHONPATH` and passing in the custom logger module on the command line. The environment variable `GOOGLE_LOG_NAME` sets the log name.

The access log forges the App Engine log format so the logs viewer shows it nicely. Logs from the application are collected on the WSGI environ object and passed back to gunicorn. This includes stack traces for uncaught exceptions. The error log (which includes gunicorn info) is written straight with minimal formatting.

If an error occurs while writing log entries, the exception is printed to stderr. I'm not too sure what to do in this case.

Test by setting `DEBUG` = False in `server/settings/dev.py` and running `env GOOGLE_LOG_NAME=knrafto-test-log PYTHONPATH=$PYTHONPATH:$PWD gunicorn --logger-class server.logging.gunicorn.Logger --bind 127.0.0.1:5000 --workers 3 wsgi:app`. Play around and look at the logs. They'll be under Global > knrafto-test-log in the logs viewer.

The kubernetes health checks are very spammy. One solution would be to have a separate endpoint for the health check (that just returns 200), and filter that out before we ship the log. Should that happen in this PR?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,13 @@ RUN mv docker/nginx/default.conf /etc/nginx/conf.d/default.conf
 
 RUN ./manage.py assets build
 
-CMD nginx; gunicorn --timeout 45 --bind unix:/tmp/server.sock wsgi:app --workers 3
+CMD nginx && \
+    env PYTHONPATH=$PYTHONPATH:$PWD gunicorn \
+        --logger-class server.logging.gunicorn.Logger \
+        --timeout 45 \
+        --bind unix:/tmp/server.sock \
+        --workers 3 \
+        wsgi:app
 
 RUN rm -rf /var/cache/apk/*
 
--- a/server/logging/gunicorn.py
+++ b/server/logging/gunicorn.py
@@ -28,7 +28,7 @@ class ProcessCloudLogger:
     def __init__(self):
         self.logger_pid = None
         self.logger = None
-        self.log_name = os.environ.get('GOOGLE_LOG_NAME', 'unknown')
+        self.log_name = os.environ.get('GOOGLE_LOG_NAME', 'ok-default')
 
     def get_instance(self):
         pid = os.getpid()
--- a/server/logging/gunicorn.py
+++ b/server/logging/gunicorn.py
@@ -1,0 +1,124 @@
+import datetime
+import json
+import logging
+import os
+import traceback
+
+import gcloud.logging
+from google.protobuf.json_format import ParseDict
+from google.protobuf.struct_pb2 import Struct
+import gunicorn.glogging
+
+from server.logging import REQUEST_LOG_VARIABLE
+
+ACCESS_LOG_FORMAT = '%(message)s (%(pathname)s:%(lineno)d, in %(funcName)s)'
+ERROR_LOG_FORMAT = '[gunicorn] %(message)s'
+
+access_formatter = logging.Formatter(ACCESS_LOG_FORMAT)
+error_formatter = logging.Formatter(ERROR_LOG_FORMAT)
+
+def format_time(dt):
+    """Formats a naive datetime as UTC time"""
+    return dt.isoformat() + 'Z'
+
+class ProcessCloudLogger:
+    """Call get_logger() to get a Google Cloud logger instance. Ensures that
+    each process has its own logger.
+    """
+    def __init__(self):
+        self.logger_pid = None
+        self.logger = None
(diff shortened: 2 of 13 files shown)


knrafto:
One issue to resolve before merge: gunicorn is showing internal IPs. We should use X-Forwarded-For instead

Sumukh:
> The kubernetes health checks are very spammy. One solution would be to have a separate endpoint for the health check (that just returns 200), and filter that out before we ship the log. Should that happen in this PR?

Yes - I think so. The k8s convention for health checks is `/healthz`

Potentially useful: http://stackoverflow.com/a/879937/411514

knrafto:
This uses an old version of the `google-cloud` library `gcloud==0.18.3`. Trying to use the newest version fails with `TypeError: Can not find message descriptor by type_url: type.googleapis.com/google.appengine.logging.v1.RequestLog.` There's an open GitHub issue about it, and it should be pretty trivial to upgrade after the fix (whatever it is).

It turns out this version of the library uses the v2 stackdriver API (specifically, `v2beta1`), so actually I don't think we have a deadline to upgrade.

knrafto:
Finally managed to log client IPs by creating a kubernetes ingress service for staging. This is ready for review now

knrafto:
Anything else we should log? We can add them to `start_request_logging`. Maybe `Referer:`?

knrafto:
The stackdriver monitoring (user agent `GoogleStackdriverMonitoring-UptimeChecks(https://cloud.google.com/monitoring)`) is also generating a lot of noise. It hits `/` and `/api/v3/version/`. Perhaps we should also do some filtering by user agent?

Sumukh:
Some observations: 
- Measured the effect of logging locally: On average, gunicorn is spending an additional of 129ms on each request in logging. This time is spent after serving the request so it is not slowing down the user but it does mean the worker is unavailable to serve another request for an extra 129ms. 
- We might log that a request takes 100ms - but the user is actually waiting 150ms for a response. The remainder of the time is somewhere between Nginx -> Google Cloud Load Balancer -> User? 
- Instantclick (The library we use pjax/turbolinks on the student side) can make it seem the user actually went to a page - but they just hovered over the link and the logs can't tell either way - It doesn't include any special HTTP headers when prefetching. 

Re: Ignoring the user agent - I don't think we should be ignoring by user-agent. It's very easy to filter those user agents out in the log viewer. Also if someone didn't want their requests to be logged (because they found a vulnerability) - they could just spoof that user agent.

Dockerfile
```diff
@@ -15,7 +15,13 @@ RUN mv docker/nginx/default.conf /etc/nginx/conf.d/default.conf
 
 RUN ./manage.py assets build
 
-CMD nginx; gunicorn --timeout 45 --bind unix:/tmp/server.sock wsgi:app --workers 3
+CMD nginx && \
+    env PYTHONPATH=$PYTHONPATH:$PWD gunicorn \
```
Sumukh: Curious: Why is this `env PYTHONPATH=$PYTHONPATH:$PWD`

wouldn't just doing `PYTHONPATH=$PYTHONPATH:$PWD gunicorn` work?

server/logging/gunicorn.py
```diff
@@ -0,0 +1,128 @@
...
+    return dt.isoformat() + 'Z'
+
+class ProcessCloudLogger:
+    """Call get_logger() to get a Google Cloud logger instance. Ensures that
+    each process has its own logger.
+    """
+    def __init__(self):
+        self.logger_pid = None
+        self.logger = None
+        self.log_name = os.environ.get('GOOGLE_LOG_NAME', 'unknown')
```
Sumukh: Nit: `ok-default` or something might be better than `unknown`

Dockerfile
knrafto: Oh probably. My shell (fish) actually [doesn't support that syntax](https://fishshell.com/docs/current/faq.html#faq-single-env)

knrafto:
RE gunicorn post-processing time: I'll bet most of that 129ms is waiting for IO, so if we use async gunicorns workers we won't have to wait for the whole thing. #1031

PR #1032, opened 2017-01-17
Notebook viewer

### LINKED ISSUE #1015: Render .ipynb files
Design choices: 
- Provide a URL for nbviewer to view the file and generate a nbviewer embed/iframe. 
   - Doesn't allow for much commenting. :(  
- Run a separate service that uses nbconvert (internally available) that converts the ipynb to html 
- Client Side Rendering: 
  - https://github.com/kokes/nbviewer.js - This appears to work well. 
  - https://github.com/jsvine/notebookjs - Seems to work well too. 

Diffs: Out of scope. 

Security Concerns:
- We can't allow custom CSS/JS/HTML if we render it directly. 

Links: https://nbconvert.readthedocs.io/en/5.x/

I'm very much drawn to the client side rendering instead of running our own microservice or installing jupyter on our server.

Sumukh:
Resolves #1015 

ok-ipython-notebook


server/assets.py
```diff
@@ -25,6 +27,9 @@
 
 common_js = Bundle(
     'js/main.js',
+    'js/notebook.js',
+    'lib/notebookjs/notebook.min.js',
```
Sumukh: out of this PR scope - but for the future: We should figure out how to load these assets (along with all of the others used for code rendering) only in the code view. It seems unnecessary otherwise. 

I'm thinking assets bundles along the lines of code_js / code_css

server/highlight.py
```diff
@@ -8,13 +9,16 @@
 from server.constants import DIFF_SIZE_LIMIT
 
 class File:
-    def __init__(self, lines, too_big=False):
+    def __init__(self, name, lines, source, too_big=False):
         self.lines = lines
+        self.name = name.lower()
```
Sumukh: I don't really use this anywhere - we could remove it.

server/static/js/notebook.js
```diff
@@ -0,0 +1,30 @@
+function initalizeNotebookMarkdown() {
+    // Customize the markdown function
+    var mdRenderer = window.markdownit({
+        html: true
```
Sumukh: The authors of markdownit warn against enabling html: True (for XSS reasons) - but recommend using plugins like markdownItSanitizer (which only allows _very_ basic markdown and neutralizes everything else)

server/templates/diff.html
```diff
@@ -70,30 +106,12 @@
...
-              </td>
-            </tr>
-          {% endif %}
-        {% endfor %}
-      </tbody>
-    </table>
-  </div>
+      {% if file.extension == ".ipynb" %}
+        {{ render_notebook(file) }}
+      {% else %}
```
Sumukh: I like this because it makes it easy to implement previews for other file types (such as images or zip files)

server/highlight.py
```diff
@@ -121,15 +126,15 @@ def diff_files(files_before, files_after, diff_type):
...
+                files[filename] = File(filename, [], after, too_big=True)
             else:
                 lines = list(highlight_diff(filename, before, after, diff_type))
-                files[filename] = File(lines)
+                files[filename] = File(filename, lines, after)
     else:
         for filename, source in files_after.items():
             if len(source) > DIFF_SIZE_LIMIT:
-                files[filename] = File([], too_big=True)
+                files[filename] = File(filename, [], too_big=True)
```
Sumukh: Yup - I forgot to push eecb67e0029cfd842d50a27f2b8e906792ff570f which added default argument so this a bit neater.

server/assets.py
```diff
@@ -25,6 +27,9 @@
...
+    'lib/notebookjs/ansi_up.min.js',
```
Sumukh: I did not make changes to anything in `lib`

tests/test_highlight.py
```diff
@@ -94,7 +94,7 @@ def test_highlight_diff(self):
 
     def test_no_highlight(self):
         filename = 'data'
-        source = 'It was the best of times, it was the worst of times, ...'
+        source = 'It was the best of times, it was the worst of times, ...\n'
```
Sumukh: We changed the behavior to use the Text Lexer by default (instead of just returning the source).

I'm not actually sure if this is the best way to update the test - but it seemed like the cleanest way to keep testing what this was originally testing.

PR #1037, opened 2017-01-18
Use webassets to manage external CSS/JS dependencies

knrafto:
I tested in staging and didn't see an JS errors or weird CSS problems.

This should make it easier to include new libraries in the future, and it protects against CDN failures. From my testing it appears to make the page a few dozen to a hundred ms faster, but not much.

I wasn't able to assetize font-awesome, since if we serve it the CSS expects the fonts to be in static/public/fonts/. Webassets doesn't support binary files, so we'd need some other way of copying the fonts in when we build. Same with the Google font.

The biggest downside is that generating assets now takes quite a while, both in development and building the docker container. It should only happen the first time you get set up with development or change the assets around though. I'm not sure if Docker will cache it as well.


server/assets.py
```diff
@@ -1,19 +1,30 @@
...
 common_css = Bundle(
+    'https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.css',
+    'https://fonts.googleapis.com/css?family=Quicksand:400,700,300|Lato:300,400,700',
+    'https://code.ionicframework.com/ionicons/2.0.1/css/ionicons.css',
+    'https://cdnjs.cloudflare.com/ajax/libs/sweetalert/1.1.3/sweetalert.css',
+    'https://cdnjs.cloudflare.com/ajax/libs/dropzone/4.3.0/dropzone.css',
     'css/helper.css',
     'css/main.css',
     'css/highlight.css',
     'css/code.css',
```
Sumukh: Can these two (code & highlight) be moved to code_css?

PR #1043, opened 2017-01-25
Score audit

### LINKED ISSUE #1007: Score Audit Log

Sumukh:
Original PR: #1013

Score Dashboard:
image

Example of how this dashboard is helpful: I was able to identify that some graders made manual adjustments to give students 2.0 when this homework should only be out of one point. 

Example of what the audit job outputs (with fake data):
image

Work to be done: 
- Change the graphs: They don't really capture what I want to see as a TA
- (Maybe not part of this PR) Once we have an efficient scores query: 
   - Scores: Just grab scores for final submissions - The existing graph just plots all of the non-archived scores
   - Display Submitter information in table. 

Resolves #1007


Sumukh:
I'm going to try to add the histogram instead of the current box plot

knrafto:
A histogram would be fantastic!

I noticed the score view grabs all unarchived scores, so students with multiple submissions with scores will be overrepresented. It won't be a problem now, but it will be when we autograde all submissions

Sumukh:
Yeah - It's an expensive query right now to get ones that are representative of the actual scores you will send to bCourses.

When we autograde all scores - it should be easier to build a more efficient query. (One that grabs the highest score for each student who has submitted for each kind). That query is still inaccurate since group members get the highest score per group - but that's fixable.

knrafto:
Why would it be easier to build a more efficient query then? It won't change anything about how the scores are represented in the DB

Sumukh:
Histogram Example: 

image

re: query:
When we are just choosing the max score - the query for all scores would look like: `SELECT s.user_id, s.kind, MAX(s.score) from score as s WHERE s.assignment_id=? AND s.archived=0 GROUP BY s.user_id, s.kind;`. 

That query is accurate in the sense that those are scores we'd be sending to bCourses/exporting  (I'm ignoring groups - but it's easy to fix that with another query to get all the groups)

With our current setup we basically have to run the FS query - and then get the scores for each backup.

Sumukh:
For bar charts: 

Do you have a preference between: 
image
and 
image

I like the latter since it gives some idea of scale

Sumukh:
Another example: image

server/jobs/scores_audit.py
```diff
@@ -0,0 +1,42 @@
+from collections import defaultdict
+
+from server import jobs
+from server.models import Assignment, Score, db
+
+@jobs.background_job
+def audit_missing_scores(assign_id, omit_details=('autograder', 'regrade')):
```
knrafto: We'll want autograder scores in the future

knrafto: Overall I'm against making scores actually mean something. OK should just treat scores as arbitrary tags, which users can use however they want

knrafto:
Some comments:
* Separate graphs would be better, sorted by number of scores
* The score box should be closable (then you have to reload the page :/)
* Score tables should be full-width
* Make Scores Visible -> Make Scores Visible to Students
* Score Tag -> Score Kind

* Downloading scores for Hog failed :(

Sumukh:
> Score graphs should be full-width 
Hmm - yeah. I'll play around with the layouts

> Downloading scores for Hog failed :(
😞 - This should work better with the max score query.

Sumukh:
Here's the new layout (I've chopped off the top part of the layout) 

image

There are issues with the X axis on larger projects because of the lack of binning. We have to give these charts a bit more height. They can't be full-width and unconstrained height since that just means one chart = 1 screen. 

image

knrafto:
I haven't tried this recently - are there any other changes you want to make before review (besides `get_current_job`)?

Sumukh:
Should be ready to review

knrafto:
Looking at https://ok-staging.cs61a.org/admin/course/4/assignments/82/scores:

The labels on the axes are incredibly small and hard to read. Increasing the font size would help a lot.

The ellipses are annoying too, so here's some suggestions to get rid of some of them:
    * If the label is an integer, show "15" instead of "15.0"
    * Omit every other label (e.g. 0, 2, 4, 6 ...)
    * Make the graphs full-width (though this would take up twice as much vertical space)

screen shot 2017-02-07 at 7 06 23 pm

I noticed some backups have more than one score in the scores table:
screen shot 2017-02-07 at 7 11 51 pm

I would expect only non-archived scores, both in the graph and in the table below.

Sumukh:
It looks like this now - with up to 6 major labels. 

image

Sumukh:
Going to merge & deploy. Feel free to file issues if there's anything else that might be useful for this page.

PR #1045, opened 2017-01-27
Export Submissions

Sumukh:
This generates a zip of all final submissions. 

TODO: 
- [x] Form to generate these zips 
- [x] Generate Research Exports 

- (turns out the standard lib doesn't support this) Password for zips - I want to add this - but it seems doubtful anyone will actually use it. 
- (maybe within this PR) - Grab external files too 
- (maybe within this PR) - Generate research exports - these include every backup submitted & anonymize submission info


Sumukh:
fc28cbe is a somewhat related change but changes things outside of exports

To review just the export changes: https://github.com/Cal-CS-61A-Staff/ok/pull/1045/files/8a4db1e1317ce8ab3bc5c1a4375c9036e39a18cb

server/controllers/admin.py
```diff
@@ -29,6 +29,7 @@
 import server.jobs.example as example
 import server.jobs.moss as moss
 import server.jobs.github_search as github_search
+import server.jobs.export as export
```
knrafto: Nit: I think we can combine these four lines as
`from server.jobs import example, export, github_search, moss`

Sumukh:
For reference: Our large exports with full backup history for a project is about 80mb of memory usage. Normal exports of just submissions is about 2-5mb of memory usage. 

When we include large external files that could be ~1gb - which is probably way too much ram.

Sumukh:
Do you think you could take a crack at using zipstream sometime in the next week? I tried for a bit and ran into some errors when uploading the files (I forgot exactly what the errors were) 

If it's too hard to figure out - I wouldn't be opposed to getting rid of the ability to export all backups since that consumes a lot of memory (75mb+ - whereas the normal exports are <5mb for 61A)

knrafto:
Sure, I'll take a stab at it

knrafto:
If we use zipstream's `writestr`, then we'll still buffer all the contents in memory. We should give it a better interface. Ideally we should pass a generator of `(filename, contents)` and stream a zip file from that. We could probably alter zipstream a little to do this, but we decided it wasn't worth it for now.

PR #1046, opened 2017-01-27
Drag & Drop Uploads + Binary File Upload

Sumukh:
Quite a few things left to cleanup. 

184 needs this in production soon. 

Things I want to change: 
- Code cleanup - I haven't done a pass over it 
- Change the "Uploading" animation to use something from pace.js 
- @kvchen: Suggested a list of files at the top (useful if you have 3+ files) 
- (Maybe not in this PR) Add drag & drop for staff too? 

Things I noticed but don't feel compelled to fix right now
- If you click on the dropzone and get the native file picker - you can't upload folders.  (Edit: Fixed)

Future Ideas:
- Something that looks inside of ZIP files to see the contents.


server/controllers/student.py
```diff
@@ -120,40 +122,78 @@ def submit_assignment(name):
...
     fs = assign.final_submission(user_ids)
-
     if not assign.uploads_enabled:
         flash("This assignment cannot be submitted online", 'warning')
         return redirect(url_for('.assignment', name=assign.name))
     if not assign.active:
         flash("It's too late to submit this assignment", 'warning')
         return redirect(url_for('.assignment', name=assign.name))
 
-    form = UploadSubmissionForm()
```
Sumukh: Should be able to just remove this form entirely once the admin stops using it.

server/controllers/student.py
```diff
@@ -120,40 +122,78 @@ def submit_assignment(name):
...
-                name=assign.name,
-                submit=backup.submit,
-                bid=backup.id,
-            ))
+        assignment = backup.assignment
+        templates = assignment.files or []
+
+        files, binary_files = {}, {}
+
+        sorted_uploads = sorted(list(request.files.items()), key=lambda x: int(x[0][x[0].find('[')+1:-1]))
```
Sumukh: This should be more parseable. 

It exists because `request.files.getlist` doesn't work with the way dropzone names it's files.

server/controllers/student.py
```diff
@@ -120,40 +122,78 @@ def submit_assignment(name):
...
+
+        uploads = [v[1] for v in sorted_uploads]
+        full_path_names = [v for v in request.form.listvalues()][0]
```
Sumukh: Same thing here - it's not clear why things are being indexed

server/controllers/student.py
```diff
@@ -120,40 +122,78 @@ def submit_assignment(name):
...
+            return jsonify({
+                'error': ('Missing files: {}. The following files are required: {}'
+                          .format(', '.join(missing), ', '.join(template_files)))
+            }), 400
+
+        backup_folder_postfix = time.time()
+
+        for full_path, upload in zip(full_path_names, uploads):
+            data = upload.read()
+            if len(data) > 15 * 1024 * 1024:  # file is too large (over 15 MB)
```
Sumukh: Constant?

server/controllers/student.py
```diff
@@ -120,40 +122,78 @@ def submit_assignment(name):
...
+                # TODO: Try uploading files to cloud storage under 20MB if we have enough time.
+                return jsonify({
+                  'error': '{} is larger than the maximum file size of 15MB'.format(full_path)
+                }), 400
+            try:
+                files[full_path] = str(data, 'utf-8')
+            except UnicodeDecodeError:
+                upload.stream.seek(0) # We've already read data, so reset before uploading
+                bin_file = ExternalFile.upload(upload.stream, current_user.id, full_path,
+                                               prefix="uploads/{}/{}/{}/".format(assign.name, current_user.id, backup_folder_postfix),
```
Sumukh: make this str a variable

server/forms.py
```diff
@@ -216,7 +216,7 @@ def validate(self):
 
 
 class AssignmentTemplateForm(BaseForm):
-    template_files = FileField('Template Files', [FileRequired()])
+    template_files = FileField('Template Files')
```
Sumukh: Removed while testing. Should add it back in. 

We should not be use using templates to enforce required files - will file an issue later.

server/settings/prod.py
```diff
@@ -38,7 +38,7 @@
 WTF_CSRF_CHECK_DEFAULT = True
 WTF_CSRF_ENABLED = True
 
-MAX_CONTENT_LENGTH = 15 * 1024 * 1024  # Max Upload Size is 10MB
+MAX_CONTENT_LENGTH = 20 * 1024 * 1024  # Max Upload Size is 20
```
Sumukh: Why is this 20 but the upload is 15MB? 

15mb per file - 20mb total.  I'd support more but I'm not sure if we would timeout first.

server/static/js/dropzone.js
```diff
@@ -0,0 +1,97 @@
...
+  previewTemplate: document.getElementById('dz-preview-template').innerHTML,
+  headers: {
+    'X-CSRFToken': token
+  },
+  init: function() {
+    var submitButton = document.querySelector("#dzSubmit")
+    myDropzone = this; // closure
+    submitButton.addEventListener("click", function() {
+      myDropzone.processQueue(); // Tell Dropzone to process all queued files.
+      swal({
```
Sumukh: Use pace.js here instead.

server/static/js/dropzone.js
```diff
@@ -0,0 +1,97 @@
...
+
+    });
+
+  }
+};
+
+
+window.addEventListener("dragenter", function(e)
+{
+    console.log("Dragging");
```
Sumukh: I was thinking of highlighting the dropzone here - but I might just put that off.

server/templates/diff.html
```diff
@@ -75,6 +75,53 @@
...
+              class="no-dash dl-button source-file-download">
+                  <i class="fa fa-download"></i> Download
+              </a>
+              <a data-no-instant href="{{ url_for('files.file_url', file_id=file.id, raw=1) }}"
+              class="no-dash dl-button source-file-download">
+                  <i class="fa fa-eye"></i> Raw
+              </a>
+          </th>
+        </tr>
+        {% if 'image/' not in file.mimetype %}
```
Sumukh: Remove?

server/models.py
```diff
@@ -1682,6 +1683,10 @@ def mimetype(self):
             return 'application/octet-stream'
         return guess[0]
 
+    @property
+    def download_link(self):
+        return '/files/{}'.format(encode_id(self.id))
```
Sumukh: This is because you are bot always in a context where you can use `url_for` (mainly Jobs). I'd rather update it one place than in all of the jobs.

Sumukh:
I'd also be interested in making sure we aren't storing 5MB+ files in our database by forcing those files to be external.

server/assets.py
```diff
@@ -41,21 +40,35 @@
...
+dropzone_js = Bundle(
+    'https://cdnjs.cloudflare.com/ajax/libs/dropzone/4.3.0/dropzone.js',
+    'js/dropzone.js',
+    filters='jsmin',
+    output='public/js/dropzone.js'
+)
+
 student_css = Bundle(
     'css/student.css',
+    'https://cdnjs.cloudflare.com/ajax/libs/pace/1.0.2/themes/blue/pace-theme-loading-bar.css',
```
Sumukh: Included in student_css because now posting a comment also gets a loading bar now - usually those are processed quickly so I'm okay moving this to only be in `dropzone_css`

server/controllers/student.py
```diff
@@ -120,40 +122,89 @@ def submit_assignment(name):
...
         flash("This assignment cannot be submitted online", 'warning')
         return redirect(url_for('.assignment', name=assign.name))
     if not assign.active:
         flash("It's too late to submit this assignment", 'warning')
         return redirect(url_for('.assignment', name=assign.name))
 
-    form = UploadSubmissionForm()
-    if form.validate_on_submit():
+
+    if request.method == "POST":
```
Sumukh: CSRF validation?

tests/test_api.py
```diff
@@ -104,7 +104,7 @@ def test_non_existant_api(self):
         assert response.json['data'] == {}
         assert response.json['code'] == 404
 
-    def test_aaaget_backup(self):
+    def test_get_backup(self):
```
Sumukh: was this named `test_aaget` for a reason? I suspect it was (maybe to get the test to run first?)

server/static/js/dropzone.js
```diff
@@ -0,0 +1,78 @@
...
+        parallelUploads: 200,
+        maxFiles: 200,
+        maxFilesize: 10, // MB
+        previewTemplate: document.getElementById('dz-preview-template').innerHTML,
+        headers: {
+            'X-CSRFToken': token
+        },
+        init: function() {
+            var submitButton = document.querySelector("#dzSubmit")
+            myDropzone = this; // closure
```
Sumukh: It's unclear why this needed - it was in the example code. I suspect it's so things outside the closure can reference the dropzone.

tests/test_api.py
knrafto: Yeah I think that was me and forgot to change it back

Sumukh:
Should be ready for review now.

server/controllers/student.py
Sumukh: Update: It happens automatically since all routes are protected by default.

server/controllers/files.py
```diff
@@ -40,4 +40,10 @@ def file_url(file_id):
                                                    .format(ext_file.filename))
         return response
     else:
-        return redirect(storage.get_blob_url(storage_obj.name))
+        postpend = ''
+        if request.args.get('raw'):
+            postpend = '&response-content-type={}'.format(ext_file.mimetype)
+        elif request.args.get('download'):
+            postpend = '&response-content-disposition=attachment; filename={0!s}'.format('ext_file.filename')
```
knrafto: This filename should be `os.path.basename(ext_file.filename)` instead of the string `'ext_file.filename'`. Also, use `urllib.urlparse.urlencode` to encode the query string in case the filename has characters that must be escaped.

server/controllers/student.py
```diff
@@ -120,40 +122,89 @@ def submit_assignment(name):
...
+
+        message = Message(kind='file_contents', contents=files)
+        backup.messages.append(message)
+
+        db.session.add(backup)
+        db.session.commit()
+        return jsonify({
+            'backup': backup.hashid,
+            'url': url_for(
+            '.code',
```
knrafto: indent arguments to `url_for` for readability

server/controllers/student.py
```diff
@@ -120,40 +122,89 @@ def submit_assignment(name):
...
-            return redirect(url_for(
-                '.code',
-                name=assign.name,
-                submit=backup.submit,
-                bid=backup.id,
-            ))
+        assignment = backup.assignment
+        templates = assignment.files or []
+
+        files, binary_files = {}, {}
```
knrafto: `binary_files` is a write-only variable, so it should be removed

Sumukh:
Example of the progress bar: 
progress-bar

Side note: At some point we should ensure required files exist with javascript

server/controllers/student.py
```diff
@@ -120,40 +122,89 @@ def submit_assignment(name):
...
+            data = upload.read()
+            if len(data) > MAX_UPLOAD_FILE_SIZE:  # file is too large (over 15 MB)
+                return jsonify({
+                  'error': ('{} is larger than the maximum file size of {}MB'
+                            .format(full_path, MAX_UPLOAD_FILE_SIZE))
+                }), 400
+            try:
+                files[full_path] = str(data, 'utf-8')
+            except UnicodeDecodeError:
+                upload.stream.seek(0) # We've already read data, so reset before uploading
```
knrafto: We already buffer the upload in memory as `data`, so this line is unnecessary and we can use `[data]` in the `ExternalFile.upload` function below (since it must be iterable).

server/controllers/student.py
```diff
@@ -120,40 +122,89 @@ def submit_assignment(name):
...
+                dest_folder = "uploads/{}/{}/{}/".format(assign.name, current_user.id, backup_folder_postfix)
+                bin_file = ExternalFile.upload(upload.stream, current_user.id, full_path,
+                                               prefix=dest_folder, course_id=assign.course.id,
+                                               assignment_id=assign.id)
```
knrafto: we can put `backup=backup` here instead of assigning it separately

server/models.py
```diff
@@ -993,6 +994,11 @@ def files(self):
         else:
             return {}
 
+    def external_files_dict(self):
+        """ Return a dictionary of filenames to contents."""
```
knrafto: I don't think this docstring is accurate, it has ExternalFile objects not contents

server/templates/staff/student/submit.html
```diff
@@ -43,7 +43,7 @@
           </p>
 
           {% call forms.render_form(upload_form, class_='form', enctype="multipart/form-data") %}
-            {{ forms.render_field(upload_form.upload_files, label_visible=true, multiple="", id="file-select") }}
+            {{ forms.render_field(upload_form.upload_files, label_visible=true, multiple="", id="file-select", webkitdirectory='true') }}
```
knrafto: Can we upload folders in non-webkit browsers (Firefox, IE Edge)?

server/templates/student/assignment/code.html
```diff
@@ -66,6 +66,9 @@ <h3 class="text-center">
             {% for filename, file in files | dictsort %}
               {{ diff.source_file(backup, filename, file) }}
             {% endfor %}
+            {% for filename, file in external_files | dictsort %}
+              {{ diff.external_file(backup, filename, file) }}
+            {% endfor %}
```
knrafto: I'd like these to all be in alphabetical order somehow. Since this code is also in the staff template perhaps we can macrotize it

Sumukh:
The tests are hard to fix. I gave it a shot but it just started hanging (which suggests we're doing something selenium doesn't have great support for)

server/templates/staff/student/submit.html
Sumukh: Browser support for this is limited since it's still a draft spec: 
https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/webkitdirectory


IE Edge - Yes. 
Safari (which uses webkit...) - No

server/controllers/student.py
Sumukh: Using data directly through `iter([data])` or `iter(data)` would be nice - but every time I tried using those in staging - the sample for the 184 project 1 would timeout - whereas they wouldn't with `seek(0)`. 

This was reproducible - so I'm going to stick with `upload.stream`

PR #1050, opened 2017-01-31
Do not use string formatting for URL generation, even in jobs

knrafto:
This is mostly a response to [this comment](https://github.com/Cal-CS-61A-Staff/ok/pull/1046#discussion_r98144227): workers have an app context, so we should use url_for in jobs as well instead of string formatting.

### DIFF
--- a/server/jobs/example.py
+++ b/server/jobs/example.py
@@ -1,5 +1,7 @@
 import time
 
+from flask import url_for
+
 from server import jobs
 from server.models import ExternalFile
 from server.utils import encode_id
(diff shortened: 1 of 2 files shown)


Sumukh:
Here's what I got. It looks like we'll need to set a SERVER_NAME in the config

```
Traceback (most recent call last):
  File "/Users/sumukh/ok/server/server/jobs/__init__.py", line 56, in job_handler
    return_value = f(*args, **kwargs)
  File "/Users/sumukh/ok/server/server/jobs/example.py", line 27, in test_job
    .format(url_for('files.file_url', file_id=upload.id), duration))
  File "/usr/local/lib/python3.5/site-packages/flask/helpers.py", line 299, in url_for
    raise RuntimeError('Application was not able to create a URL '
RuntimeError: Application was not able to create a URL adapter for request independent URL generation. You might be able to fix this by setting the SERVER_NAME config variable.
```

knrafto:
Huh, it works with `./manage.py worker` but not `python3 worker.py`

knrafto:
Closing because we don't want to hardcode `SERVER_NAME` in the config and we don't actually need this yet.

PR #1052, opened 2017-02-01
/admin redirect to first active course

### LINKED ISSUE #1051: Default admin view should be an active course
61A veterans report than okpy.org/admin by default shows fa16 instead of sp17

colinschoen:
Fixes #1051

### DIFF
--- a/server/controllers/admin.py
+++ b/server/controllers/admin.py
@@ -87,7 +87,7 @@ def get_courses(cid=None):
 def index():
     courses, current_course = get_courses()
     if courses:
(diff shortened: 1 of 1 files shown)


knrafto:
Oh I actually made a fix for the issue but didn't put it up because I knew the build would fail (it's related to https://wzrd.in being down). My bad

This won't work if the user is an admin in at least one course, but none of those courses are active. I'll push the fix I have up

colinschoen:
Ah good point. Didn't think of no courses being active. Sounds good. I'll close this.

PR #1053, opened 2017-02-01
Redirect /admin to most recently created course

### LINKED ISSUE #1051: Default admin view should be an active course

knrafto:
Fixes #1051

### DIFF
--- a/server/models.py
+++ b/server/models.py
@@ -196,8 +196,8 @@ class User(Model, UserMixin):
     def enrollments(self, roles=None):
         if roles is None:
             roles = [STUDENT_ROLE]
-        query = (Enrollment.query.options(db.joinedload('course'))
-                           .join(Enrollment.course)
+        query = (Enrollment.query.join(Enrollment.course)
+                           .options(db.contains_eager(Enrollment.course))
                            .filter(Enrollment.user_id == self.id)
                            .filter(Enrollment.role.in_(roles))
                            .order_by(Course.created.desc()))
--- a/server/models.py
+++ b/server/models.py
@@ -196,9 +196,10 @@ class User(Model, UserMixin):
     def enrollments(self, roles=None):
         if roles is None:
             roles = [STUDENT_ROLE]
-        query = (Enrollment.query.options(db.joinedload('course'))
+        query = (Enrollment.query.join(Enrollment.course)
                            .filter(Enrollment.user_id == self.id)
-                           .filter(Enrollment.role.in_(roles)))
+                           .filter(Enrollment.role.in_(roles))
+                           .order_by(Course.created.desc()))
         return query.all()
 
     @cache.memoize(120)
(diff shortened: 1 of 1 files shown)


knrafto:
Will test on staging when our deploys work again

server/controllers/admin.py
```diff
@@ -78,7 +78,7 @@ def get_courses(cid=None):
     matching_courses = [c for c in courses if c.id == cid]
     if len(matching_courses) == 0:
         abort(401)
-    current_course = matching_courses[0]
+    current_course = max(matching_courses, key=lambda course: course.created)
```
Sumukh: This is fine - but adding an order_by in the enrollment query seems like the right place for this

knrafto:
Okay, *now* it works. It turns out messing with `current_course` messes with a lot of stuff, so instead I just calculate the best course when we redirect. Tested in staging as a non-admin

Sumukh:
tl;dr - Can we add `.options(db.contains_eager(Enrollment.course))`

Interestingly, just doing the join doesn't seem to save us a query when we later fetch attributes like `r.course`

I think we need to keep the joinedload - the only problem there is that it emits a redundant join

http://docs.sqlalchemy.org/en/latest/orm/loading_relationships.html#the-zen-of-eager-loading

> What we see above is that our usage of Query.join() is to supply JOIN clauses we’d like to use in subsequent query criterion, whereas our usage of joinedload() only concerns itself with the loading of the User.addresses collection, for each User in the result. In this case, the two joins most probably appear redundant - which they are

contains_eager seems like the solution 

```
        query = (Enrollment.query.join(Enrollment.course)
                           .options(db.contains_eager(Enrollment.course))
                           .filter(Enrollment.user_id == self.id)
                           .filter(Enrollment.role.in_(roles))
                           .order_by(Course.created.desc()))
```

Sumukh:
Just saw your commit - I'm fine with keeping the `joinedload`, `contains_eager` might be a better fit but I haven't read into what it really does.

knrafto:
joinedload does indeed created a redundant join:
```SELECT enrollment.created AS enrollment_created, enrollment.user_id AS enrollment_user_id, enrollment.course_id AS enrollment_course_id, enrollment.role AS enrollment_role, enrollment.sid AS enrollment_sid, enrollment.class_account AS enrollment_class_account, enrollment.section AS enrollment_section, course_1.created AS course_1_created, course_1.id AS course_1_id, course_1.offering AS course_1_offering, course_1.institution AS course_1_institution, course_1.display_name AS course_1_display_name, course_1.website AS course_1_website, course_1.active AS course_1_active, course_1.timezone AS course_1_timezone 
FROM enrollment JOIN course ON course.id = enrollment.course_id LEFT OUTER JOIN course AS course_1 ON course_1.id = enrollment.course_id 
WHERE enrollment.user_id = ? AND enrollment.role IN (?, ?, ?, ?, ?) ORDER BY course.created DESC```

PR #1055, opened 2017-02-01
Use our own build of markdown-it-anchor due to wzrd.in instability

knrafto:
It works: http://ok-staging.cs61a.org/admin/grading/dP0qgA

I forked the repo to https://github.com/okpy/markdown-it-anchor and added an npm script that will run browserify. I prefer this to committing the build file into this repo because it should be easy to pull and rebuild if the upstream markdown-it-anchor changes.


Sumukh:
👍

Issue #1031, opened 2017-01-17
Use async gunicorn workers


knrafto:
A lot of our backend makes network requests (db, redis, logging) so using async workers will give us more throughput. There's [a few types to choose from](http://docs.gunicorn.org/en/stable/design.html#async-workers) and it depends on which concurrency framework we want to use (gevent or asyncio) when we possibly use async in our app too.

PR #1056, opened 2017-02-03
Upload larger files (longer timeout)

Sumukh:
We can bump the timeout of the backend service for the load balancer - which enables us to support uploading larger files (184 wants to upload zips as well) 

Side Effect: Pages that were timing out before (hog stats/scores won't now).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,8 @@ RUN ./manage.py assets build
 CMD nginx && \
     env PYTHONPATH=$PYTHONPATH:$PWD gunicorn \
         --logger-class server.logging.gunicorn.Logger \
-        --timeout 180 \
+        --timeout 60 \
+        --threads 2 \
         --bind unix:/tmp/server.sock \
         --workers 3 \
         wsgi:app
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,7 @@ RUN ./manage.py assets build
 CMD nginx && \
     env PYTHONPATH=$PYTHONPATH:$PWD gunicorn \
         --logger-class server.logging.gunicorn.Logger \
-        --timeout 120 \
+        --timeout 180 \
         --bind unix:/tmp/server.sock \
         --workers 3 \
         wsgi:app
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,7 @@ RUN ./manage.py assets build
 CMD nginx && \
     env PYTHONPATH=$PYTHONPATH:$PWD gunicorn \
         --logger-class server.logging.gunicorn.Logger \
-        --timeout 45 \
+        --timeout 120 \
         --bind unix:/tmp/server.sock \
         --workers 3 \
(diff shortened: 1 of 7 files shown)


tests/test_web.py
```diff
@@ -133,72 +133,72 @@ def test_server_is_up_and_running(self):
...
+            self.assertEqual(response.status_code, 200)
+            data = response.json()
+            self.assertEqual(data['message'], 'success')
+            self.assertEqual(data['data']['version'], 'v3')
+
+        def test_static_pages(self):
+            about_url = "{}/about/privacy/".format(self.get_server_url())
+            self.page_load(about_url)
+            self.assertIn('Privacy Policy | Ok', self.driver.title)
+
```
Sumukh: These tests were commented out unnecessarily

tests/test_web.py
```diff
@@ -297,34 +297,34 @@ def test_web_submit_no_file(self):
...
-            self.page_load("{}/admin/course/{}/{}/{}/submit".format(
-                self.get_server_url(),
-                self.course.id,
-                self.user1.email,
-                self.assignment.id,
-            ))
+        #     self._login_as(email=self.staff1.email)
+        #     self.page_load("{}/admin/course/{}/{}/{}/submit".format(
+        #         self.get_server_url(),
+        #         self.course.id,
```
Sumukh: No particular reason that this was commented out. It could be kept - but all of the other file upload tests are commented out too.

Sumukh:
Instead of a much longer timeout for gunicorn - threads appear promising: 

> One benefit from threads is that requests can take longer than the worker timeout while notifying the master process that it is not frozen and should not be killed.

Dockerfile
```diff
@@ -18,7 +18,8 @@ RUN ./manage.py assets build
 CMD nginx && \
     env PYTHONPATH=$PYTHONPATH:$PWD gunicorn \
         --logger-class server.logging.gunicorn.Logger \
-        --timeout 45 \
+        --timeout 60 \
```
Sumukh: 45 seconds would probably also be fine - but I'm increasing it anyway.

Sumukh:
Merging & Deploying this for 184. Feel free to leave comments after the merge.

knrafto:
Threads are only for synchronous workers, so I think it'll be unnecessary when we switch to async workers as in #1031. Can you describe how you tested this PR so we can avoid regressions?

Sumukh:
Test in staging: I uploaded two ~7mb zip files at once and the request completed in about 1.5 minutes. 
When threads were disabled a timeout of 120-180 was required to do the same upload consistently (and would timeout at lower timeout values)

I also had to bump the timeout on the backend services created by k8s on the google networking console.

Sumukh:
#1058 fixes the root cause of the slow uploads. The 2x 7mb test that used to take ~90s now takes 17s (~5x speedup) 

Regardless I think having threads is a good stop gap until #1031

PR #1057, opened 2017-02-03
HTTPS by default

Sumukh:
All requests are made over SSL 

Tested in staging (and briefly in production - by accident 😱). HSTS works as expected. 

There's a low HSTS value. It should be at least 90 days. We can bump it once we roll this out and don't see too many issues. 

This is possible thanks to ok-client using requests with bundled SSL certs. Ok Client's `--insecure` flag should now only be used for development.

### DIFF
--- a/docker/nginx/default.conf
+++ b/docker/nginx/default.conf
@@ -37,6 +37,8 @@ server {
       }
 
       proxy_set_header X-Forwarded-Proto https;
+      add_header Strict-Transport-Security "max-age=3600; includeSubdomains" always;
+
       proxy_set_header Host $http_host;
       # we don't want nginx trying to do something clever with
       # redirects, we set the Host: header above already.
--- a/docker/nginx/default.conf
+++ b/docker/nginx/default.conf
@@ -25,10 +25,8 @@ server {
         if ($http_x_forwarded_proto = "http") {
             return 301 https://$host$request_uri;
         }
-        if ($http_x_forwarded_proto = "https") {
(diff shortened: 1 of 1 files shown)


docker/nginx/default.conf
```diff
@@ -22,18 +22,36 @@ server {
     send_timeout                300;
 
     location /static/ {
+        if ($http_x_forwarded_proto = "http") {
+            return 301 https://$host$request_uri;
+        }
+        if ($http_x_forwarded_proto = "https") {
+            add_header Strict-Transport-Security "max-age=3600;" always;
```
Sumukh: I suppose https://www.nginx.com/resources/wiki/start/topics/depth/ifisevil/ applies here . I could use a map over `$http_x_forwarded_proto` - but it's not clear how to only apply it to https requests. 

Alternatively, we could ignore the rfc and just add it on all requests...

knrafto: That article is scary. I think we can assume that of the protocol is not http then it will be https, so we don't need the if at all.

Sumukh: FWIW - The header was showing up even on HTTP requests - so I think it will anyway (which was surprising) 

Regardless I think it's ok even if we do serve the STS header on HTTP requests.

knrafto:
We should deploy to staging one more time just to be sure

Sumukh:
I agree - thanks for removing the if. I can test it in staging.

Sumukh:
It's not serving the HSTS header on the HTTP request 💯

PR #1058, opened 2017-02-05
Faster Uploads

Sumukh:
Our google cloud storage uploads were really slow because they were being buffered into memory and not streamed by libcloud. 

This PR enables chunked encoding for the GoogleCloudStorage driver from libcloud.  [https://cloud.google.com/storage/docs/streaming](https://cloud.google.com/storage/docs/streaming) 
I've filed an upstream issue about enabling it. 

Another libcloud bug: 
- libcloud doesn't verify the computed hash and hash google provides - it very easily could but that requires a small change to the function (I've noted it in the issue)  https://github.com/apache/libcloud/blob/trunk/libcloud/storage/drivers/s3.py#L478
   - It'd be somewhat annoying for us to fix it without making a fork - I'm waiting to see what the libcloud devs suggest.


Sumukh:
Going to merge & deploy soon because of reports of slow/failed downloads in CS184

PR #1059, opened 2017-02-05
Remove Pace.js

Sumukh:
Pace.js basically amounted to a fake progress bar for file uploads. I added it because I thought it would be more effective than the a swal with a loader gif. 

After #1058, the progress bar seen in dropzone.js is more or less accurate. Whether or not we should remove the sweetalert loading alert is now also worth debate 

Example of Pace vs dropzone:  
image

sweetalert example:
image

For now I'm going with keeping the sweetalert since it really makes it clear that users should wait. The downside is that the alert can obscure the accurate progress bars

### DIFF
--- a/server/assets.py
+++ b/server/assets.py
@@ -48,14 +48,12 @@ common_js = Bundle(
 
 dropzone_css = Bundle(
     'https://cdnjs.cloudflare.com/ajax/libs/dropzone/4.3.0/dropzone.css',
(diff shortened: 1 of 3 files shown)


Sumukh:
Was about to merge and realized there was one more thing I could include. `main.js` (I think) defines `window.paceOptions` - we can remove that.

Issue #1003, opened 2016-12-08
Autograding should be a job


knrafto:
Autograding all final submissions can be a background job. This job can watch for scores getting posted back and report its progress.

* The final submissions query won't timeout the request
* We can report progress on autograding, instead of using the autograder RQ interface
* We can retry submissions that did not get scores after some timeout (#1002)

Sumukh:
The current workflow for most people was just wait for a while and see what happens 

Providing visibility about what's going on would be great. How do you think we can get scores posted back? A query for recent scores on the assignment?

knrafto:
That could work. Another approach would be to generate an "autograder job ID" (this could just be the Ok job ID) and give that to the autograder. The autograder would gives this back to Ok when it posts a score, and we can query for scores for this job ID.

Sumukh:
Resolved by #1066

PR #1066, opened 2017-02-07
Autograding job

### LINKED ISSUE #1002: Some submissions do not have scores after autograding
It seems on every assignment there are one or two submissions that do not get scores when we autograde all submissions. We should investigate why.

knrafto:
Resolves #1002, #1003.

To guard against autograder mishaps, the ok server job will try to autograde a backup again if it doesn't see a score for that backup. Reliability is an end-to-end problem.

In more detail:

We set up a state machine for each backup to check its progress through
the autograder. If any step takes too long, we'll retry autograding that
backup. Ultimately, a backup is considered done when we confirm that
we've received a new score, or if we have reached the retry limit.

Feel free to do some more testing.

### DIFF
--- a/server/autograder.py
+++ b/server/autograder.py
@@ -4,14 +4,14 @@ This module interfaces the OK Server with Autopy. The actual autograding happens
 in a sandboxed environment.
 """
 import collections
+import datetime
 import enum
+import json
+import logging
 import time
 
-import datetime
-import json
-import requests
-import logging
 import oauthlib.common
+import requests
 
 from server import constants, jobs, utils
 from server.models import User, Assignment, Backup, Client, Score, Token, db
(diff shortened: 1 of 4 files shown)


server/controllers/admin.py
```diff
@@ -675,11 +676,14 @@ def autograde(cid, aid):
...
     if form.validate_on_submit():
-        try:
-            autograder.autograde_assignment(assign)
-            flash('Submitted to the autograder', 'success')
-        except ValueError as e:
-            flash(str(e), 'error')
+        job = jobs.enqueue_job(
+            autograder.autograde_assignment,
+            description='Autograde {}'.format(assign.display_name),
+            timeout=2 * 60 * 60,  # 2 hours
```
Sumukh: Given that we have autoscaling turned on again - do you think it's reasonable to lower this timeout? I can also lower the timeout for jobs on the autograder to 5 minutes or so.

Sumukh: I don't have an issue with the overall job timeout being 2 hours -  My comment is more directed at `QUEUED_TIMEOUT`

knrafto: I'm worried that if autoscaling is not on for whatever reason, we'll retry jobs while they're still queued and the autograder will have 4 times as much work to do

server/autograder.py
```diff
@@ -105,3 +107,146 @@ def submit_continous(backup):
...
+            logger.error('Did not receive a score for backup {} after {} retries'.format(
+                utils.encode_id(task.backup_id), MAX_RETRIES))
+            task.set_status(GradingStatus.FAILED)
+        else:
+            task.set_status(GradingStatus.QUEUED)
+            task.job_id = autograde_backup(token, assignment, task.backup_id)
+            task.retries += 1
+
+    while True:
+        time.sleep(POLL_INTERVAL)
```
Sumukh: If there are no errors during autograding - TAs will have to wait 100 seconds before getting updates.

I have a suggestion that might help with that: If there are new scores - show a sampling of the scores that we relieved - that way I can quickly tell if this autograding run is producing all 0s or all errors. Right now there's a line that says that it got a score for backup X - but it's a debug line - so it won't show up on the log. 

What if we collect those scores, choose a random sample and print them out  (or print a distribution of a larger sample)

server/jobs/__init__.py
```diff
@@ -30,17 +30,17 @@ def handle(self, record):
     def contents(self):
         return self.stream.getvalue()
 
-def get_job_id():
```
Sumukh: 👍  - I'll update my export branch to use `get_current_job`

server/autograder.py
knrafto: I agree that the log is not very interactive, but since there's no way to kill a job I don't think getting immediate feedback would be that useful anyway. We already have a score plot in the works and I think that would be a better way to monitor the distribution of scores after the run completes. Should we link to it somehow?

Sumukh: I don't think there's a good way to link to it. I'm ok going without it for now - but even when we do have the score audit - I think having the sampling would be nice if only to get some feedback without going to another page.

knrafto: Yeah that would be nice. Maybe we can share some code with the static scores graph, and stick that in when jobs have better output than just a text log.

PR #1069, opened 2017-02-09
Publications page in about

Sumukh:
I recently wanted a publications page that I could link to. Nothing inside the app links to this page (nor does it need to) 

image


Sumukh:
Bypassing review since this is mainly just adding a HTML page that is not linked anywhere within the app.

PR #1076, opened 2017-02-17
Admin UI Facelift

### LINKED ISSUE #1071: Admin Navbar/Sidebar Changes
Thoughts on these designs?

- They all work well on mobile 
- The exact items in the navbar are tentative - I think we'd need to add course specific actions into the bar 
   - I'm thinking the at the exact actions might change based off what page the user is in 

Design 01 (Navbar with a container) 

image

Design 02: (Navbar with no container - full width) 

image

Design 03 (Collapsed to a mini sidebar) 
image

Sumukh:
This is a UI Facelift centered around removing the sidebar and replacing it with a navbar

This PR also addresses a few other issues: 
- Reduces the docker image size size by 22mb (a 25% reduction in compressed image size) 
- Handles redirects for admins/staff who are not also students 
- Handles the unwieldy list of courses for admins by creating a route that lists courses
- Makes the deploy script not use git tag for images with `-` (which indicates a staging deployment) 
- Hides the grading menu option if there are no grading tasks pending 

Resolves #1071 

Screenshots from local environment (with fake data)

image
image
image


Sumukh:
For context - this was what it looked like before: 
image

It's also deployed to staging if you want to try it out.

Sumukh:
I agree that a course dashboard is a good idea - we can scope a course dashboard into this PR

Sumukh:
There's now a course dashboard - about 13 months after the route was created in v3 :)

Sumukh:
Screenshot of the dashboard: 
image

I'm pretty happy with this. I'm going to put off the Vue.js action bar (the search bar in the navbar) because the enrollment search should be good enough for now.

knrafto:
Nice, looks awesome :)

PR #1077, opened 2017-02-18
Job to notify students of scores

### LINKED ISSUE #1072: Way to notify users about scores
I'm thinking that this is a staff triggered job that just emails students with the scores and the contents of their score tag. 

Whether or not it actually ends up being a job is TBD based off how long it takes to send 1000 emails. 

Testing this at scale might become annoying. I might take this chance to upgrade our sendgrid library version.

Sumukh:
Resolves #1072 

The emails look something like this: 

image

Just tried it out with a dry-run mode on hog and it worked well. 

image

This PR also upgrades our sendgrid dependency to the latest version.

### DIFF
--- a/server/controllers/admin.py
+++ b/server/controllers/admin.py
@@ -577,9 +577,13 @@ def send_scores_job(cid, aid):
 
     form = forms.EmailScoresForm()
     if form.validate_on_submit():
+
+        description = ("Email {assign} scores ({tags})"
+                       .format(assign=assign.display_name,
+                               tags=', '.join([k.title() for k in form.kinds.data])))
         job = jobs.enqueue_job(
             scores_notify.email_scores,
-            description='Emails for {} scores'.format(assign.display_name),
+            description=description,
             timeout=600,
             course_id=cid,
             user_id=current_user.id,
--- a/server/controllers/admin.py
+++ b/server/controllers/admin.py
@@ -578,6 +578,21 @@ def send_scores_job(cid, aid):
     form = forms.EmailScoresForm()
     if form.validate_on_submit():
 
+        for kind in form.kinds.data:
+            if kind not in assign.published_scores:
+                flash(("{0} scores are not visible to students. Please publish "
+                            " those scores and try again").format(kind.title()),
+                      'error')
+
+            return render_template(
+                'staff/jobs/notify_scores.html',
+                courses=courses,
+                current_course=current_course,
+                assignment=assign,
+                form=form,
+            )
+
+
         description = ("Email {assign} scores ({tags})"
                        .format(assign=assign.display_name,
                                tags=', '.join([k.title() for k in form.kinds.data])))
--- a/server/controllers/admin.py
+++ b/server/controllers/admin.py
@@ -610,6 +610,7 @@ def send_scores_job(cid, aid):
             score_tags=form.kinds.data)
         return redirect(url_for('.course_job', cid=cid, job_id=job.id))
     else:
+        form.kinds.data = assign.published_scores
         return render_template(
             'staff/jobs/notify_scores.html',
             courses=courses,
@@ -1188,7 +1189,7 @@ def student_assignment_graph_detail(cid, email, aid):
     user_ids = list(assign.active_user_ids(student.id))
     user_ids.remove(student.id) # should always exist
     user_ids.insert(0, student.id) # insert to front of list
-    
+
     line_charts = []
 
     all_backups = (Backup.query.options(db.joinedload('messages'),
--- a/server/controllers/admin.py
+++ b/server/controllers/admin.py
@@ -32,6 +32,7 @@ import server.jobs.example as example
 import server.jobs.moss as moss
 import server.jobs.scores_audit as scores_audit
 import server.jobs.github_search as github_search
+import server.jobs.scores_notify as scores_notify
 
 import server.highlight as highlight
 import server.utils as utils
@@ -564,6 +565,40 @@ def audit_scores(cid, aid):
                            assign_id=assign.id)
     return redirect(url_for('.course_job', cid=cid, job_id=job.id))
 
+@admin.route("/course/<int:cid>/assignments/<int:aid>/scores/notify",
+             methods=["GET", "POST"])
+@is_staff(course_arg='cid')
+def send_scores_job(cid, aid):
+    courses, current_course = get_courses(cid)
+    assign = Assignment.query.filter_by(id=aid, course_id=cid).one_or_none()
+    if not assign or not Assignment.can(assign, current_user, 'grade'):
+        flash('Cannot access assignment', 'error')
+        return abort(404)
+
+    form = forms.EmailScoresForm()
+    if form.validate_on_submit():
+        job = jobs.enqueue_job(
+            scores_notify.email_scores,
+            description='Emails for {} scores'.format(assign.display_name),
+            timeout=600,
+            course_id=cid,
+            user_id=current_user.id,
+            assignment_id=assign.id,
+            subject=form.subject.data,
+            reply_to=form.reply_to.data,
+            body=form.body.data,
+            dry_run=form.dry_run.data,
+            score_tags=form.kinds.data)
+        return redirect(url_for('.course_job', cid=cid, job_id=job.id))
+    else:
+        return render_template(
+            'staff/jobs/notify_scores.html',
+            courses=courses,
+            current_course=current_course,
+            assignment=assign,
+            form=form,
+        )
+
 
 @admin.route("/course/<int:cid>/assignments/<int:aid>/scores.csv")
 @is_staff(course_arg='cid')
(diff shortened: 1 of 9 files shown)


server/controllers/admin.py
```diff
@@ -32,6 +32,7 @@
 import server.jobs.moss as moss
 import server.jobs.scores_audit as scores_audit
 import server.jobs.github_search as github_search
+import server.jobs.scores_notify as scores_notify
```
Sumukh: #1076 will make all of the imports into a single statement.

server/controllers/admin.py
```diff
@@ -564,6 +565,40 @@ def audit_scores(cid, aid):
...
+    assign = Assignment.query.filter_by(id=aid, course_id=cid).one_or_none()
+    if not assign or not Assignment.can(assign, current_user, 'grade'):
+        flash('Cannot access assignment', 'error')
+        return abort(404)
+
+    form = forms.EmailScoresForm()
+    if form.validate_on_submit():
+        job = jobs.enqueue_job(
+            scores_notify.email_scores,
+            description='Emails for {} scores'.format(assign.display_name),
```
Sumukh: Could also specify what scores kinds you are exporting

server/jobs/scores_notify.py
```diff
@@ -0,0 +1,72 @@
...
+                        .options(db.joinedload('user'))
+                        .filter(Enrollment.role == STUDENT_ROLE,
+                                Enrollment.course == assign.course)
+                        .all())]
+
+    for kind in score_tags:
+        if kind not in assign.published_scores:
+            log.warning(("{0} scores are not visible to students. "
+                         " Please publish them and try again").format(kind))
+            return "Not sent - {} is not published".format(kind)
```
Sumukh: This is to avoid accidentally sending a score that's not ready to be seen yet. 

We could add an override if that use case comes up.

server/templates/email/base.html
```diff
@@ -37,8 +37,7 @@
 }
 
 .btn-primary {
-  Margin-bottom: 10px;
-  width: auto !important;
```
Sumukh: This makes buttons full width - which they weren't before.

server/templates/email/base.html
```diff
@@ -217,7 +219,7 @@
         <table>
           <tr>
             <td align="center">
-              <p> This email was sent to you as a user of <a href="http://okpy.org">okpy.org</a></p>
+              <p> This email was sent to you by a course on <a href="http://okpy.org">okpy.org</a></p>
```
Sumukh: Either "by" or "on behalf of" works

server/templates/staff/course/assignment/assignment.scores.html
```diff
@@ -81,6 +81,9 @@ <h3 class="box-title">Actions</h3>
                   <li> <a href="{{ url_for('.audit_scores', cid=current_course.id, aid=assignment.id) }}" type="button">
                         <i class="fa fa-search"></i> Find Submissions Without Scores
                   </a></li>
+                  <li> <a href="{{ url_for('.send_scores_job', cid=current_course.id, aid=assignment.id) }}" type="button">
```
Sumukh: Nit: At this point - it's worth getting these into a button block instead of having them be nav-pills. 

Probably won't change it in this PR.

server/templates/staff/jobs/notify_scores.html
```diff
@@ -0,0 +1,48 @@
...
+{% block title %} Email {{ assignment.display_name }} scores {% endblock %}
+
+{% block main %}
+<section class="content-header">
+    <h1>
+        Email {{ assignment.display_name }} Scores
+        <small>{{ current_course.offering }}</small>
+    </h1>
+    <ol class="breadcrumb">
+        <li><a href="{{ url_for(".index") }}"><i class="fa fa-dashboard"></i> Home</a></li>
```
Sumukh: This line should not exist but I'll leave it to #1076 to remove it.

server/controllers/admin.py
knrafto: I think that would be better e.g. total vs composition

server/templates/email/base.html
knrafto: by course staff on ...?

knrafto:
I want to test it out myself - probably sometime this weekend

Sumukh:
@knrafto - It's on staging if you want to give it a shot. I'll message you with a link that should work

server/forms.py
```diff
@@ -473,7 +473,23 @@ class GithubSearchRecentForm(BaseForm):
     issue_title = StringField('Issue Title (Optional)', validators=[validators.optional()],
                                 default="Academic Integrity - Please Delete This Repository")
     issue_body = TextAreaField('Issue Body (Optional)', validators=[validators.optional()],
-                               description="The strings '{repo}' and '{author}' will be replace with the approriate value")
+                               description="The strings '{repo}' and '{author}' will be replace with the appropriate value")
```
knrafto: while fixing typos: `will be replaced`

server/jobs/scores_notify.py
knrafto: I feel like this should be in the controller, so staff is notified before the job starts

server/jobs/scores_notify.py
```diff
@@ -0,0 +1,72 @@
...
+    seen_ids = set()
+    for student in students:
+        if student.id in seen_ids:
+            continue
+        user_ids = assign.active_user_ids(student.id)
+        seen_ids |= user_ids
+        scores = [s for s in assign.scores(user_ids) if s.kind in score_tags]
+        if scores:
+            users = User.query.filter(User.id.in_(user_ids))
+            primary, cc = users[0].email, [u.email for u in users[1:]]
```
knrafto: Is the primary email chosen arbitrarily?

server/jobs/scores_notify.py
```diff
@@ -0,0 +1,72 @@
...
+            result = send_email(primary,
+                subject, body,
+                cc=cc,
+                template='email/scores.html',
+                title=subject,
+                from_name=assign.course.display_name,
+                scores=scores,
+                reply_to=reply_to,
+                link_text="View on okpy.org",
+                link="https://okpy.org/" + assign.name, # Don't have url_for
```
knrafto: :(

server/forms.py
```diff
@@ -473,7 +473,23 @@ class GithubSearchRecentForm(BaseForm):
...
+
+class EmailScoresForm(BaseForm):
+    subject = StringField('Subject', validators=[validators.required()])
+    body = TextAreaField('Body', validators=[validators.required()],
+                       description="Is there anything the students need to know about the scores?",
+                       default="An instructor has published scores. If you have questions, please contact the course staff")
+    reply_to = EmailField('Reply To Address', default="no-reply@okpy.org",
```
knrafto: What's the difference between "Reply To" and "From"?

server/forms.py
```diff
@@ -473,7 +473,23 @@ class GithubSearchRecentForm(BaseForm):
     issue_title = StringField('Issue Title (Optional)', validators=[validators.optional()],
                                 default="Academic Integrity - Please Delete This Repository")
     issue_body = TextAreaField('Issue Body (Optional)', validators=[validators.optional()],
-                               description="The strings '{repo}' and '{author}' will be replace with the approriate value")
+                               description="The strings '{repo}' and '{author}' will be replace with the appropriate value")
+
+class EmailScoresForm(BaseForm):
+    subject = StringField('Subject', validators=[validators.required()])
```
knrafto: The default subject "Hog Scores" is pretty good. I wish we didn't have to type it in again

server/templates/email/base.html
Sumukh: How about "as part of a course using" -  

These emails with saying "by a course" (or course staff) is that this applies to partner emails too - and those aren't quite "sent by staff"

server/forms.py
Sumukh: Now you don't! I forgot about the value parameter in the wtforms render function.

server/templates/email/base.html
knrafto: hmm...I think it's fine

server/controllers/admin.py
```diff
@@ -578,6 +578,21 @@ def send_scores_job(cid, aid):
     form = forms.EmailScoresForm()
     if form.validate_on_submit():
 
+        for kind in form.kinds.data:
+            if kind not in assign.published_scores:
+                flash(("{0} scores are not visible to students. Please publish "
+                            " those scores and try again").format(kind.title()),
+                      'error')
+
+            return render_template(
```
knrafto: This should not be inside of the for loop

server/forms.py
Sumukh: From always has to be an okpy.org domain to send reliably - we can't (reliably) spoof the user's email address. 

When someone hits reply their mail client uses the reply-to address if available (and if it's not it uses the "From" address). 

The only thing I'm unsatisfied with in this PR is that we still use the no-reply email even if there is a valid reply to address (making it seem like the user can't reply). I could generate an email for the course I guess?

knrafto:
Feel free to merge

PR #1080, opened 2017-02-18
File API

### LINKED ISSUE #1079: File API
Needed for CS184 to be able to bulk download submissions. 

Should be just a redirect with a permissions check. 

GET `/api/v3/files/a1bc3d` -> redirects to the download URL 

I'm thinking of moving  the logic in `/files/a1bc3d` into the api and making that route a redirect.

Sumukh:
Resolves #1079

### DIFF
--- a/server/controllers/api.py
+++ b/server/controllers/api.py
@@ -664,6 +664,8 @@ class ExportFinal(Resource):
                 'group': [models.User.get_by_id(uid) for uid in backup.owners()],
                 'assignment': assign,
                 'is_late': backup.is_late,
+                'external_files': backup.external_files,
+                'submission_time': backup.submission_time,
                 'submitter': backup.submitter,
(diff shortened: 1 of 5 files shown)


server/controllers/api.py
```diff
@@ -885,6 +889,17 @@ def post(self, user, backup_id):
...
+class File(Resource):
+    """ Redirect (or download) a file. No Schema due to redirect
+        Authenticated. Permissions: >= User/Staff
+        Used by: Course Scripts.
+    """
+    schema = None
+    model = models.ExternalFile
+
+    def get(self, user, file_id):
+        return files.file_download(file_id, user)
```
Sumukh: In a truly RESTful API this would probably have to return a JSON with another URL to download the file - but I don't mind ignoring REST so that we can use the simple solution.

PR #1081, opened 2017-02-18
fixed wrong tag in assignment.html, line 258

Konstantinov-Innokentii:


knrafto:
Thanks for pointing this out! It looks like the "correct" is actually the `h5` though, based on other places where "Total:" appears:

https://github.com/Cal-CS-61A-Staff/ok/blob/master/server/templates/staff/course/enrollment/enrollment.list.html#L73

I found two other instances of the wrong closing tag for "Total":

https://github.com/Cal-CS-61A-Staff/ok/blob/master/server/templates/staff/grading/overview.html#L156
https://github.com/Cal-CS-61A-Staff/ok/blob/master/server/templates/staff/grading/queue.html#L145

PR #1082, opened 2017-02-19
Allow jobs to choose when to save log output

Sumukh:
If a message is logged with a level of critical (or higher) - the log is immediately saved (instead of having to wait for 10 more messages) 

This allows jobs some level of control over what gets displayed when. It's especially useful for jobs that might not print lots of information (like the autograding job - previously the user had to wait for about a minute before they got any output)

### DIFF
--- a/server/autograder.py
+++ b/server/autograder.py
@@ -178,6 +178,9 @@ def autograde_assignment(assignment_id):
     ]
     num_tasks = len(tasks)
 
+    if not num_tasks:
+        return "No submissions to grade"
+
     def retry_task(task):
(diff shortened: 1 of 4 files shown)


Sumukh:
Sure - that should work. What do you think a reasonable interval is? 15?

knrafto:
Sure, it might require some manual tuning to find the right number

colinschoen:
@Sumukh Are you still planning on working on this PR?

Sumukh:
Nope. Feel free to take this over.

PR #1086, opened 2017-02-21
Extensions

Sumukh:
image

image


server/controllers/admin.py
```diff
@@ -1241,6 +1246,91 @@ def student_assignment_graph_detail(cid, email, aid):
...
+        if not form.email.data:
+            form.email.data = request.args.get('email')
+        if not form.expires.data:
+            extra_week = assign.due_date + dt.timedelta(days=7)
+            form.expires.data = utils.local_time_obj(extra_week, current_course)
+
+    if form.validate_on_submit():
+        student = User.lookup(form.email.data)
+        if Extension.get_extension(student, assign):
+            flash("{} already has an extension".format(form.email.data), 'error')
```
Sumukh: I think it might be `danger` instead of `error`

server/controllers/api.py
```diff
@@ -351,14 +352,24 @@ def store_backup(self, user):
...
         backup = make_backup(user, assignment['id'], args['messages'],
-                             elgible_submit)
-        if args['submit'] and lock_flag:
-            raise ValueError('Late Submission of {}'.format(args['assignment']))
-        if elgible_submit and assignment['autograding_key']:
+                             eligible_submit)
+        if args['submit'] and past_due:
+            assign_obj = models.Assignment.by_name(args['assignment'])
+            extension = models.Extension.get_extension(user, assign_obj)
+            # Submissions after the deadline with an extension are allowed
```
Sumukh: It's more that the custom time is set - if the lock date hasn't passed - we'll still accept it regardless of extensions.

server/models.py
```diff
@@ -20,7 +20,7 @@
 
 import contextlib
 import csv
-from datetime import datetime as dt
+import datetime as dt
```
Sumukh: Context for change: This style is suggested by [the style guide](https://github.com/amontalenti/elements-of-python-style#the-standard-library)

server/templates/staff/course/extension/extension.list.html
```diff
@@ -0,0 +1,86 @@
...
+                  <td>
+                        <span class="label label-{{ 'success' if extension.active else 'default'}}">
+                            {{ utils.local_time(extension.expires, current_course) }}
+                        </span>
+                  </td>
+                  <td>
+                      {% if extension.custom_submission_time %}
+                        {{ utils.local_time(extension.custom_submission_time, current_course) }}
+                      {% else %}
+                        N/A
```
Sumukh: `&mdash;` is also an option here

server/models.py
```diff
@@ -1778,3 +1778,56 @@ def can(cls, obj, user, action):
...
+
+    @classmethod
+    def get_extension(cls, student, assignment):
+        """ Returns the extension if the student has an extension """
+        group_members = assignment.active_user_ids(student.id)
+        extension = cls.query.filter(cls.assignment == assignment,
+                                     cls.expires >= dt.datetime.utcnow(),
+                                     cls.user_id.in_(group_members)).first()
+        if not extension:
+            return False
```
knrafto: What benefit does this have over just returning `extension`?

server/models.py
```diff
@@ -1778,3 +1778,56 @@ def can(cls, obj, user, action):
...
+##############
+
+class Extension(Model):
+    """ Extensions allows students to submit after the deadline.
+    """
+    id = db.Column(db.Integer, primary_key=True)
+    staff_id = db.Column(db.ForeignKey("user.id"), nullable=False)
+    assignment_id = db.Column(db.ForeignKey("assignment.id"), nullable=False)
+    user_id = db.Column(db.ForeignKey("user.id"), nullable=False)
+    message = db.Column(mysql.MEDIUMTEXT)
```
knrafto: Does this work with sqlite too?

Sumukh: Yup. We have a `@compiles(mysql.MEDIUMTEXT, 'sqlite')` earlier in the file

server/models.py
Sumukh: thanks :man_facepalming:

Sumukh:
The missing link between course level list of extensions and granting a new extension was bothering me too. 

I think having the course level view of extensions is worth sticking with - we should just fix the form to account for that. I'll make the form have a dropdown for assignment selection (and a way to have it be pre filled if you click in through an assignment)

Sumukh:
Should be ready to go. It's still easy to grant an extension from an assignment (or from the student history overview) 

image

PR #1091, opened 2017-02-25
Ensure client-id is unique

### LINKED ISSUE #970: Server error on duplicate OAuth client IDs
If you create or edit an OAuth client to have the same client ID as an existing client, we'll throw an error. We should check for duplicate IDs before trying to write the update.

colinschoen:
Fixes #970 


server/forms.py
```diff
@@ -409,6 +409,17 @@ class ClientForm(BaseForm):
         'Default Scope',
         description='Comma-separated list. Valid scopes are "email" and "all".')
 
+    def validate(self):
+        # if our validators do not pass
+        if not super(ClientForm, self).validate():
+            return False
+        existingClient = Client.query.filter_by(client_id=self.client_id.data).first()
```
Sumukh: nit: lets not camel case.

colinschoen: `existing_client`?

Sumukh:
Cool! Looks like a good idea to move this validation to the form! (if it was happening at all before)

server/forms.py
Sumukh: Sure!

colinschoen: Done.

server/forms.py
```diff
@@ -419,6 +430,21 @@ class EditClientForm(ClientForm):
         description="Do not fill out or render.",
         validators=[validators.optional()])
 
+    def __init__(self, obj=None, **kwargs):
+        self.obj = obj
+        super(ClientForm, self).__init__(obj=obj, **kwargs)
+
+    def validate(self):
+        # if our validators do not pass
+        if not super(ClientForm, self).validate():
```
knrafto: Should be `EditClientForm`. Actually, this method can be omitted because it's inherited from `ClientForm`, right?

colinschoen: No this needs to be declared in EditClientForm as it has different functionality (checks to see if the `client_id` is changed before checking the query).

server/forms.py
```diff
@@ -419,6 +430,21 @@ class EditClientForm(ClientForm):
...
+            return False
+        if self.client_id.data != self.obj.client_id:
+            existingClient = Client.query.filter_by(client_id=self.client_id.data).first()
```
Sumukh: This one is still camelCased

knrafto:
Thanks. Shame that there isn't an easy way to unify the logic for creating and editing (maybe filter out the origin client-id when checking for uniqueness?) but I think this is good enough

PR #1092, opened 2017-02-25
Improved Moss integration

jathak:
So far, I've changed the Moss job to submit without using the Perl script and then parse the result pages into MossResult entries in the database.

The backend of the new Moss job now allows you to provide two different lists of submissions (`submissions` and `ref_submissions`). Anything in either list will be sent to Moss, but only results where one or both submissions involved is in `submissions` will be stored in the database. The job's form doesn't yet use this, but, in the future, it could allow for referencing submissions from past offerings while only storing results involving current ones.

I now want to work on actually using the stored data to both render a better version of the existing Moss viewer and to allow for similarity statistics (e.g. checking if a student is copying on multiple assignments).


server/templates/staff/plagiarism/list.student.html
```diff
@@ -0,0 +1,93 @@
+{% import 'staff/_helpers.html' as helpers %}
+
```
Sumukh: Could you add an if block here (or outside) so that the box just doesn't show up if there are no cheating results? Many courses will not be able to use the cheating detection tools.

epai:
What's your status on this PR @jathak ?  (and would you like help with any of it?)

jathak:
I believe I have the last changes I need (adding tag support) ready to go, but it appears Moss is down, so I can't test it.

jathak:
Okay, I think this should be good to go; it just needs a database migration. Moss results now support tags. When you run Moss, you can set a threshold where all results with similarity >= to it (on either submission) will be tagged with "review". At the moment, there's no other way to add/edit tags, but that can be done when we actually add an interface for viewing a Moss result.

epai:
@sequoia-tree Would you be interested in taking this project on?

jathak:
My internship ended so I can write code again. What needs to be done here?

epai:
Awesome! 

So I don't remember if I told you this, but there were multiple big bugs that needed fixing:
- (FIXED) The error message for failed moss submission was wrong.  It said something like "Invalid language", but the real problem was that the MOSS ID was not typed correctly.
    - fix:  I've updated the fail message so it's more clear.
- (FIXED) Midfiles weren't being included in matches.  Before, if a match spanned multiple files, it would only include matches in the first and last files, but not any of the middle files.
    - fix:  I've figured out a way to calculate midfiles, and added them as matches
- (NEEDS WORK) Currently, previous moss results aren't cleared away.  We should delete all previous moss results for the assignment if they already exist, so we only show the latest moss info in the moss results page.
- (NEEDS WORK) There is no way to see the diff view in the assignments

There are also some nice-to-haves:
- (ENHANCEMENT) Currently, this moss thing doesn't work for ipynb, which have custom json-like formatting that messes with the similarity matching.  We could fix this by rendering the ipynbs as py files and then sending those over.  Not sure how difficult this is.
- (ENHANCEMENT) lines matched/percentage thresholding is a very poor metric, since it is very much assignment dependent.  It would be better to somehow figure out which questions the lines are part of.  Maybe we could add a way to select question boundaries on the template files?  This is very hard...

jathak:
Moss results now include the time the run was made at and one will now be made for each direction (the backend should already support excluding certain submissions from being the primary on a Moss result, but there's no form for it yet).

The assignment Moss view only displays results from the last run for that assignment.

The student Moss view filters out duplicate submission pairs, but otherwise includes all runs.

You can now specify the number of results to request from Moss as well (instead of just using the default 250).

server/jobs/moss.py
```diff
@@ -113,35 +118,26 @@ def parse_moss_results(base_url, hashed_ids, logger, pattern, template, review_t
...
+            result = MossResult(primary=submissionA, secondary=submissionB,
+                primary_matches=matchesA, secondary_matches=matchesB,
+                tags=['review'] if similarityA >= review_threshold else [], 
+                similarity=similarityA, run_time=run_time)
+            db.session.add(result)
+        if not hashed_ids or hashidB in hashed_ids:
+            result = MossResult(primary=submissionB, secondary=submissionA,
+                primary_matches=matchesB, secondary_matches=matchesA,
+                tags=['review'] if similarityB >= review_threshold else [],
+                similarity=similarityB, run_time=run_time)
```
epai: Instead of `MossResult(...)` why _don't_ we just do `get_or_create` minus the runtime?  If the MossResults are the same, it'll just update the runtime.  This gets rid of needing to filter out duplicates later.

server/models.py
```diff
@@ -1677,18 +1677,28 @@ class Job(Model):
...
+    
+    # Primary submission and matches of this result (filter by this)
+    primary_id = db.Column(db.ForeignKey("backup.id"), nullable=False)
+    # Each file key maps to a list of 2 item lists
+    # (which represent the start and end of the range, inclusive)
+    primary_matches = db.Column(JsonBlob, nullable=False)
+    
+    # Secondary submission and matches of this result
+    secondary_id = db.Column(db.ForeignKey("backup.id"), nullable=False)
+    secondary_matches = db.Column(JsonBlob, nullable=False)
```
epai: I really like the name changes from `...A`/`...B` to `primary`/`secondary`

server/jobs/moss.py
jathak: I think `get_or_create` is a Django thing. I also now think storing once for each run might make sense, so we could easily view all results from a particular run.

jathak:
What do we still need to do before merging this? I think interfaces can be done in separate PRs

epai:
Let's just get this merged in asap, and then we can merge sequoia's views later.  I've just added the new migration file.  (with much fumbling)

jathak:
Yeah, but we should wait for the CI first.

epai:
The test that times out is for test_highlight (which we don't touch), and it succeeds in all but one container.  I think we're good to merge.

PR #1095, opened 2017-02-26
Indicate if submission was made by a staff member in admin view

### LINKED ISSUE #1075: Indicate if submission was made by staff in admin view
The admin panel does not indicate if a submission was made by staff - so it can be confusing if you create an upload through the staff dashboard and it's not obvious that your upload was successful (in the table). 

Files to change: https://github.com/Cal-CS-61A-Staff/ok/blob/master/server/templates/staff/student/assignment.html#L212 (edited)

What to change: you'd probably just change that line in the template to have an if block that says `if backup.creator` -> staff@example.org on behalf of student@example.com  `else` student@example.com

reported by: @kavigupta

wrobe0709:
Fixes #1075. If a submission was made by a staff member then it reads "staff member email" on behalf of 
"student email"

### DIFF
--- a/server/templates/staff/student/assignment.html
+++ b/server/templates/staff/student/assignment.html
@@ -211,7 +211,11 @@
                             </td>
                             <td> {{ utils.local_time(item.created, current_course) }} </td>
                             <td>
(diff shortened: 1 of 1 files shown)


Sumukh:
Thanks. I'm going to try this out and then deploy it

PR #1098, opened 2017-02-28
API endpoint to edit client versions

knrafto:
I'm taking a stab at on automating ok-client releases (https://github.com/Cal-CS-61A-Staff/ok-client/issues/264) and realized we'd need this

### DIFF
--- a/server/settings/test.py
+++ b/server/settings/test.py
@@ -30,7 +30,7 @@ STORAGE_CONTAINER = os.path.abspath("./local-storage")
 if not os.path.exists(STORAGE_CONTAINER):
     os.makedirs(STORAGE_CONTAINER)
 
-CACHE_TYPE = 'null'
+CACHE_TYPE = 'simple'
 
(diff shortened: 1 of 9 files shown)


server/settings/test.py
```diff
@@ -30,7 +30,7 @@
 if not os.path.exists(STORAGE_CONTAINER):
     os.makedirs(STORAGE_CONTAINER)
 
-CACHE_TYPE = 'simple'
+CACHE_TYPE = 'null'
```
Sumukh: Just curious: What was failing because of this?

knrafto: The version test does GET /API/v3/versions/ again to see if it actually changed, but the result is cached from the first call. I'm not sure if there's another way around that

knrafto: The right thing to do is invalidate the cache. I've pushed a fix

knrafto:
I've made some major changes, mind taking another look?

PR #1099, opened 2017-03-01
Notebook Rendering: HTML tables & KaTeX

Sumukh:
image

### DIFF
--- a/server/static/css/notebook.css
+++ b/server/static/css/notebook.css
@@ -1,5 +1,5 @@
 .notebook-pad {
-    width: calc(100% - 5em);
+    width: calc(100% - 7em);
     float: right;
     margin-top: 10px;
 }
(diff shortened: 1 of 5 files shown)


server/static/css/notebook.css
```diff
@@ -15,10 +15,6 @@
     width: inherit !important;
 }
 
-.nb-markdown-cell * {
-    font-family: Helvetica Neue, Helvetica, sans-serif !important;
```
Sumukh: This messed with KaTeX rendering of text. I just opted to remove it and allow notebook fonts to match the surrounding page.

Sumukh:
The impact on page load time is small: 
- An extra 85kb on the first load of a code page - which is mostly webfonts for KaTeX from a cdn. 
- Then it's cached and there's no additional impact. 
- KaTeX is pretty performant so it doesn't take that long to render.

Sumukh:
This isn't quite ready to merge yet. 

- `</tbody>` is showing up for some reason 
- KaTeX fonts use relative URLs when they are getting the fonts - which works fine in dev but not in production when bundle everything up.

knrafto:
I hade a similar problem when bundling font-awesome. I tried bundling the fonts too somehow but webassets does not support binary files

Sumukh:
Things should work now. I'm not super happy about having to resort to using `vw` for the max-width of an output - but it works.

Sumukh:
Any comments? This is mostly just some minor CSS tweaks and pulling in some libraries (though it didn't feel like that) 

Probably going to merge this sometime tomorrow if not.

Issue #1103, opened 2017-03-02
Improve Landing Page


Sumukh:
For SIGCSE: 

- More details on benefits for instructors (What problems do we solve other than composition feedback) 
   - Extensions 
   - Slip Day Calculations 
   - TA Support 
   - JupyterHub integration 
   - MOSS Uploads
- Video demo (I have one recorded - it could be better but it should work for now) 
- Include publications on home page?

PR #1107, opened 2017-03-04
Landing Page Improvements & Course Creation by any user

Sumukh:
Resolves: #1103 

landingpage

- ~The video/demo block is hidden until someone clicks "View Demo"~


Sumukh:
The landing page is no longer a WiP but I'm still calling this a PR since I'll just add the create courses  form into into this branch.

Sumukh:
New Course Dashboard Content: 
image

Sumukh:
This is going to be deployed by Wed morning for a conference but I also don't want to merge it without someone else looking at the authentication changes. 

If you are short on time, feel free to just look at the auth changes and comment on the other changes later.

server/controllers/admin.py
```diff
@@ -290,7 +302,7 @@ def autograde_backup(bid):
     return redirect(url_for('.grading', bid=bid))
 
 @admin.route("/versions/<name>", methods=['GET', 'POST'])
-@is_staff()
+@is_admin()
```
knrafto: We should make the argument to `is_staff` required (i.e. `is_staff(course_arg)` instead of `is_staff(course_arg=None)`) so that we're not tempted to use it, or make sure we got all the uses

server/controllers/admin.py
```diff
@@ -352,14 +362,17 @@ def create_course():
 @is_staff(course_arg='cid')
 def course(cid):
     courses, current_course = get_courses(cid)
+
     students = (Enrollment.query
                           .options(db.joinedload('user'))
                           .filter(Enrollment.role == STUDENT_ROLE,
                                   Enrollment.course == current_course)
                             .all())
+    needs_intro = len(students) < 5
```
knrafto: This seems a little arbitrary. Not sure of a better condition, though. If the example assignment is the only assignment?

server/templates/staff/course/course.html
```diff
@@ -149,30 +149,50 @@ <h3 class="box-title">Quick Links</h3>
...
               <div class="box-body">
                 <div class="col-xs-12">
+                  <h4> Getting Started with OK: </h4>
+                  <p> We have created an <a href="{{ url_for('.course_assignments', cid=current_course.id) }}">example assignment</a> for you. </p>
+                  <ol>
+                    <li> Check out <a href="{{ url_for('student.course', offering=current_course.offering) }}">the student interface</a> and submit to the assignment </li>
+                    <li> Download a sample assignment and change the endpoint in the <code>.ok</code> file to <code>{{ current_course.offering }}/example</code>
+                        <ul>
+                            <li><a href="http://inst.eecs.berkeley.edu/~cs61a/fa16/proj/maps/" >Example Python 3 Project</a> &mdash; <a href="http://inst.eecs.berkeley.edu/~cs61a/fa16/proj/maps/maps.zip">(Download)</a></li>
+                            <li>Example Data Science Project  &mdash; <a href="https://github.com/data-8/data8assets/blob/gh-pages/materials_zips/sp17/hw/hw04.zip">(Download)</a></li>
```
knrafto: Is there a link to the Example Data Science Project spec like there is for the Example Python project?

knrafto:
We're gonna need a way to delete a course. Both so users can do it themselves, and admins can do it easily.

We should validate the semester to be `(fa|sp|su)\d\d`. Maybe winter too?

screen shot 2017-03-06 at 5 47 44 pm

server/templates/staff/course/course.html
Sumukh: Nope - the description/spec is in the notebook.

I'm working on a full featured assignment bank - so this is only a temporary placeholder.

Sumukh:
I think soft deletes make sense for courses. 

Related question: Also what about the quarter system? How do schools notate those?

knrafto:
Based on some googling it looks like Autumn/Winter/Spring/Summer. What if we add `wi` for winter?

server/controllers/admin.py
Sumukh: There are definitely still use cases for `is_staff()` like the listing of all courses or the grading view. There are a few views in the admin interface that don't know/have a course attached - but also aren't admin actions.

Sumukh: I did a pass and confirmed that `is_staff()` is used appropriately but I could see a strong case for renaming it so that future maintainers are less confused.

Sumukh:
I don't want to run a migration right now so I've filed an issue for deleting courses #1117 . 

I'm going to merge & deploy when tests pass.

Issue #1084, opened 2017-02-21
Use SendGrid personalizations for sending score emails


knrafto:
Currently, we send emails one at a time to SendGrid, which is very slow. We should be able to send ~950 at a time instead, just swapping out the score text.

PR #1112, opened 2017-03-05
Send group emails correctly

Sumukh:
These were previously 400'ing because we were passing a list when sendgrid expected a single string (which changed because we upgraded the sendgrid library version a while back) 

The "right" way to do this is to use sendgrid personalizations - but that's an issue we can shelve for later #1084

### DIFF
--- a/server/controllers/student.py
+++ b/server/controllers/student.py
@@ -14,8 +14,7 @@ from server.constants import VALID_ROLES, STAFF_ROLES, STUDENT_ROLE, MAX_UPLOAD_
 from server.forms import CSRFForm, UploadSubmissionForm
 from server.models import (User, Course, Assignment, Group, Backup, Message,
                            ExternalFile, Extension, db)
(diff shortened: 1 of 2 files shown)


server/controllers/student.py
```diff
@@ -396,29 +395,28 @@ def group_respond(name):
...
         try:
-
             if action == "accept":
                 group.accept(current_user)
                 subject = "{0} has accepted the invitation to join your group".format(current_user.email)
                 body = "Your group for {0} now has {1} members".format(assignment.display_name,
                                                                        len(group.members))
-                group_action_email(group.members, subject, body)
+                members = [m.user.email for m in group.members]
+                send_emails(members, subject, body)
```
Sumukh: We call `send_emails` at the bottom of each cases  - ideas for something better that's a bit more DRY?

Sumukh:
Going to merge this since our existing implementation is broken.

PR #1124, opened 2017-03-21
Assign credit job

Sumukh:
This helps courses that: 
- Have checkpoint grading that doesn't involve running the autograder
- Award credit for submission 

image


server/jobs/checkpoint.py
```diff
@@ -0,0 +1,55 @@
...
+    current_user = jobs.get_current_job().user
+
+    assignment = Assignment.query.get(assign_id)
+    students = [e.user_id for e in assignment.course.get_students()]
+    submission_time = server_time_obj(deadline, assignment.course)
+
+    # Find all submissions (or backups) before the deadline
+    backups = Backup.query.filter(
+        Backup.assignment_id == assign_id,
+        or_(Backup.created <= deadline, Backup.custom_submission_time <= deadline)
```
knrafto: I think the backup class that has some magic property that computes the real submission time. Would that work here?

server/jobs/checkpoint.py
```diff
@@ -0,0 +1,55 @@
...
+    ).order_by(Backup.created.desc()).group_by(Backup.submitter_id)
+
+    if not include_backups:
+        backups = backups.filter(Backup.submit == True)
+
+    total_count, all_backups = backups.count(),  backups.all()
```
knrafto: Not sure if this issues two queries. You could do len(all_backups) instead

Sumukh:
Data 8 needs to do checkpoint grading - assigning credit to all submissions submitted before X. 

It's slightly different from autograding all backups and also matches behavior I've seen in other courses where they assign credit to all submissions (regardless of whats in the content)

server/jobs/checkpoint.py
Sumukh: I tried. The hybrid_property returned zero results :(

knrafto:
Slight wording nitpicks: Change "Due Date" to "Checkpoint Date" and I think the question mark in "include submissions too?" is confusing, since it's the opposite question asked by the checkbox.

Other than that, feel free to merge

PR #1128, opened 2017-03-25
Enhancement/wrobe0709/list unenrolled submissions

wrobe0709:
If a student is not enrolled in a class but submitting assignments for the class their submissions won't be seen by the staff.

This adds a view called "Un-Enrolled Submitters" which allows staff members to see any students who are submitting assignments to the course but are not enrolled.

### DIFF
--- a/server/controllers/admin.py
+++ b/server/controllers/admin.py
@@ -450,7 +450,7 @@ def assignment_stats(cid, aid):
         return abort(401)
 
     stats = Assignment.assignment_stats(assign.id)
-
+    
     submissions = [d for d in stats.pop('raw_data')]
 
     pie_chart = pygal.Pie(half_pie=True, disable_xml_declaration=True,
@@ -941,7 +941,7 @@ def enrollment(cid):
     students = current_course.get_students()
     staff = current_course.get_staff()
     lab_assistants = current_course.get_participants([LAB_ASSISTANT_ROLE])
-
+    
     return render_template('staff/course/enrollment/enrollment.html',
                            enrollments=students, staff=staff,
                            lab_assistants=lab_assistants,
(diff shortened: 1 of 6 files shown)


server/controllers/admin.py
```diff
@@ -1043,6 +1043,24 @@ def client(client_id):
         return redirect(url_for(".clients"))
     return render_template('staff/edit_client.html', client=client, form=form, courses=courses)
 
+@admin.route("/course/<int:cid>/unenrolled", methods=['GET'])
+@is_staff(course_arg='cid')
+def unenrolled(cid):
+    courses, current_course = get_courses(cid)
+    
+    submissions = set(b.submitter_id for b in (Backup.query.join(Backup.assignment).filter(Assignment.course_id == cid).all()))
+    enrollment = set(e.user_id for e in (Enrollment.query.filter(Enrollment.course_id == cid).all()))
```
Sumukh: These two lines are a bit too long. I would suggest putting the queries on their own lines

Sumukh: Maybe you can use course.get_students here? It also does a joined load with Users - so you can just do e.user without the cost of another query

server/controllers/admin.py
```diff
@@ -1043,6 +1043,24 @@ def client(client_id):
...
+    unenrolled_submitters = []
+    
+    for s in submissions:
+        if s not in enrollment:
+            unenrolled_submitters.append(User.query.get(s))
```
Sumukh: See above - you should be able to do e.user if you use enrollment objects directly

server/templates/staff/course/enrollment/enrollment.unenrolled.html
```diff
@@ -0,0 +1,71 @@
...
+
+                        <tbody class="list">
+                            {% for unenrolled_submitter in unenrolled_submitters %}
+                            <tr>
+                                <td class="email">
+                                    <a href="{{ url_for('.student_view', cid=current_course.id, email=unenrolled_submitter.email) }}">
+                                        {{ unenrolled_submitter.email }}
+                                    </a>
+                                </td>
+                                <td class="sid">{{ unenrolled_submitter.id }}</td>
```
Sumukh: The ID isn't particularly useful (it's our database ID)  Maybe their name could go here instead?

wrobe0709:
I'm pulling in staff and lab assistants because I was getting them in the list of un-enrolled submitters. I'm assuming we don't want them in that list and that we just want students?

Because I can get the user objects from `get_students()`, `get_staff()`, and `get_participants([LAB_ASSISTANT_ROLE]))` I made `submissions` a set of user objects. Does that look like the best way to do it?

PR #1157, opened 2017-06-19
fix #1156: score ordering

### LINKED ISSUE #1156: Scores not properly sorted by date on `view scores` assignment view
Shows up as:
default

But after toggling the sort-by arrow on and off:
after

dementrock:
This should fix issue #1156


epai:
Seems like `./manage.py assets build` is failing because of an out-of-date asset.  Otherwise, looks good!

PR #1167, opened 2017-08-06
Automatically Grade on Effort

epai:
Instead of grading on correctness and lowering the number of questions needed to get correct, it'd be nice if there was a way to magically give students credit for the questions they put effort into solving, but didn't quite get correct.

To address this, I propose the following heuristic for determining whether a student has put in enough effort.  A student gets full credit for a question if either:
1.  They get the question correct
2.  They pass at least one testcase
3.  They attempted the question at least three times

We can figure out this info from the `messages` that the ok-client collects when students submit backups.  1 and 2 can be extracted from `messages[grading][<question>]`, and 3 can be found in  `messages[analytics][history][questions][<question>]`.

Because this is question-based, my implementation gives students partial credit for the questions they do spend effort on.  This is better than the all-or-nothing approach currently done, as it awards students who try the entire assignment, even if they couldn't get specific questions correct.

I am planning to merge this in by Sunday, August 6.  Any feedback before then is welcome!

snip20170806_153

snip20170807_166

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -19,3 +19,4 @@ test:
   override:
     - py.test --junitxml $CIRCLE_TEST_REPORTS/django/results.xml --cov-report term-missing --cov=server tests/ :
         parallel: true # not a huge advantage since the test suite is short, but evenutally could be useful
+        timeout: 1200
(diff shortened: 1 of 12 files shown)


epai:
(I'll take a look at the failed tests tomorrow - a little burnt out right now)

epai:
For some reason, 1 out of 4 containers times out in the web tests.  Don't know why that specific container times out, but it doesn't look like it's an issue with the effort tests I wrote.  Going to merge.

PR #1176, opened 2017-09-02
Update Effort Policy

epai:
Notable Changes:

- [X] New effort policy:  Full credit for question if at least one test passed OR at least 5 attempts
- [X] Respects custom submission times now, and also chooses the backup amongst submitted and late that gives the best score
- [x] Use autograder as a fallback for when effort grading fails (enhancement)


epai:
Looks like the 1 failing test is caused by two dates that are off by a second.  I don't think that matters...

PR #1179, opened 2017-09-08
MOSS Result Viewer

sequoia-tree:
Fun fact: MOSS is a plant, not a fungus!
Also the MOSS viewer is done. Click the percent similarity between 2 MOSS results, to see the lines that MOSS tells us are similar (read: "mossy").


epai:
Hmm, @sequoia-tree I'm not sure what I'm seeing exactly.  It's kinda hard to parse like this - do you think it'd be possible to have two columns like the github diff viewer?

snip20170908_6

PR #1185, opened 2017-10-12
Data8fa17/upload scores

yixuantang:


AustenZhu:
Creates a new page under every assignment called upload that allows uploading csv's with columns: 
student email | score

Also allows for uploading scores through text.

PR #1193, opened 2017-10-26
Bugfix/score display order

ruk91:


server/controllers/admin.py
```diff
@@ -557,6 +557,10 @@ def view_scores(cid, aid):
 
     if not include_all:
         query = query.filter_by(archived=False)
+
+    # sort scores by submission time in descending order, to match front end display
+    query = query.order_by(Score.created.desc())
```
colinschoen: Can you just chain .order_by to the original query on line `555` since this it not conditionally happening.

colinschoen:
Bump

colinschoen:
Feel free to reopen when you are available to make this change.

PR #1195, opened 2017-10-31
Add instruction on how to run workers

### LINKED ISSUE #1175: Add instructions for starting rq workers
How to run jobs using rq workers is not documented anywhere, apparently.  We should add a small bit to the README with instructions for how to install redis, and that they can run a worker with `./manage.py worker`

jezeniel:
This will close #1175

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,19 @@ The server will listen on http://localhost:5000.
 
 If you are running into issues - see `documentation/SETUP.md` or file an issue
 
+Running Workers
+---------------
+To run workers locally:
+
+```bash
+$ ./manage.py worker
+```
+
+To be able to run the workers you should have a `redis` server installed and running.
+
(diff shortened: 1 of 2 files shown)


jezeniel:
@colinschoen any tips to do this? there a lots of platform or should we focus on only one like ubuntu?

colinschoen:
Ah, let's just provide a link to here:

https://redis.io/topics/quickstart

That should be enough.

jezeniel:
@colinschoen i added the changes. i squashed the commit for your review.

colinschoen:
Can you try pushing one more commit. It can just be a space or word change. It looks like circleci is timing out. Pushing another commit should trigger it once more.

jezeniel:
i think there is a problem with circle ci.

colinschoen:
Fixed in #1198.

Will merge this after.

PR #1204, opened 2017-11-23
Default grading link to full diff (#1203)

### LINKED ISSUE #1203: Opening files in the grading tab should default to full diff
For composition grading, it would make sense that the files that we open should default to the full diff version as we may need to grade portions of the code not visible in the short version of the diff.

ethanlee16:
Resolves #1203 - just a template change to make the grading tab show the full diff by default.

### DIFF
--- a/server/templates/staff/grading/queue.html
+++ b/server/templates/staff/grading/queue.html
@@ -122,7 +122,7 @@
                     {% endif %}
                     <td>
                     {% if item.score_id %}
-                    <a href="{{ url_for(grading_view, bid=item.backup.id, diff='full')}}" class="btn btn-block btn-info btn-flat">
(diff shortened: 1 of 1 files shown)


colinschoen:
Do we want to do this exclusively for when following links for grading or just when opening the diffs by default?

ethanlee16:
Are there other ways that you can reach the diff view (besides going to the direct address - but even then I'm not sure you would ever navigate directly to a grading link with a backup ID)?

If there are other references to `grading_view`, I'll definitely change it to take the `full` option by default in the [controller](https://github.com/Cal-CS-61A-Staff/ok/blob/master/server/controllers/admin.py#L124) instead.

colinschoen:
Grr, can you try pushing again to trigger CircleCI to build again or @epai can you merge ignoring that check as a repo owner?

epai:
Sure - sorry, my bad for not seeing this earlier, this would've been so helpful.  Thanks @ethanlee16!

PR #1212, opened 2017-12-18
Export Course Grades as CSV

epai:
Allows student grades to be exported as a CSV file.  

Currently splits scores by assignment and "score type", where score type is either "Total", "Composition", "Checkpoint 1", or "Checkpoint 2".  

Only assignments that are visible to students and have published scores will be "checked" by default.

Note:  Exporting may take a while (1-2 hours)

### DIFF
--- a/server/jobs/export_grades.py
+++ b/server/jobs/export_grades.py
@@ -41,13 +41,14 @@ def get_score_types(assignment):
 
 def get_headers(assignments):
     headers = ['Email', 'SID']
+    new_assignments = []
     for assignment in assignments:
         new_headers = ['{} ({})'.format(assignment.display_name, score_type.title()) for
                 score_type in get_score_types(assignment)]
         if new_headers:
             new_assignments.append(assignment)
             headers.extend(new_headers)
-    return headers, assignments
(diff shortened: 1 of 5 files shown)


server/jobs/export_grades.py
```diff
@@ -0,0 +1,109 @@
...
+        scores['total'] = total_score
+    if scores_checker(scores, COMP_KINDS):
+        composition_score = max(score_grabber(scores, COMP_KINDS))
+        scores['composition'] = composition_score
+    return scores
+
+
+def get_headers(assignments):
+    headers = ['Email', 'SID']
+    new_assignments = []
```
colinschoen: This doesn't seem to be used.

epai:
Whoops, turns out `new_assignments` was necessary (just forgot to return it instead of "assignments")

epai:
Great!  I'm going to try testing it out on staging to make sure it works, then I'll merge it in

colinschoen:
@ericpai is this good to go?

epai:
should be good to go, although it's a bit slow

PR #1214, opened 2017-12-19
Export grades CSV fix

colinschoen:
Fixes issue of incorrectly formatted CSV caused by a student not having a score type that exists for an assignment. Defaults to issuing `0` score when a score does not exist. 


colinschoen:
Fixes PR #1212

Issue #1173, opened 2017-08-22
Batch Enrollment Should Take CSV/JSON Files


epai:
# Problem
The Batch Enrollment form currently only accepts a fixed-schema, line-separated CSV in a big text box, which is very annoying.  We should be able to specify a file to upload, and the field-names in that file that correspond to the fields we care about.

## Old
![snip20170822_171](https://user-images.githubusercontent.com/8680381/29576131-9bc1551c-871b-11e7-8fff-b2e89434a3f7.png)

Sumukh:
Nice to have - but this might be way too much work for the effort.

colinschoen:
@epai Still attached to this? I can take if you are busy Eric.

PR #1221, opened 2018-01-03
Better Email Check

addcninblue:
It wasn't in the issue tracker, but it was listed as a TODO (discovered while I was looking at issue #1173). Please let me know if this pull request was done wrong (because I'm not sure how pr's work when I'm 4 commits ahead of 61a/ok)

Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ To install:
   - If brew cannot find `virtualenv`, use `brew install pyenv-virtualenv`.
 * Create a virtualenv with `virtualenv -p python3 env`
 * Activate the virtualenv with `source env/bin/activate`
-* (Optional, but reccomended) Install `redis-server`. You can do `brew install redis` on a mac or `apt-get install redis-server`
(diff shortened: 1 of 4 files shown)


Sumukh:
Looks good. Let's be sure to use github's squash and merge when merging this in.

re: 
> (because I'm not sure how pr's work when I'm 4 commits ahead of 61a/ok)

This is fine, but ideally the PR would just contain the latest commit (38ab101). This post has some tips on how you can do that: https://spin.atomicobject.com/2017/04/23/maintain-clean-git-history/

Sumukh:
Note that CircleCI is not going to pass since this was a PR from an external repo. Someone should re-evaluate if we really need the protected branch or if CircleCI really needs to be configured to not run external PRs

colinschoen:
Any reason CircleCi should not run on external PRs?

epai:
hmm, can you try making this change off of the `epai/grading_export` branch?  I made a bunch of changes to how csvs are validated.  see #1212

PR #1222, opened 2018-01-09
Missing current_user import

khalilsarwari:
Not entirely sure if this will fix things, but we seem to have dropped this import in the process of cleaning code.


colinschoen:
I will review. Just a note I’m out for the next couple days so it may be delayed.

khalilsarwari:
Resolved by 9060112b703c8259bf9194030c3182f10f1c4533

PR #1237, opened 2018-03-24
Fix api.md formatting

jathak:
The Markdown here was invalid in some places, which lead to some weird formatting issues on the published page.

screenshot 2018-03-23 at 10 14 29 pm


jathak:
Also, does anyone know how to actually get these updated docs into the [documentation repo](https://github.com/okpy/documentation). @colinschoen?

PR #1241, opened 2018-04-09
Fixes migrations on sqlite by using absolute path

colinschoen:
If using `sqlite` and trying to run migrations, an additional `oksqlite.db` file was being added one directory higher than the root repo directory since we are using relative paths in the config. Since `sqlite` is the only file ordinary disk db that is being used this issue won't exist for other databases like `mysql` or `postgres`. Migrations still won't work `sqlite` due to limitations for commands like `ALTER`, but it is plausible that in the future `flask_migrate` will have library support for rewriting those commands into support `sqlite`. 

This PR rewrites paths in `sqlite` db URIs to absolute paths on the fly to ensure we are working on the same `.db` file regardless of the current working directory.

tldr: This fixes an issue most won't notice, but still should probably be merged. 


colinschoen:
CircleCi is timing out on the coverage report. This isn't failing any tests or failing to build.

PR #1242, opened 2018-04-16
Enable tests to run on Windows

c-w:
I checked out the repository, followed the setup instructions and then ran the tests. I saw some errors so I thought that there is something wrong with my setup. After a bit of digging, I found out that my setup of the repository is fine but that some of the tests assume a *nix runtime or specific running external services.

In order to prevent this confusion for future developers, I thought to fix the tests on Windows such that after following all the setup steps, a new developer would get a clean test run and that any tests which can't run on the developer's machine are [skipped](https://docs.python.org/3/library/unittest.html#skipping-tests-and-expected-failures) instead of failed.

Specifically, this pull request implements the following:
- 06b7f3b: Change documentation on how to run tests to use `manage.py` instead of `py.test`.
- 14f4a98: Skip tests that rely on *nix features like `fork` and `patch` when running on Windows.
- 8141fb6: Fix error due to default file encoding differences between Windows and *nix.
- b2c45f3: Skip tests that rely on Redis unless the service is running.

The pull request also contains one unrelated fix that I noticed while working:
- 623f48c: Remove an undefined variable.

For more detail on the changes above, please refer to the commit descriptions.

After applying this PR, following the setup and running the tests will lead to a clean test run on Windows:

Clean test run on Windows

Note that I do still get one error when running the tests (as seen in the screenshot above) but I also see that error when running the tests via Docker on the latest master (see screenshot below), so I'm assuming that this is a known issue and unrelated to my PR.

Test error on Docker

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ install:
   - if [ -n "$USE_NATIVE" ]; then pip install -r requirements.txt; fi
   - if [ -n "$USE_NATIVE" ]; then pip install pytest-timeout python-coveralls; fi
 script:
-  - if [ -n "$USE_DOCKER" ]; then docker run -it cs61a/ok-server py.test tests/ --ignore=tests/test_job.py; fi
+  - if [ -n "$USE_DOCKER" ]; then docker run -it cs61a/ok-server py.test tests/; fi
   - if [ -n "$USE_NATIVE" ]; then py.test --cov-report term-missing --cov=server tests/  --ignore tests/test_web.py --timeout=30; fi
 after_success:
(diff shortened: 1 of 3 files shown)


colinschoen:
Thanks for your work on this. Will review ASAP.

c-w:
Thanks for the super quick review, @colinschoen. Is there anything else I should be doing before you can merge this?

colinschoen:
Nope, this looks great. CircleCI is borked because it is being run from an outside fork instead of a local branch on this repo (tracking at #1243). Will merge in a moment.

colinschoen:
@jathak Actually can you merge, we have `master` protected so I can't even manually merge from the CLI. I think `owners` should be able to override that.

PR #1250, opened 2018-04-25
Transfer OAuth Client Ownership

### LINKED ISSUE #1249: Transfer OAuth Client Access
When #1245 is merged OAuth clients will not be able to be transferred or access assigned to another user. Admins or the OAuth owner should be able to transfer client ownership another staff member.

colinschoen:
Existing owner's of OAuth clients as well as admins can now transfer ownership by changing the owner email for the respective OAuth client.

This change attempts to respect the current "ownerless" state of some OAuth clients by allowing them to remain "ownerless" even during future client setting edits.

If the owner is changed however it is verified that the new owner does indeed exist and is a staff member of some course. (Considering if they weren't they would be unable to access the `/admin/client` page to manage the OAuth client.)

Includes three new test cases to cover for attempts at changing the owner email with a non-staff email, nonexistent account and with a valid staff email.

Preview:

image

Fixes #1249

### DIFF
--- a/server/controllers/admin.py
+++ b/server/controllers/admin.py
@@ -1351,12 +1351,17 @@ def client(client_id):
     courses, current_course = get_courses()
 
     client = Client.query.get(client_id)
+    # Show the client owner's email in edit form when owner exists
+    client.owner = client.user.email if client.user else ""
     form = forms.EditClientForm(obj=client)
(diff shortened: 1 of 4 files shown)


colinschoen:
Thanks for reviewing.

Issue #1255, opened 2018-05-01
TestApi.test_get_backup Flaking


colinschoen:
The `test_get_backup` test is flaking due to a time stamp comparison in the `created` field. As suggested in #1246 this could be fixed by adding a tolerance threshold to the timestamp comparison.

colinschoen:
This is flaking more often. Priority to fix next.

PR #1259, opened 2018-05-03
Refactoring the settings config to remove duplicates

liliankasem:
There are many shared settings between the environment setting files, this commit creates a base class that is shared across all environments; a prodbase class which contains settings shared across simple, staging and prod; and a devbase class which has settings shared across dev and test. The individual env setting files then inherit from the bases and overwrite as needed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,6 @@ _cache/*
 database.db
 *.db
 *.db-journal
+
+# Visual Studio Code
+.vscode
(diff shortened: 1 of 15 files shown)


colinschoen:
Thanks for the PR. I will review soon.

liliankasem:
@colinschoen thanks for pointing those out, changes made!

colinschoen:
Disregard test failure on travis. Noted as flake. Tracking at #1255 

Merging.

PR #1260, opened 2018-05-03
Add support for Azure MySQL

navjotk:
This pull requests adds support for using the MySQL PaaS provided by
Azure [1] as the backend for SQLAlchemy. Lighting up this feature
required two changes:

- Enabling SSL support in Flask-SQLAlchemy since Azure MySQL requires
  SSL by default.

- Enabling configuration of the MySQL row format since Azure MySQL does
  not support compressed rows.

The Azure MySQL backend can be enabled by setting the following
environment variables:

```
DATABASE_URL=mysql://{UserName}@{ServerName}:{Password}@{ServerName}.mysql.database.azure.com:3306/{DatabaseName}?
DB_ROW_FORMAT=DEFAULT
```

[1] https://azure.microsoft.com/en-us/services/mysql/


migrations/versions/94d2839daa08_.py
```diff
@@ -151,7 +151,7 @@ def upgrade():
     sa.Column('kind', sa.String(length=255), nullable=False),
     sa.ForeignKeyConstraint(['backup_id'], ['backup.id'], name=op.f('fk_message_backup_id_backup')),
     sa.PrimaryKeyConstraint('id', name=op.f('pk_message')),
-    mysql_row_format='COMPRESSED'
+    mysql_row_format=os.getenv('DB_ROW_FORMAT', 'COMPRESSED')
```
colinschoen: It would be nice to store `DB_ROW_FORMAT` in the `settings/*.py` files and use `app.config` here rather than `os.getenv`. This looks to be related: https://stackoverflow.com/questions/49409691/flask-sqlalchemy-acess-different-schema

colinschoen: I added this for the wrong  file. This comment is referring to the change in `models.py` line 962.

server/models.py
```diff
@@ -958,7 +959,7 @@ def create(cid, enrollment_info, role=STUDENT_ROLE):
 
 class Message(Model):
     __tablename__ = 'message'
-    __table_args__ = {'mysql_row_format': 'COMPRESSED'}
+    __table_args__ = {'mysql_row_format': os.getenv('DB_ROW_FORMAT', 'COMPRESSED')}
```
colinschoen: It would be nice to store `DB_ROW_FORMAT` in the `settings/*.py` files and use `app.config` here rather than `os.getenv`. This looks to be related: https://stackoverflow.com/questions/49409691/flask-sqlalchemy-acess-different-schema

Sumukh: +1 on storing on settings/*.py

c-w:
Rebased the settings refactor from master in 85ffddf.

server/models.py
c-w: Could you provide some guidance on how we'd achieve this? In the `models.py` file we don't have access to the `app`. I tried accessing the app via `flask.current_app` (see [current-app.patch.txt](https://github.com/Cal-CS-61A-Staff/ok/files/1974530/current-app.patch.txt)) but that gives me an error:

image

server/settings/_base.py
```diff
@@ -53,6 +53,11 @@ def initialize_config(cls):
             cls.SQLALCHEMY_DATABASE_URI = (sqlite_prefix +
                     os.path.abspath(cls.SQLALCHEMY_DATABASE_URI[len(sqlite_prefix) + 1:]))
 
+        sql_ca_cert = os.getenv('SQL_CA_CERT')
```
colinschoen: Should this fall back to `BaltimoreCyberTrustRoot.crt.pem` since this is being included in the repo now?

colinschoen:
Other than that LGTM. Tested and this is working properly. 👍

server/settings/_base.py
c-w: Good idea. Done and rebase the latest master.

c-w:
The builds keep failing due to intermittent issues so I added a retry in 21d0b74.

PR #1261, opened 2018-05-03
Add support for Azure Blobs storage backend

navjotk:
To enable the Azure Blobs storage [1] backend, set the following
environment variables:

```
STORAGE_PROVIDER=AZURE_BLOBS
STORAGE_KEY={AzureStorageAccountName}
STORAGE_SECRET={AzureStorageSecretKey}
STORAGE_CONTAINER=okpyfiles
```

Note that, just like the GCP storage backend, the Azure Blobs storage
backend assumes that the storage container already exists. This
assumption could be relaxed (by modifying the code to create a container
on the fly) since for Azure Storage, the account name and key are enough
to create a new container programmatically. However, this would conflict
with the current implementation of GCP storage since its permissions
don't allow for container creation.

There are two follow-up fixes that will be contributed in the future:

- Moving the logic to create the signed access signature into libcloud
  and removing the dependency on the azure-storage module from OK.
  @martinpeck will work on the libcloud feature pull request.

- Re-enable the `test_simple` after the `download_object_as_stream`
  method is fixed for Azure Blobs in libcloud. Note that this test
  is currently only meaningful for the local storage provider so the
  current disabling the test is not a risk for OK. @jmspring will
  work on the libcloud fix.

[1] https://azure.microsoft.com/en-gb/services/storage/blobs/

### DIFF
--- a/server/storage.py
+++ b/server/storage.py
@@ -11,6 +11,7 @@ from urllib.parse import urlencode, urljoin
 from werkzeug.utils import secure_filename
 from libcloud.storage.types import Provider
 from libcloud.storage.providers import get_driver
+from azure.storage.blob import BlockBlobService, BlobPermissions
 
 logger = logging.getLogger(__name__)
 
@@ -108,12 +109,12 @@ class Storage:
         """
         driver_name = self.driver.name.lower()
         if 's3' in driver_name:
-            base_url = 'https://%s'.format(self.driver.connection.host)
+            base_url = 'https://{}'.format(self.driver.connection.host)
             url = urljoin(base_url, object_path)
         elif 'google' in driver_name:
             url = urljoin('https://storage.googleapis.com', object_path)
         elif 'azure' in driver_name:
-            base_url = 'https://%s.blob.core.windows.net'.format(self.driver.key)
+            base_url = 'https://{}.blob.core.windows.net'.format(self.driver.key)
             url = urljoin(base_url, object_path)
         else:
             raise Exception('Unsupported Storage Driver for URL')
@@ -126,14 +127,25 @@ class Storage:
         if container_name is None:
             container_name = self.container_name
         driver_name = self.driver.name.lower()
-        expires = (dt.datetime.now() + dt.timedelta(seconds=timeout))
-        expires = int(expires.timestamp())
+        expiry_timestamp = (dt.datetime.now() + dt.timedelta(seconds=timeout)).timestamp()
 
         if 's3' in driver_name or 'google' in driver_name:
             keyIdName = "AWSAccessKeyId" if "s3" in driver_name else "GoogleAccessId"
+            return self._generate_google_aws_signed_url(keyIdName, obj_name, expiry_timestamp)
+        elif 'azure' in driver_name:
+            return self._generate_azure_signed_url(obj_name, expiry_timestamp)
         else:
             raise Exception('{0} does not support signed urls'.format(driver_name))
 
+    def get_object_stream(self, object):
+        """ Data stream of the libcloud object.
+        """
+        return self.driver.download_object_as_stream(object)
+
+    def _generate_google_aws_signed_url(self, keyIdName, obj_name, expiry_timestamp):
+        """ Generates a signed URL compatible with Google Cloud and AWS S3
+        """
+        expires = int(expiry_timestamp)
         obj_path = "{0}/{1}".format(self.container.name, obj_name)
         s2s = ("GET\n\n\n{expires}\n/{object_name}"
                .format(expires=expires, object_name=obj_path).encode('utf-8'))
@@ -148,7 +160,24 @@ class Storage:
         secure_url = self.get_url(obj_path)
         return "{0}?{1}".format(secure_url, url_kv)
 
-    def get_object_stream(self, object):
-        """ Data stream of the libcloud object.
+    def _generate_azure_signed_url(self, obj_name, expiry_timestamp):
+        """ Generates a signed URL compatible with Azure Blob Storage
         """
-        return self.driver.download_object_as_stream(object)
+        # libcloud encodes the access key, so it needs decoding
+        secret_as_string = base64.b64encode(self.driver.secret).decode()
+
+        # azure expiry times need to be ISO8601 (no milliseconds, and with the timezone character)
+        expires = dt.datetime.utcfromtimestamp(expiry_timestamp).replace(microsecond=0).isoformat() + 'Z'
+
+        blob_service = BlockBlobService(self.driver.key, secret_as_string)
+
+        signed_url = blob_service.make_blob_url(
+            self.container.name,
+            obj_name,
+            sas_token=blob_service.generate_blob_shared_access_signature(
+                self.container.name,
+                obj_name,
+                permission=BlobPermissions(read=True),
+                expiry=expires))
+
+        return signed_url
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 FROM python:3.5-alpine
 
-RUN apk add --update patch ca-certificates nginx perl;
+RUN apk add --update patch ca-certificates nginx perl musl-dev openssl-dev libffi-dev python-dev gcc;
 
 RUN mkdir /code/
 WORKDIR /code/
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,8 +10,9 @@ RUN pip3 --timeout=60 install --no-cache-dir -r requirements.txt
 
 ADD . .
 
-RUN mv docker/nginx/nginx.conf /etc/nginx/nginx.conf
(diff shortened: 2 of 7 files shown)


colinschoen:
Thanks for the contribution. 

This will be difficult for me to immediately test as we will need to set up an organizational Azure account for OK. Have you run the integration tests locally for the new Azure Blob Storage backend?

In the meantime I will work on getting one setup.

c-w:
@colinschoen I can send you some credentials for testing. Will reach out in Slack.

colinschoen:
Thanks Clemens. 

I am getting the following test output. It looks like 4 tests are erroring
[test_output.txt](https://github.com/Cal-CS-61A-Staff/ok/files/1971729/test_output.txt)
.

c-w:
How are you running the tests? They work for me with a clean virtualenv:

```cmd
set AZURE_STORAGE_CONTAINER=...
set AZURE_STORAGE_KEY=...
set AZURE_STORAGE_SECRET=...

python3 -m venv venv
venv\Scripts\activate
pip install -r requirements.txt

python manage.py test
```

Screenshot of the output is below (after modifying `manage.py` to only run `test_files_cloud.py`):

Screenshot of test run

colinschoen:
Here are the steps:

1. Cloned the repo
2. Fetched and checked out the pull request branch
3. Created a fresh virtual env
4. Installed the pip dependencies
5. Set the environment variables
6. Modified `manage.py` to only run `test_files_cloud.py`
7. Ran `python3 manage.py test`. 

Bash history and test output attached. 
[test_output.txt](https://github.com/Cal-CS-61A-Staff/ok/files/1972115/test_output.txt)

colinschoen:
Odd, I ran them at about the time I made the comment so that would be around 

`2018-05-03T6:00:00Z`

Running `date` now is giving:

`Thu May  3 14:13:35 PDT 2018`

Which seems correct.

Let me try running the test suite again and see if I am still getting strangeness in the datetimes.

server/storage.py
```diff
@@ -148,7 +160,24 @@ def get_blob_url(self, obj_name, container_name=None, timeout=60):
...
-        """ Data stream of the libcloud object.
+    def _generate_azure_signed_url(self, obj_name, expiry_datetime):
+        """ Generates a signed URL compatible with Azure Blob Storage
         """
-        return self.driver.download_object_as_stream(object)
+        # libcloud encodes the access key, so it needs decoding
+        secret_as_string = base64.b64encode(self.driver.secret).decode()
+
+        # azure expiry times need to be ISO8601 (no milliseconds, and with the timezone character)
+        expires = expiry_datetime.replace(microsecond=0).isoformat() + 'Z'
```
colinschoen: It looks like here it is assuming Zulu or UTC time zone when local time is being passed to `expiry_datetime`.

martinpeck:
OK. I'll take a look at this tomorrow (UK time). 

Those times look wrong to me, because it's currently ( as I type) 21:25 UTC. Your date looks like Pacific Time with "Z" on the end, incorrectly labelling it as Universal Coordinated Time.

I <3 timezones!

martinpeck:
almost certainly - let me review it tomorrow, and we'll get this PR updated
sorry for mucking you about

colinschoen:
No worries! Thanks for your work on this.

c-w:
Worker has been stable for >1 hour so looks like the fix worked.

image

colinschoen:
Hmm I am still getting failures

[test_output.txt](https://github.com/Cal-CS-61A-Staff/ok/files/1976118/test_output.txt)

martinpeck:
@c-w: is this related to the config updates we addressing yesterday? The "expected 302, got 200" errors were certainly being seen with the config issues. Have we issued a PR to address the config stuff?

c-w:
@martinpeck: Yes, we did submit a PR for the config changes (#1264) and it was merged in 5b13a02. I rebased this branch to include those changes. I send you the credentials in Slack.

c-w:
NB: I just ran the tests with a totally clean build in Docker with the latest version of the branch (which includes the config fix). All tests seem to pass on my system.

I noticed that due to Docker caching the requirements would sometimes not be rebuilt so I issued a `docker system prune -a` beforehand to ensure that all the Docker image layers will be up to date.

```sh
# ensure that all layers of the image are re-built and we don't run into
# any odd issues with requirements not being installed correctly
docker system prune -a
make docker-build

# same command as `make docker-test` but activating the Azure storage tests
# but with the fix from https://github.com/Cal-CS-61A-Staff/ok/pull/1266
# applied to ensure that the compose command uses the image created by the
# earlier `make docker-build` command
docker-compose \
  -f docker/docker-compose.yml \
  -f docker/docker-compose.test.yml \
  run \
  -e AZURE_STORAGE_CONTAINER=redacted \
  -e AZURE_STORAGE_KEY=redacted \
  -e AZURE_STORAGE_SECRET=redacted \
  --rm \
  web \
  py.test --cov-report term-missing --cov=server tests/test_files_cloud.py
```

Screenshot of clean test run in Docker

Oddly, when I run the same sequence of tasks natively on Ubuntu on WSL, I get failures:

```sh
python3.5 -m venv ./venv
./venv/bin/pip install -U pip setuptools
./venv/bin/pip install -r requirements.txt

./venv/bin/python ./manage.py createdb

export AZURE_STORAGE_CONTAINER=redacted
export AZURE_STORAGE_KEY=redacted
export AZURE_STORAGE_SECRET=redacted

./venv/bin/python ./manage.py test
```

Screenshot of test errors on WSL

The question then is: what is the difference between the docker environment and the native environment and how could this affect the tests?

c-w:
Okay, I think I figured this one out. `make docker-test` is using `py.test` inside of the container to run the tests. `python manage.py test` on the other hand is using the standard library unittest runner instead. For some reason, the latter seems to fail some of the tests... When I run `py.test` natively in my virtualenv on WSL, the tests pass.

```sh
./venv/bin/py.test --cov-report term-missing --cov=server tests/test_files_cloud.py
```

Screenshot of clean test run using py.test

So maybe the fix here is to move the `manage.py test` command to use the `py.test` runner instead of the unittest one. This will unfortunately fail some of the tests on Windows (see https://github.com/Cal-CS-61A-Staff/ok/pull/1242 for context) but it'll eliminate the differences between test runs in the container and natively. Thoughts?

colinschoen:
I suppose I am more confused why runs invoked via `py.test` would be successful and the standard `unittest` library wouldn't. Hmm. @Sumukh Do you have any context about why the container is set to use py.test and `manage.py` is using `unittest`?

c-w:
This is getting weirder and weirder. Replacing the unittest runner command in `manage.py` with a pytest invocation ([diff](https://github.com/Cal-CS-61A-Staff/ok/files/1979881/pytest-runner.patch.txt)) also fails with the same error. So maybe the issue is running the tests from inside of `manage.py` instead of unittest vs pytest.

c-w:
I've confirmed that running the tests without `manage.py` consistently works (both on Windows and Ubuntu), so I've removed the `manage.py test` command in 68b180a and updated the documentation to recommend running tests via `python -m unittest`.

I think this might be the most pragmatic way to fix this issue and avoid us going down the rabbit-hole of investigating the flask-script/pytest integration. Thoughts, @colinschoen?

colinschoen:
@c-w Yes, this seems reasonable. Thanks. Will retest.

colinschoen:
Still getting the failures when running with `python3 -m unittest`.

[test_output.txt](https://github.com/Cal-CS-61A-Staff/ok/files/1982031/test_output.txt)

c-w:
I think I finally figured this one out (see fix in 28b0ab0).

The root cause of the confusion is due to the change from having the settings previously in python modules to now having them in classes. The settings classes look up the values for the class variables at first import time and don't reload them for subsequent app creations since the Flask app is just referencing an existing class instance for loading of the settings. This means that the previous way to switch out settings for a particular cloud file test by swapping out the environment variables doesn't work reliably and depends on which test gets executed first (which explains the differences in test results between the various test runners and between the various development machines). I've now changed the settings overrides to directly modify the settings object which should be much more reliable.

Please let me know if the tests now pass for you, @colinschoen.

c-w:
Still trying to get a clean build on this due to all the intermittent test failures. Got Travis to pass but CircleCI timed out.

In order to speed things up, I pushed a change that should drastically decrease the execution time of the docker test suite in 2cc4133 (on my machine it dropped from 13 minutes to just under 2 minutes).

c-w:
@colinschoen Could you please take another look at this so that we can close this off? Thanks in advance!

Sumukh:
Looks ok to me, @colinschoen LMK if you want me to review if you are short on time this week.

colinschoen:
Apologies for delay. Still in the midst of a move. Taking a look now

PR #1262, opened 2018-05-03
Compares timestamps with assertAlmostEqual

### LINKED ISSUE #1255: TestApi.test_get_backup Flaking

colinschoen:
The test_get_backup test has been flaking due to
doing equality comparisons on timestamps before
and after database transactions. These comparisons
have been changed to ensuring the timestamps are
within a certain threshold (currently 5 seconds).

Fixes #1255


colinschoen:
Self approving since this is causing issues.

PR #1263, opened 2018-05-04
Add support for Azure logging

navjotk:
Currently the cloud-based logging uses Google Stackdriver and Sentry.
This commit adds support for an alternative cloud-based logging system
based on Azure Application Insights [1]. Metrics about all requests, all
stack traces for exceptions and all manual log traces are sent to
Application Insights. Alerting rules, dashboards showing performance
metrics, and so forth can then be set up on top of this data via the
Azure Portal.

To enable Application Insights logging and disable Stackdriver and
Sentry, set the following environment variable:

```
APPLICATION_INSIGHTS_KEY={InstrumentationKey}
```

[1] https://azure.microsoft.com/en-gb/services/application-insights/


server/logging/gunicorn.py
```diff
@@ -126,3 +126,7 @@ def access(self, resp, req, environ, request_time):
 
         struct_pb = ParseDict(message, Struct())
         self.cloud_logger.log_proto(struct_pb, severity=severity)
+
+
+if os.getenv('APPLICATION_INSIGHTS_KEY'):
```
colinschoen: Should this be `APPLICATION_INSIGHTS_KEY` or `APPINSIGHTS_INSTRUMENTATIONKEY`?

c-w: Good catch. This should be `APPINSIGHTS_INSTRUMENTATIONKEY` as per https://github.com/Microsoft/ApplicationInsights-Python/pull/79. Fixed and also rebased latest master.

PR #1271, opened 2018-05-07
Edit Extensions

### LINKED ISSUE #1240: Edit Extensions
Extensions can currently only be created or deleted. There is a use case where the length of an extension changes and it would be nice to be able to modify this without having to remove and recreate the extensions.

colinschoen:
- Adds functionality to edit existing extensions (0065fdd)
 - Adds missing test coverage for creating extensions (d7f73c5)
 - Adds test coverage for new edit extension functionality (e210ab6)

Closes #1240


colinschoen:
@jathak Can you PTAL when able. This should be good to go.

PR #1276, opened 2018-05-16
Add Support for Redis connection URLs

martinpeck:
This pull request adds support for Redis connection URLs such as...

`rediss://:reallystrongpassword@myazureredisinstance.redis.cache.example.net:6380/0`

This change allows you to configure Redis with a single environment variable, but more importantly it allows secure SSL connections, that are then used by Flask-Cache and RQ, through support of the `rediss://` url scheme.

(for more details, see <https://www.iana.org/assignments/uri-schemes/prov/rediss>)

The changes are relatively simple. The Redis configuration has been refactored, and uses the environment variable `REDIS_URL` if present. Otherwise everything works as it always did with the only change being that `REDIS_PORT` is now a configuration item.

I've pulled almost all of the Redis configuration up into `_base.py` and then use overrides in `prod.py`.

I've also updated the SETUP.md documentation to include a section on configuring Redis. I'm happy to make this clearer if needed.

### DIFF
--- a/documentation/SETUP.md
+++ b/documentation/SETUP.md
@@ -26,7 +26,7 @@ Run tests via
 
     $ ./manage.py test
 
-In order to run the tests in `tests/test_web.py`, `phantom.js` needs to be installed. 
+In order to run the tests in `tests/test_web.py`, `phantom.js` needs to be installed.
 On OSX: First make sure [Homebrew](http://brew.sh/) is installed. To install `phantom.js`:
 
     $ brew install phantomjs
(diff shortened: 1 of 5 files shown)


marrobi:
@colinschoen @martinpeck whats left on this one? It's the last PR that modifies the Ok.py codebase, so be good to get it merged. Then we have the docs and deployment templates (which depend on this PR) and will have the full solution in in the repo.

colinschoen:
This is fine to merge. I was waiting for https://github.com/Cal-CS-61A-Staff/ok/pull/1275 but we can merge this first.

martinpeck:
Thanks @colinschoen!

PR #1280, opened 2018-05-25
Fixing the issue where OAUTH_PROVIDER is not being set properly

liliankasem:
Fixing the issue where OAUTH_PROVIDER was always set to 'GOOGLE' (the default value) because the config variable was not being properly setup.


liliankasem:
Still failing on `tests/test_extension.py:38: in _submit_to_api self.assert_403(response)`; not sure how to fix that - on my last PR I just kept triggering a build till it was green 🤔

PR #1292, opened 2018-06-25
Protect against set but empty OAuth variables

c-w:
Currently we're only checking whether the variables like `GOOGLE_ID` or `MICROSOFT_APP_SECRET` are set as environment variables but we aren't validating whether the variables are non-empty. This means that the
server start might be successful but we'll fail later at login time which is harder to debug. This change fixes that behavior and prevents server start if the OAuth variables are set but contain empty values.


c-w:
@liliankasem @cicorias FYI

liliankasem:
@c-w thanks for cleaning up!

PR #1293, opened 2018-06-26
Make score_amount comparison stricter

### LINKED ISSUE #1277: Manual score upload does not allow scores of 0
Not critical, since a missing assignment is the same as scoring 0, but something I noticed while uploading final exams this semester

colinschoen:
`score_amount` could potentially be 0 causing the validator to fail
We not explicitly check if `score_amount is not None`.

Fixes #1277


jathak:
Thanks

PR #1294, opened 2018-06-26
Parameterize all uses of the autograder URL

c-w:
This issue was found and reported by @taupalosaurus: the autograder URL can now be over-written with an environment variable so all places where the URL is used should reflect that.


colinschoen:
Have you updated `constants.py`?

PR #1297, opened 2018-06-27
Add sendgrid username/password authentication

c-w:
This change fixes an issue where the Azure PaaS deployments were unable
to send mail via Sendgrid.

The issue was caused by the fact that the Sendgrid API assumes
authentication via an API key, but in the ARM template we only have
access to the Sendgrid account admin username and password. Due to the
limitations of the ARM template format, it's difficult to automatically
generate the API key outside of the OKpy application during the
deployment: we'd have to spin up an Azure Container Instances
deployment, generate the key in the container, upload it to a storage
like Key Vault and then access the key from the OKpy application.
Reading configuration from a storage like Key Vault instead of
environment variables would be a non-trivial change to OKpy so this
commit implements a more pragmatic work-around: generate a new API key
on the fly at runtime given the Sendgrid admin credentials. The key is
cached for the lifetime of the server so this shouldn't add overhead
past the first email sent.

Note that the Sendgrid API has a limit on 100 API keys so we always
ensure to auto-create at most one API key for OKpy and delete any old
copies of the key before creating a new one (this would happen when the
OKpy server is restarted since any auto-created API key only is
persisted in the server memory).

Another nice property of this change is that it's now possible to
authenticate to Sendgrid in OKpy in the same way as we do in AutoPY.

### DIFF
--- a/server/utils.py
+++ b/server/utils.py
@@ -2,11 +2,11 @@ import csv
 import datetime as dt
 import logging
 from io import StringIO
-import os
 import random
 import re
 import urllib
 from urllib.parse import urlparse, urljoin
+from functools import lru_cache
 
 import bleach
 from flask import render_template, url_for, Markup
@@ -17,6 +17,7 @@ import markdown
 from pynliner import fromString as emailFormat
 import pytz
 import requests
+from requests.auth import HTTPBasicAuth
 import sendgrid
 import sendgrid.helpers.mail as sg_helpers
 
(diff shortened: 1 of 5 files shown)


c-w:
CC @marrobi since this affects the ARM templates.

server/utils.py
```diff
@@ -170,13 +170,60 @@ def send_emails(recipients, subject, body, **kwargs):
...
+    key_name = 'OKpy'
+    key_scopes = ['mail.send']
+    api_keys_root = 'https://api.sendgrid.com/v3/api_keys'
+    auth = HTTPBasicAuth(sg_username, sg_password)
+
+    response = requests.get(api_keys_root, auth=auth)
+    if not response.ok:
+        raise ValueError('unable to list sendgrid api keys')
+
+    existing_key = next((key['api_key_id']
```
colinschoen: Nice :)

PR #1298, opened 2018-06-27
Create custom docker image for worker

c-w:
AppService for Containers seems to have an issue with the custom run
command for the Docker image. According to the logs, the worker's
AppService was starting an instance of the OK server instead of the
worker. As a simple work-around, this change introduces a Dockerfile for
the OK worker which bakes the correct CMD into the image. This does
create some level of redundancy, but on the bright side it also means
that in the future we could publish a more optimized image for the
worker, e.g. excluding the frontend assets, etc.

Also: pass-through auto-grader URL in ARM template

### DIFF
--- a/Dockerfile-worker
+++ b/Dockerfile-worker
@@ -1,0 +1,3 @@
+FROM cs61a/ok-server
+
+CMD ["./manage.py", "worker"]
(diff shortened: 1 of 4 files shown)


c-w:
Going to close this since we faced a few issues with running workers in AppService. Going to open a new pull request soon which moves the compute from AppService to Azure Kubernetes Service.

PR #1299, opened 2018-06-29
Remove hard-coded links to static files

c-w:
This change replaces all direct links to static files (such as
`/static/foo/bar`) with their equivalent `url_for` call. This introduces
consistency with code that creates links to the app's Flask routes and
also ensures that the links to the static files will always be correct
even if the application is for example hosted at a non-root path.


c-w:
Thanks for the review @colinschoen and the CI fixes. Build now passes.

PR #1326, opened 2018-09-29
Fix datetimepicker to choose from 24hr

### LINKED ISSUE #1328: The date-time picker for assignments won't let me select pm
I can't edit the deadline for assignment to be 11:59pm -- it keeps reverting my entry to 11:59am.  I've tried Safari, Chrome, and Firefox on Mac OS X, and am not successful with any of them.  I have tried using both the pop-up picker, and also tried manually entering in 23:59:59 into the textbox, and neither works: when I try to leave that field or save, it gets reverted to 11:59:59, which refers to pm.

Austen tells me that he had the same problem with Chrome but by using Firefox he was able to set it to 11:59pm, but I wasn't able to reproduce that on my computer; neither browser works for me.

itsnshaw:
- [ ] Glyphicons also fail to load on browser

Fixes okpy/ok-client#351
Fixes #1328

### DIFF
--- a/server/static/js/main.js
+++ b/server/static/js/main.js
@@ -52,5 +52,5 @@ function debounce(func, wait, immediate) {
 };
 
 $('.datetime-picker').datetimepicker({
-  format: 'YYYY-MM-DD hh:mm:ss'
+  format: 'YYYY-MM-DD HH:mm:ss'
 });


davidwagner:
Looks good to me!

PR #1331, opened 2018-10-11
PR has been merged can now use upstream webassets

colinschoen:
webassets has merged the PR in question https://github.com/miracle2k/webassets/pull/482.

We can now go back to using the pypi version of this package.


Sumukh:
Are you sure - last I checked a few months ago for another project - the package wasn’t actually published?

colinschoen:
I didn’t look closely enough. I guess they aren’t supporting it any more? Reverting

PR #1340, opened 2018-12-09
Fix course list link

MaxWofford:
This fixes a typo that doesn't escape the course list link. Here's what the link used to do:

screenshot 2018-12-08 22 59 35
screenshot 2018-12-08 22 59 45


Sumukh:
Thanks!

PR #1345, opened 2019-01-15
Navigate to index page after OK logo clicked in staff mode

### LINKED ISSUE #1344: "OK" Logo in Admin Mode should navigate to the main index page
Currently, clicking the "OK" logo in the top left corner of the admin mode (or staff mode, not sure about terminology), will navigate the user to the _assignments page of the first course for which the user is an admin (staff)._ This feels very strange to me. Clicking the "OK" logo should navigate to either  the list of all courses (https://okpy.org/admin/course/) or the okpy student mode index page (https://okpy.org/). 

screen shot 2019-01-15 at 12 01 06 am

Royserg:
Instead of getting assignments page, navigate to the student mode index page.
My first contribution to OS, not sure if I did it right 🤞🤞

fix #1344

### DIFF
--- a/server/templates/staff/navbar.html
+++ b/server/templates/staff/navbar.html
@@ -42,7 +42,11 @@
 <nav class="navbar navbar-static-top" role="navigation">
   <div class="container">
     <div class="navbar-header">
-      <a href="{{ url_for('admin.index') }}" class="navbar-brand">
+      {% if current_user.is_admin %}
+        <a href="{{ url_for('admin.index') }}" class="navbar-brand">
+      {% else %}
+        <a href="{{ url_for('student.index') }}" class="navbar-brand">
+      {% endif %}
         <b>ok</b>
       </a>
       <button type="button" class="navbar-toggle collapsed" data-toggle="collapse" data-target="#navbar-collapse">


server/controllers/admin.py
```diff
@@ -129,9 +129,8 @@ def get_courses(cid=None):
 @admin.route("/")
 @is_staff()
 def index():
-    courses, current_course = get_courses()
-    if courses and not current_user.is_admin:
-        return redirect(url_for(".course_assignments", cid=courses[0].id))
+    if not current_user.is_admin:
```
colinschoen: Admins are a small subset of users that can have super user privileges on ok. We still want instructors and TAs to end up in the either the "instructor" view of their course. Furthermore, the only folks which will be able to access this route are staff because the `@is_staff` decorator is used. 

You should be changing where the `OK` logo links to rather than modifying the `index` controller.

Royserg: @colinschoen Thank you for more information,
Second attempt :)

I modified only `staff/navbar.html` 
users who are not admins will get link to the `student.index` page

Was this a way to solve it or am I still confusing something?

server/templates/staff/navbar.html
```diff
@@ -42,7 +42,11 @@
 <nav class="navbar navbar-static-top" role="navigation">
   <div class="container">
     <div class="navbar-header">
-      <a href="{{ url_for('admin.index') }}" class="navbar-brand">
+      {% if current_user.is_admin %}
+        <a href="{{ url_for('admin.index') }}" class="navbar-brand">
+      {% else %}
+        <a href="{{ url_for('student.index') }}" class="navbar-brand">
+      {% endif %}
```
pbitutsky: If the current user is an admin, isn't the url for student.index the same as the url for admin.index? In other words, isn't this entire code block the same as the following? 

```
<a href="{{ url_for('student.index') }}" class="navbar-brand">
```

colinschoen: I don't think so. Those are two entirely different routes.

colinschoen:
The issue @pbitutsky filed suggests one of the places the header can link to is the student index. I don't think this is super useful. I think we should just link to the list of courses (in the staff view) for everyone (not just admins).

server/templates/staff/navbar.html
pbitutsky: I'm running a local server and when I impersonate a random non-admin user, I can view both 
http://localhost:5000/ and http://localhost:5000/admin/course/. But when I'm impersonating an admin, http://localhost:5000/ (aka student.index) seems to redirect to http://localhost:5000/admin/course/ (admin.index) automatically.

colinschoen: Is that true even if you enroll the admin as a student in a course?

pbitutsky: Yes

colinschoen: Ok, disregard then.

pbitutsky:
@colinschoen If I'm understanding you correctly, you would like for the OK logo to navigate to http://localhost:5000/admin/course/. However, there is already navigation to get there: in the navbar, under courses > View All Courses. However, there is no way to navigate to the student index (http://localhost:5000/) via the UI. So won't it be better if the logo were to navigate to the student index?

colinschoen:
@pbitutsky Ah, ok that does sound ok too. Yeah navigating to the student index makes sense.

server/templates/staff/navbar.html
Royserg: @pbitutsky @colinschoen 
I think that replacing that if else block with link to student.index only won't do the trick
`<a href="{{ url_for('student.index') }}" class="navbar-brand">`

If only link to `student.index` would stay, then an user who is an admin(has **is_admin** true) by clicking on `OK` logo will be redirected to admin dashboard of a first course they have on the list. Even on the all courses view clicking logo would redirect to that one particular course and it feels weird. Little demo below. Logged in as: **chelseacurry9@berkeley.edu**

admin redirected to one of the courses

In case of those users who are also a staff for a course, they will get to the `student.index` but staff edition (staff dashboard). Logged in as: **ctf3@gmail.com**

staff_logged

And finally normal student (not a staff for any course) will always get redirected to the `student.index` and is not able to access admin dashboard
Logged in as: **gzg@gmail.com**

image

Please, let me know if I understood it correctly and if any more changes are neccessary :)

pbitutsky: In that case, this PR LGTM.

Issue #1274, opened 2018-05-14
Edit enrollment information


papajohn:
The UI for editing an enrolled student's info could be quicker. For example, if a known student has no SID listed, and that SID must be added. 

Current UI:
- Look up the student using search
- Open the Add/Edit Participant panel
- Manually copy name and email to the Add/Edit Participant panel
- Copy in the SID
- Click "enroll"

Better UI:
- Look up the student using search
- Click an "edit" button to open and auto-populate the Add/Edit Participant panel
- Copy in the SID
- Click "enroll"

PR #1346, opened 2019-01-16
Add form in Student View for editing enrollment information

pbitutsky:
This resolves issue #1274.

I've added a form to the student view that allows admin/staff to edit information about students like their name, SID, secondary auth, section, and role. Users cannot, however, change students' email addresses since much of the back-end code relies on using the email as an identifier. Maybe this could be a feature in the future. 

screen shot 2019-01-16 at 12 30 51 am

screen shot 2019-01-16 at 12 30 55 am

screen shot 2019-01-16 at 12 31 37 am

screen shot 2019-01-16 at 12 31 45 am

### DIFF
--- a/server/controllers/admin.py
+++ b/server/controllers/admin.py
@@ -1448,10 +1448,10 @@ def student_view(cid, email):
         else:
             email, role = form.email.data, form.role.data
             Enrollment.enroll_from_form(cid, form)
-            flash("Edited User Successfully", 'success')
+            flash('Edited User Successfully', 'success')
     else:
         if form.is_submitted():
-            flash("There is an issue with your request.", 'error')
+            flash('There is an issue with your request.', 'error')
 
     courses, current_course = get_courses(cid)
     assignments = current_course.assignments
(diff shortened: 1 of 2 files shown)


colinschoen:
Awesome work. I’ll review tomorrow.

server/controllers/admin.py
```diff
@@ -1437,9 +1437,22 @@ def client(client_id):
...
 def student_view(cid, email):
+
+    form = forms.EnrollmentForm()
+    if form.validate_on_submit():
+        if form.email.data != email:
+            flash("You cannot change the user's email.", 'error')
+        else:
+            email, role = form.email.data, form.role.data
+            Enrollment.enroll_from_form(cid, form)
+            flash("Edited User Successfully", 'success')
```
colinschoen: Nit, prefer single quotes unless a single quote is used in the string. (I know this isn't consistent everywhere).

server/controllers/admin.py
```diff
@@ -1437,9 +1437,22 @@ def client(client_id):
...
+    else:
+        if form.is_submitted():
+            flash("There is an issue with your request.", 'error')
```
colinschoen: same as above

colinschoen:
This might also be a nice opportunity for a test, but I won't hold you to it if you want to get it shipped now :).

PR #1351, opened 2019-04-02
Modify edit student form to allow email editing

### LINKED ISSUE #1229: Change Student Email
Sometimes students sign up with a non berkeley email and want to use it later. Transferring their scores is quite painful now. It would be nice to be able to change the email associated with an account.

ja5087:
fixes #1229 

I've enabled the ability to edit emails from within the "Edit Student Form" within each course's enrollment page. Since email and name are shared between all courses (and perhaps SID should be too), this change is reflected across every single course the student is enrolled in.

### DIFF
--- a/server/controllers/admin.py
+++ b/server/controllers/admin.py
@@ -1445,7 +1445,7 @@ def student_view(cid, email):
     form = forms.EnrollmentForm()
     if form.validate_on_submit():
         if form.email.data != email:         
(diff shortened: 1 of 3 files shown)


pbitutsky:
Problem: Malicious users can enroll an admin into their class and change the admin's email. By doing that, the admin will not be able to log in to their okpy account. 

Please make it so that users cannot change admin's emails.

PR #1353, opened 2019-04-04
Add ability to delete backup/submissions

ja5087:
This adds a new "delete" button the submissions view for each student along with backend code.

Next steps are to make the button pop an alert first (to avoid accidental clicks), and to test it thoroughly, seeing as this is destructive.


ja5087:
Unfortunately, since the SQL schema doesn't have cascades set up to support this, this feature will likely require a database migration.

PR #1359, opened 2019-04-22
When editing student or staff, default to their current role, not Student

### LINKED ISSUE #1358: Edit Participant window Role defaults to Student
As an instructor, when I use Edit Participant window, my Role defaults to Student.

Unless I change the dropdown to Instructor each time, I will unintentionally change my Role to Student
Sang Yun Oh   PSTAT134 234  Spring 2019

pbitutsky:
This resolves #1358. 

I've tried many different approaches to setting the Role select's value and nothing has worked except adding javascript. It is not ideal and I'm not proud of this PR, but it works.

### DIFF
--- a/server/controllers/admin.py
+++ b/server/controllers/admin.py
@@ -1480,6 +1480,10 @@ def student_view(cid, email):
     if not enrollment:
         flash("This email is not enrolled", 'warning')
 
+    # Change the value in the "role" select to not default to STUDENT_ROLE
+    form.role.default = enrollment.role
+    form.process()
+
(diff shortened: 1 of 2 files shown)


ja5087:
You can pass in data via the WTForms constructor i.e. 

`form = forms.EnrollmentForm(role=enrollment.role)`

That probably requires moving some code around and making sure the logic still works.

pbitutsky:
@ja5087, I was not able to pass values through the constructor but found a better way to do this. Please see new changes.

PR #1375, opened 2019-12-14
Include submission times in grade book exports

cycomachead:
This is helpful when trying to calculate slip days.

I think this works, but dunno much about okpy. Written based on just reading the code. :)

### DIFF
--- a/server/jobs/export_grades.py
+++ b/server/jobs/export_grades.py
@@ -44,16 +44,18 @@ def get_headers(assignments):
     new_assignments = []
     for assignment in assignments:
         new_headers = ['{} ({})'.format(assignment.display_name, score_type.title()) for
-                score_type in get_score_types(assignment)]
+                score_type in get_score_types(assignment)
         if new_headers:
             new_assignments.append(assignment)
             headers.extend(new_headers)
+            headers.append('{} (Submitted At)'.format(assignment.display_name))
     return headers, new_assignments
 
 def export_student_grades(student, assignments):
     student_row = [student.user.email, student.sid]
     for assign in assignments:
         status = assign.user_status(student.user)
+        submission_time = status.subm_time
         scores = {s.kind.lower(): s.score for s in status.scores}
         scores = score_policy(scores)
         score_types = get_score_types(assign)
@@ -62,6 +64,10 @@ def export_student_grades(student, assignments):
                 student_row.append(scores[score_type])
             else:
                 student_row.append(0)
+        if submission_time:
+            student_row.append(submission_time)
+        else:
+            student_row.append('No Submission')
     return student_row
 
 


kavigupta:
this looks good to me but @rahularya50 how does this interact with your change for a faster export?

cycomachead:
Thanks -- also if it's useful to others, I could try to format this as a lateness calculation -- the result in HH:MM:SS.

rahularya50:
Sorry for the delay. Unfortunately, I don't think this PR interacts well with #1376, since `export_student_grades` only gets access to `all_scores`, not the assignment status object from the ORM. I'll see if I can modify my PR to support this functionality.

cycomachead:
No problem. 

Hmmm, #1376 uses the DB more... could we use TIMESTAMPDIFF to get the difference right in that query and then here convert it to HH:MM:SS

https://dev.mysql.com/doc/refman/5.5/en/date-and-time-functions.html#function_timestampdiff

I’m willing to help figure that out.

rahularya50:
In some cases, the assigned score comes from multiple submissions. For instance, the `total` score may come from one backup, but the `composition` score from another. Currently, your implementation returns the submission time of the earliest backup. Should I keep this behavior?

rahularya50:
Also, I'm pushing a test implementation of this to the other PR now.

cycomachead:
I think the earliest backup time makes sense, or at least the earliest of the ones that’s graded

Thank you!

-- 
Michael Ball
From my iPhone
michaelball.co

rahularya50:
I did some testing, and it seems that this PR finds the _latest_, _submitted_ backup, whereas the other PR finds the _earliest_, _graded_, backup.

Which of these behaviors is desirable?

cycomachead:
Wow, thanks! I suppose the other PR's behavior of earliest backup that's included sounds best to me.

rahularya50:
Actually, Kavi and I were discussing this, and your PR's behavior may make more sense. Otherwise in multi-part projects, a backup could be given a checkpoint score, and then it'd push the submission time earlier.

Whereas your PR looks at the latest submission / flagged backup, which might be more reasonable. I'm not sure though. The worry is that the behavior is probably course-dependent, so if we arbitrarily pick one way of computing the time, then it could hurt other courses.

rahularya50:
What could be done is leaving Export Grades the way it is, but adding a new job that exports just the submission times of various assignments computed in many different ways (e.g. earliest backup for each kind, as well as time of earliest / latest submission). That way every course can do what it wants, but we don't clutter up Export Grades with non grade-related behavior.

cycomachead:
> Otherwise in multi-part projects, a backup could be given a checkpoint score, and then it'd push the submission time earlier.
> The worry is that the behavior is probably course-dependent, so if we arbitrarily pick one way of computing the time, then it could hurt other courses.

I do think there's a reasonable default, which is that if you have multiple parts you count the main one -- slip days don't really make sense for a checkpoint, and if you have N equal weights, you'd probably do that as 2 separate assignments. I definitely think stuff like comp scores and revisions would be excluded. At least, as a default. 

Right now isn't there only 1 submission time on the assignment grades export? That's what I think I was following....

> What could be done is to leave Export Grades the way it is, but add a new job that exports just the submission times of various assignments computed in many different ways (e.g. earliest backup for each kind, as well as time of earliest / latest submission).

This actually seems pretty good to me, as long as it's still one sheet. 😄 Though, as I said, for slip days, the delta is much more useful than the absolute time.

cycomachead:
oh this is neat, lol; https://github.com/okpy/ok/pull/1148

matulka:
@cycomachead Hello Michael,
I have looked through #1148 and it looks functional, but there is one problem. It calculates slip days for all of user's submissions in a csv with the following format: <email> - <slip days used>. So it becomes kind of complicated to look through each submission. I think it would be reasonable to make a csv such that rows are different users and columns are different submission kinds (total, composition, etc.). For each user and each kind the csv would indicate the slip days used for the submission that gained that user the biggest score on that kind. Do you agree with this architecture? And what exact kinds should I add there? One possible solution would be to list just two kinds: Total score and Composition score. Another would be listing a submission for each of the following: Effort, Total, Regrade, Composition and Revision. What would be the most useful way for staff to calculate slip days?

cycomachead:
> . It calculates slip days for all of user's submissions in a csv with the following format

It's important for usability that all submissions are in a single table / row, since the point is to sum things up / review.

In general, I think slip days only really matter for the main submission, and if that's what you means by the biggest score, then I agree. Breaking down the score is potentially helpful, but not critical.

matulka:
@cycomachead Try taking a look at #1384 , I took #1148  and updated it so it works the way we discussed and also took into account most of the comments under it.

Issue #1273, opened 2018-05-14
Grade export is slow


papajohn:
The grade export job for CS 61A this semester looks like it's going to take about an hour to finish. That's fine, but we could probably make it faster one day. This issue isn't urgent.

https://okpy.org/admin/course/91/jobs/5050/

PR #1376, opened 2019-12-31
Revamped final grade export

rahularya50:
Addresses #1273 , hopefully.

Also makes the checkboxes on the form actually do something, and implements functionality originally in #1375 .

### DIFF
--- a/server/jobs/export_grades.py
+++ b/server/jobs/export_grades.py
@@ -54,10 +54,58 @@ def get_headers(assignments, *, export_submit_time):
     return headers, new_assignments
 
 
-def export_student_grades(student, assignments, all_scores, *, export_submit_time):
+def collect_records(user_ids, assignments):
+    all_records = {}
+
+    for assign in assignments:
+        raw_assign_records = (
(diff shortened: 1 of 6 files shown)


server/jobs/export_grades.py
```diff
@@ -65,6 +68,48 @@ def export_student_grades(student, assignments):
...
+                Score.user_id.in_(user_ids),
+                Score.assignment_id == assign.id,
+                Score.archived == False,
+            )
+            .group_by(Score.user_id, Score.kind)
+            .order_by(Score.score)
+            .all()
+        )
+
+        members = GroupMember.query.filter(
```
kavigupta: any reason to use db.session.query in the first case but GroupMember.query in the second case?

rahularya50: GroupMember.query is used when you want to get the member objects back, db.session.query is used when you want arbitrary columns. I need to use it here since I'm using an aggregation function to get the max.

kavigupta:
In case someone is finalizing grades, **do not merge** until 2020-02-01

PR #1384, opened 2020-03-04
Slips Calculator Job

matulka:
This pull request is supposed to solve the problem addressed in #1375 

This is an updated version of #1068 which takes into account most of the comments for that PR. The main changed made is that in the initial PR slips were calculated for each of user's submissions and in this PR they are calculated only for the final submission.

Slips can be calculated both for the whole course (with an option to change which exact assignments should be included) and for individual assignments.

Calculating for the whole course

Calculating slips for an assignment

Output of the job is an .csv table which contains and SID, email and the amoung of slip days (or hours, minutes) used by a student. If a student has not used slip days, they will not be included in the table.

Calculating slips log

Output csv for a specific assignment

### DIFF
--- a/server/constants.py
+++ b/server/constants.py
@@ -21,6 +21,8 @@ SCORE_KINDS = ['composition', 'correctness', 'effort', 'total', 'partner a', 'pa
                'regrade', 'revision', 'checkpoint 1', 'checkpoint 2',
                'private', 'autograder', 'error']
 
+TIMESCALES = ['days', 'hours', 'minutes']
+
 API_PREFIX = '/api'
 OAUTH_SCOPES = ['all', 'email']
 OAUTH_OUT_OF_BAND_URI = 'urn:ietf:wg:oauth:2.0:oob'
--- a/server/controllers/admin.py
+++ b/server/controllers/admin.py
@@ -33,9 +33,8 @@ import server.canvas.jobs
 from server.extensions import cache
 import server.forms as forms
 import server.jobs as jobs
-from server.jobs import (example, export, moss, scores_audit, github_search,
-                         scores_notify, checkpoint, effort, upload_scores,
-                        export_grades)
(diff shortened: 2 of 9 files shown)


cycomachead:
I think this will be great, but is it possible to have per-assignment slip time in the course CSV? So the header is `..., Total Slip {}, Assignment 1, Assignment 2, ...`

matulka:
@cycomachead Yeah sure, I will make it this week

matulka:
@cycomachead Hey! Sorry for the delay. Check the course slips once again, I think I have made it the way you ask.
Also, I have noticed a problem that sometimes it displays an SID and sometimes it doesn't. I think this is connected to the fact that in the current architecture of the database SID and User are connected using Enrollment, so it supports the possibility of a student being enrolled in different courses with different SIDs. When calculating course slips, I take the SID connected to the current course, while when calculating assignment slips, I take the SID from some random enrollment associated with this student (because it is faster this way than manually looking through all enrollments and picking the right one, O(1) vs. O(n)). Should I fix this and change to picking the SID from a specific enrollment? This shouldn't be a problem if in the actual database all students have the same SIDs for each course.

cycomachead:
@matulka Yes, this looks awesome now!! Thank you so much!! 🎉 🎊

matulka:
@cycomachead Great! You are welcome :)

cycomachead:
Any chance this could be merged soon or there is a way to run this, just to double check?

cycomachead:
Can I get a bump on this? :) @matulka @rahularya50 
Thanks!

matulka:
Hey @cycomachead ! Sorry so much for a late response. The thing is that I have not been working on okpy since the May of last year, so I have no idea about the possibility of merging this :(

cycomachead:
No worries, thank you! :)

PR #1399, opened 2020-08-20
Add API to control allowed redirect URLs for OAuth clients you own

rahularya50:


server/controllers/api.py
```diff
@@ -1026,6 +1050,54 @@ class File(Resource):
...
+        if not models.Client.can(client, user, "view"):
+            restful.abort(403)
+        return {
+            'client_name': client.name,
+            'owner_email': client.user.email,
+            'description': client.description,
+            'client_id': client.client_id,
+            'is_confidential': client.is_confidential,
+            'allowed_redirects': client.redirect_uris,
+        }
```
kavigupta, line 1077: wait did this not how exist? how does the current UI element work?

rahularya50: This is just for the API. `models.Client` always existed, and was used directly in the UI.

PR #1423, opened 2025-05-07
Upgrade Python->3.5, Gunicorn->19.9.0

rayzhao1:
okpy.org has been down since 5/2/25. This is a potential fix.

CS 61A infra team will retroactively document the root cause once we find the fix.


HuanzhiMao (approved):
LGTM

